PR #40, opened 2015-12-26
typofix

mLuby:


mxstbr:
Merged, thanks so much for finding and correcting that!

PR #42, opened 2015-12-27
Enforce linting in test suite

philihp:
Fixes https://github.com/mxstbr/react-boilerplate/issues/23. Looks like tab characters.

There's little reason not to lint code as early as possible, and rely on your build system to enforce style. By adding it to run after your test suite, hopefully this will catch bad code before it's merged into master.

Also, you can just `extend: "airbnb"` in eslintrc :smiley:

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,17 +1,18 @@
 {
-  	"extends": "./node_modules/eslint-config-airbnb/.eslintrc",
+  "extends": "airbnb",
     "env": {
       "browser": true,
       "node": true,
       "mocha": true
     },
-  	"ecmaFeatures": {
-      	"forOf": true,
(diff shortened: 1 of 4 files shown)


mxstbr:
This looks great, thanks so much for finding a better fix than waiting for eslint 2.0!

I'll merge this if you add a seperate `npm` command called `npm run lint` that runs `eslint .`, and add that to the `npm run test` command. That would make it so people can lint without having to run the tests!

mxstbr:
Closed in favor of #43, which does the same thing and has the changes above.

Thanks very much anyway for finding the solution for the issue and even submitting a PR, incredibly much appreciated!!

philihp:
No problem!

PR #53, opened 2015-12-28
Fix typo

kohei-takata:


mxstbr:
A transpiler is a source-to-source compiler, which PostCSS is. ([more info](https://en.m.wikipedia.org/wiki/Source-to-source_compiler)) that means transpiled is the correct word here!

Thanks for the PR anyway, much appreciated!

PR #58, opened 2015-12-29
Add Babel to README

### LINKED ISSUE #55: Mention Babel in the README
You mention PostCSS, but not Babel. I think these days babel is to JS what postcss is to CSS so you should add minimal doc for Babel :)

MoOx:
That's a start.

Closes #55


mxstbr:
Thanks so much @MoOx, much appreciated!

PR #60, opened 2015-12-29
Add component testing with Enzyme

philihp:
Adds an example of how to do a shallow component test.

In order for this to run, the hot reloading module must be moved from `.babelrc`, and into the Webpack config, as `npm test` is picking this up in the development configuration. This might be better, anyway, as all of the "hot reloading stuff" is contained inside of webpack configs.


app/containers/NotFoundPage/NotFound.react.js
```diff
@@ -1,5 +1,4 @@
 import React from 'react';
-import { Link } from 'react-router';
```
philihp: unused import

mxstbr:
Nice, enzyme looks very nice and like a good solution for component unit testing! PR much appreciated!

Can you walk me through why you had to move the babel config to the webpack query? What did it mess up? Is there any way we could keep the babel config inside .babelrc?

Two things before I merge:
Could you submit the history version bump as a separate PR so I can merge that individually, and could you remove the unused Widget component and instead add a test to the [Button Component](https://github.com/mxstbr/react-boilerplate/blob/v3.0.0/app/components/Button/Button.react.js)? That'd be great! :+1: 

Again, this PR is much appreciated, thanks for the help!

mxstbr:
While testing this, I noticed this failes if the component that is tested imports a `.css` file.

screen shot 2015-12-30 at 20 14 05

I assume this happens because we transpile the code with `babel` which has no idea how to handle the CSS. Any ideas how to work around this?

mxstbr:
The solution is to use https://github.com/css-modules/css-modules-require-hook and to run this mocha command:

```
$ mocha --compilers js:babel-core/register,css:css-modules-require-hook --recursive 'app/**/*.test.js'
```

Working on this atm

mxstbr:
Fixed an merged :+1:

Thanks so much for the PR!

philihp:
Cool, I wasn't actually sure yet how to work around that :) Glad you got a solution!

Issue #14, opened 2015-12-17
Better documentation


mxstbr:
- [x] **Files**
- [x] **Explanation of the possible `npm` commands**
- [x] Testing
    - [x] **How to write unit tests**
        - [x] **Mocha**
        - [x] **expect**
        - [x] **Real world usage**
            - [x] **Reducers**
            - [x] **`rewire`**
            - [x] **Actions**
    - [ ] *How to write good unit tests* (see [this article](https://medium.com/javascript-scene/what-every-unit-test-needs-f6cd34d9836d) and [this response](https://medium.com/@bruderstein/fancy-assertion-libraries-a40d7a45ceb0) for inspo)
    - [x] **How to write component tests**
        - [x] **Shallow rendering**
        - [x] **enzyme**
    - [x] **Browser testing**
        - [x] **`npm run serve`**
- [ ] Explanation of the main design decisions behind this boilerplate (maybe incomplete)
    - [ ] User side
        - [ ] JS
            - [ ] ESNext Features (see #165)
            - [x] **ImmutableJS**
            - [x] React
                - [ ] react-router
                - [x] **Components/Containers split** (see #33, [this article](http://jamesknelson.com/should-i-use-react-createclass-es6-classes-or-stateless-functional-components/))
                - [x] **Redux**
                    - [x] **reselect** (see #61)
                    - [ ] redux-simple-router
                    - [x] **redux-saga**
            - [ ] eslint
                - [ ] Airbnb style
            - [ ] *FontFaceObserver*
        - [x] CSS
            - [x] **CSS Modules**
            - [x] sanitize.css
            - [x] **PostCSS**
                - [x] **cssnext**
                - [x] **autoprefixer**
                - [x] postcss-focus
                - [x] cssnano
        - [x] **Offline First**
            - [x] **ServiceWorker**
            - [x] **AppCache**
    - [x] **Developer side**
        - [ ] Webpack
        - [ ] Babel
        - [x] **Beginner**
            - [ ] Modules (`export` and `import`)
            - [x] **Prerequisites**
                - [x] **node**
                    - [x] *npm*
                - [x] **git**
- [x] **FAQs**
    - [x] **nodejs-legacy for linux** (see [here](http://plaidzooks.tumblr.com/post/36894116867/using-nodejs-on-debianubuntu-systems))
    - [ ] **`Error: listen EADDRINUSE 127.0.0.1:3000`**
    - [ ] Hot-reloading doesn't work?
[...]

PR #61, opened 2015-12-30
Add memoized selector example with reselect

Egor-Sapronov:
Add an example how to create memoized selectors with reselect library

### DIFF
--- a/js/components/pages/HomePage.react.js
+++ b/js/components/pages/HomePage.react.js
@@ -7,6 +7,7 @@ import { asyncChangeProjectName, asyncChangeOwnerName } from '../../actions/AppA
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { Link } from 'react-router';
+import { createSelector } from 'reselect';
 
 class HomePage extends Component {
   render() {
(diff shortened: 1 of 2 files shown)


mxstbr:
What's the benefit of memoizing the state selection?

Egor-Sapronov:
Perfomance, you could create selectors not only on full state, but on state branches like
```
{
  branch1: {},
  branch2: {}
}
```
Create selectors form one branch and you can use it in actions too, it's a good way to use store only for storing data and making decisions in actions by query data from state with same selectors

mxstbr:
I see, that's quite nice!

Sadly, I don't think it's suitable for integration by default, because it's too advanced for beginners.

You got me thinking that we need to add documentation for advanced patterns though. I've added your code example to the todo list (see #14) and if you're so inclined i'd love if you could write about the benefits and use cases of reselect!

Thanks so much for taking the time to submit this PR, very much appreciated!! :heart:

mxstbr:
@Egor-Sapronov you were right, reselect is actually really useful. :100:

Sorry for not including it before, I've now worked with it for a while and have decided to integrate it into the boilerplate by default. (Take a look at the `v3.0.0` branch of this repo)

PR #64, opened 2015-12-30
Buttons display inline-block.

adjohnson916:


philihp:
Can you make this change to the v3.0.0 branch?

adjohnson916:
@philihp Looks like it's already in v3.0.0: https://github.com/mxstbr/react-boilerplate/blob/v3.0.0/app/components/Button/Button.css#L7

I didn't realize you were working on v3.0.0, is master still being maintained?

mxstbr:
Master is still being maintained, but I'm not merging any big changes. `v3.0.0` is where the party's at!

mxstbr:
Thanks so much for contributing and the PR! Have a look at `v3.0.0` and see if you like it!

PR #82, opened 2016-01-11
fix npm command

johannesnagl:
"run" is missing i guess. see "npm run serve" above ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ With `npm run serve` you can start a server that's accessible in the entire loca
 
 ### Unit testing
 
-Unit tests live in the `test` directory, and are run with `npm test`. There are two files in there already, testing that the current AppActions and Reducers are working correctly. Mocha is used as a unit testing framework, check out the official [documentation](http://mochajs.org)!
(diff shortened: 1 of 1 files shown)


mxstbr:
Only arbitrary commands have to be run with `npm run-script` (long version of `npm run`), some common, standardised commands like `npm start`, `npm test` and others can be run while omitting the `run`.

It gets even better, if you omit the `run` and enter `npm test`, npm will run the following commands in that order: `npm run pretest`,  `npm run test`, `npm run posttest`. So if you have a test script and need to do something beforehand, put it in the `pretest` script and it'll run automatically before `test`! 

See: https://docs.npmjs.com/misc/scripts

Thanks for the PR, much appreciated! Also check out the v3.0.0 branch while you're at it.

johannesnagl:
:+1: i'm sorry! 

tl;dr of this PR:
1336405435_fat_guy_trampoline_dunk_fail

mxstbr:
Hahaha don't worry, a big thanks for even submitting a PR!

Issue #81, opened 2016-01-11
Development server startup logging


mxstbr:
This is a starter issue for newcomers to Open Source, feel free to claim it!

----

Jekyll has some very nice logging when starting the development server:

```
Configuration file: /YourUsername/YourSite/_config.yml
            Source: /YourUsername/YourSite
       Destination: /YourUsername/YourSite/_site
      Generating...
                    done.
 Auto-regeneration: enabled for '/YourUsername/YourSite'
Configuration file: /YourUsername/YourSite/_config.yml
    Server address: http://0.0.0.0:4000/
  Server running... press ctrl-c to stop.
```

We should add more logging, both when starting the development server and when starting the production server.

It would be nice if our servers logged `Press CTRL-C to stop` after they finished starting, and if the development server also logged `Hot reloading: enabled` shortly before the server has started.

**Tip:** [`webpack/server.prod.js`](https://github.com/mxstbr/react-boilerplate/blob/v3.0.0/webpack/server.prod.js) is the production server, and [`webpack/server.dev.js`](https://github.com/mxstbr/react-boilerplate/blob/v3.0.0/webpack/server.dev.js) the development server.

----

Steps to do this:

- [x] Claim this issue by commenting below (@7rulnik)
- [x] Setup a local copy of react-boilerplate
    - [x] Clone the repo with `git clone git@github.com:mxstbr/react-boilerplate`
    - [x] Enter the newly created folder with `cd react-boilerplate`
    - [x] **Go to the `v3.0.0` branch of this repo with `git checkout v3.0.0`**
    - [x] Install the dependencies with `npm install`
- [x] Add the feature mentioned above
- [x] Start a pull request
- [x] Review
- [x] Merge

**Feel free to ping me on [twitter](https://twitter.com/mxstbr) if you need any help/tips/pointers/are stuck!**

> Note: This feature is scheduled for v3 of this boilerplate, please don't forget to make your changes on the `v3.0.0` branch of this repo.

7rulnik:
I can do it tonight

mxstbr:
All yours @7rulnik!

mxstbr:
Oops, closed it on accident.

7rulnik:
@mxstbr
Should I add logs about directories?
If yes, what is the best option for log source dir?
Something like that? Or more smart method?
```
console.log(path.join(__dirname, '..', 'app'));

```

mxstbr:
[...]

PR #83, opened 2016-01-11
Add logs #81

7rulnik:
So, I made this. Maybe add some styling via [Chalk](https://github.com/chalk/chalk)?
#81

### DIFF
--- a/webpack/server.dev.js
+++ b/webpack/server.dev.js
@@ -1,10 +1,15 @@
 /* eslint no-console:0 */
 // Gets called when running npm start
 
+const path = require('path');
 const webpack = require('webpack');
 const WebpackDevServer = require('webpack-dev-server');
 const config = require('./webpack.dev.config');
+const ip = require('ip');
 
+console.log('Source: ' + path.join(__dirname, '..', 'app'));
+console.log('Destination: ' + config.output.path);
+console.log('Hot reloading: enabled');
 console.log('Starting server...\n');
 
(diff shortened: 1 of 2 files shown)


mxstbr:
Nice work! :+1:

One suggestion: We should only log source and not destination, since we don't actually write any files when those servers are started, so the destination doesn't even exist. That makes sense, right?

How would you style the output with chalk? I haven't worked with that at all, but would love for you to check it out!

7rulnik:
Thx!
Ok, I can continue tomorrow. 
I wanna do something like [this](https://monosnap.com/file/HjsLOWKuklHMPlhN21jiG62A2fC1zH) (browser-sync)

mxstbr:
I would absolutely love that, those logs look great! Feel free to make them as nice looking as you want!

mxstbr:
How is it going with this @7rulnik, need any help?

7rulnik:
Was busy at work. I will finish it at this week.

mxstbr:
I just updated the `server.prod.js` a bit, see commit 94487c8!

7rulnik:
@mxstbr 
Are we really support Node 0.10? I think that few people use it.
I wanna use template strings for logs.

mxstbr:
I can't find them right now, but I remember seeing Node version usage statistics and a large percentage was still on 0.10, which is why I decided to support it.

If you can find statistics that contradict that, we can talk about it, but I'd much rather be on the safe side.

PR #85, opened 2016-01-12
Update README.md

kvzivn:
Cloning from there didn't work.

```
The authenticity of host 'github.com (192.30.252.130)' can't be established.
RSA key fingerprint is SHA256:nThbg6kXUpJWGl7E1IGOCspRomTxdCARLviKw6E5SY8.
Are you sure you want to continue connecting (yes/no)? yes
Warning: Permanently added 'github.com,192.30.252.130' (RSA) to the list of known hosts.
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Made with :heart: by [Max Stoiber](https://twitter.com/mxstbr)!
 
 ## Getting started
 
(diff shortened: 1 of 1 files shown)


mxstbr:
TIL that doesn't work for everybody, thanks so much for correcting that!

PR #88, opened 2016-01-14
redux-simple-router

chungchiehlun:
The official [redux-simple-router](https://github.com/rackt/redux-simple-router) just released v2.0 and updated its API. I tried to combined it with the repository and it worked well.

I implement its powerful effect : Let `react-router` and `redux` do what they did as usual but keep both in sync. Accordingly , we combined two individual state , our application and **URL**. [This article](http://jlongster.com/A-Simple-Way-to-Route-with-Redux) illustrate the conecpt much clearly.

Without `redux-simple-router` , we used `<Link>` component provided by `react-router` to change URL. For example , `<Link className="btn" to="/readme">Setup</Link>`.

Now , **our application state is contained the URL**. We re-rendered our UI after updating our state by dispatching actions as usual.

```js
<button className="btn" onClick={ ()=> dispatch(routeActions.push('/readme' )) }>Setup</button>
```

PR #98, opened 2016-01-17
update Container Component and add redux-simple-router

chungchiehlun:
I sent [#88](https://github.com/mxstbr/react-boilerplate/pull/88) which I combined the repo with  [redux-simple-router](https://github.com/rackt/redux-simple-router) a few days ago.

Sorry for sending the new PR late , because I had busy weekend for the past two day. There are two main points about the new PR.

1. Add the [redux-simple-router](https://github.com/rackt/redux-simple-router) like what I did in [#88](https://github.com/mxstbr/react-boilerplate/pull/88)

2. Extract component methods(`onChangeProjectName` ,  `onChangeOwnerName` ...) from `HomePage` component. I think this made **Presentational Components** (`HomePage` , `ReadmePage`) more "PURE" and clear. Then, export Container components **connect()-ed**  to Redux by means of `mapDispatchToProps` and `mapStateToProps` suggested by [react-redux](https://github.com/rackt/react-redux/tree/b2b9f077600d6a648e75ba0bee38131baa68a58d).

The other problem is when I ran *lint* ,  the Error : "JSX props should not use arrow functions  react/jsx-no-bind" appeared. I thought the sytax of the new PR is correct. Maybe we can modify the lint ?

### DIFF
--- a/app/components/Button/Button.test.js
+++ b/app/components/Button/Button.test.js
@@ -25,9 +25,9 @@ describe('<Button />', () => {
     expect(renderedComponent.find('a').length).toEqual(1);
   });
 
-  it('should render a Link if the route prop is specified', () => {
-    const renderedComponent = shallow(<Button route="/" />);
-    expect(renderedComponent.find('Link').length).toEqual(1);
+  it('should render a button to change route if the handleRoute prop is specified', () => {
+    const renderedComponent = shallow(<Button handleRoute={ ()=> {} } />);
+    expect(renderedComponent.find('button').length).toEqual(1);
   });
 
   it('should handle click events', () => {
--- a/app/components/Button/Button.test.js
+++ b/app/components/Button/Button.test.js
@@ -26,7 +26,7 @@ describe('<Button />', () => {
   });
 
   it('should render a button to change route if the handleRoute prop is specified', () => {
-    const renderedComponent = shallow(<Button handleRoute={ ()=> {} } />);
+    const renderedComponent = shallow(<Button handleRoute="something" />);
     expect(renderedComponent.find('button').length).toEqual(1);
   });
 
--- a/app/containers/HomePage/HomePage.react.js
+++ b/app/containers/HomePage/HomePage.react.js
@@ -17,6 +17,7 @@ class HomePage extends React.Component {
     this.onChangeProjectName = this.onChangeProjectName.bind(this);
     this.onChangeOwnerName = this.onChangeOwnerName.bind(this);
     this.onChangeRoute = this.onChangeRoute.bind(this);
+    this.changeRouteToReadme = this.changeRouteToReadme.bind(this);
   }
   onChangeOwnerName(evt) {
     this.props.changeOwnerName(evt.target.value);
@@ -27,6 +28,10 @@ class HomePage extends React.Component {
 
   onChangeRoute(url) {
     this.props.changeRoute(url);
+  }
+
+  changeRouteToReadme() {
+    this.onChangeRoute('/readme');
   }
 
   render() {
@@ -51,7 +56,7 @@ class HomePage extends React.Component {
             defaultValue="mxstbr" value={ownerName}
           />
         </label>
-        <Button handleRoute= { () => this.onChangeRoute('/readme') } >Setup</Button>
+        <Button handleRoute = { this.changeRouteToReadme }>Setup</Button>
         <p> Here is {"'"}{ pathname }{"'"}</p>
       </div>
     );
--- a/app/containers/HomePage/HomePage.react.js
+++ b/app/containers/HomePage/HomePage.react.js
@@ -6,7 +6,7 @@
 import { asyncChangeProjectName, asyncChangeOwnerName } from './HomePage.actions';
 import React from 'react';
 import { connect } from 'react-redux';
-
+import { routeActions } from 'redux-simple-router';
 import Button from 'Button/Button.react';
 
 import styles from './HomePage.css';
@@ -16,18 +16,22 @@ class HomePage extends React.Component {
     super();
     this.onChangeProjectName = this.onChangeProjectName.bind(this);
     this.onChangeOwnerName = this.onChangeOwnerName.bind(this);
+    this.onChangeRoute = this.onChangeRoute.bind(this);
   }
-
-  onChangeProjectName(evt) {
-    this.props.dispatch(asyncChangeProjectName(evt.target.value));
-  }
-
   onChangeOwnerName(evt) {
-    this.props.dispatch(asyncChangeOwnerName(evt.target.value));
+    this.props.changeOwnerName(evt.target.value);
+  }
+  onChangeProjectName(evt) {
+    this.props.changeProjectName(evt.target.value);
+  }
+
+  onChangeRoute(url) {
+    this.props.changeRoute(url);
   }
 
   render() {
     const { projectName, ownerName } = this.props.data;
+    const { pathname } = this.props.location;
 
     return (
       <div>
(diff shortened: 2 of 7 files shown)


mxstbr:
Interesting approach. Am on the road currently, will have a better look as soon as I'm back. Thanks so much already!

mxstbr:
Thanks so much for this PR and the input, I love it!

A few things:

1. The linting error is correct. One shouldn't use arrow functions in JSX because it `.bind`s the context of the function, which is expensive to do on every render. Please replace them with normal `function`s! (See [here](https://github.com/airbnb/javascript/tree/master/react#methods) for more information.)

2. I love what you did with the `<Button />`, that makes a lot of sense. I think you might have to update the unit tests in `Button.react.test.js` though, please run `npm run test` to double check.

3. I'm not sure `mapDispatchToProps` is useful, it's more work. I now have to import an action and manually add an action to that function and then call said function to use it in the component, where before I could just import the action and dispatch it. What's the benefit?

chungchiehlun:
```js
function HomePage(props) {
  const { projectName, ownerName } = props.data;
  const { pathname } = props.loc;
  return (
    <div>
      // .......
      <input className={styles.input} type="text"
          onChange={ (evt) => { props.onChangeProjectName(evt.target.value); }}
          defaultValue="React.js Boilerplate" value={projectName}
        />
      // .......      
      <input className={styles.input} type="text"
          onChange={ (evt) => props.onChangeOwnerName(evt.target.value) }
          defaultValue="mxstbr" value={ownerName}
        />
      // .......
      <Button handleRoute= { () => props.onChangeRoute('/readme') } >Setup</Button>
    </div>
  );
}
```

The function `HomePage` is [stateless component](https://medium.com/@joshblack/stateless-components-in-react-0-14-f9798f8b992d#.au3mabnoy). It behaved as a pure function : *accepted a parameter(props) and rendered the component*. It's pure and reusable. It was not like `React.createClass` or `React.Component` which have component methods and we needed to care [bind problem](https://facebook.github.io/react/docs/reusable-components.html#no-autobinding). So I wondered whether the lint is necessary or not.

I don't think `mapDispatchToProps` is more work. Any component wrapped with connect() call will receive a dispatch function as a prop ( `this.props.dispatch` ). I just extracted component methods from class `React.Componet` and used a `mapDispatchToProps` function to wrapped them. They both did  identical thing , for example , provide a `onChangeProjectName` method  which dispatched   `asyncChangeProjectName` action.
 
```js
// component methods
onChangeProjectName(evt) { 
    this.props.dispatch(asyncChangeProjectName(evt.target.value)); 
}

// mapDispatchToProps function
function mapDispatchToProps(dispatch) {
  return {
    onChangeProjectName: (val) => dispatch(asyncChangeProjectName(val))
  };
}
```
The difference between the former and the later is the later made `HomePage` pure and reusable.

I forgot to update the test. I'll do it tomorrow. :)

mxstbr:
The reason not all of the components are pure components is because **pure components are not hot-reloadable**. ([src](https://github.com/gaearon/babel-plugin-react-transform/issues/57#issuecomment-167670016)) That's a massive deal-breaker and why I'm still using classes and thus why the lint is necessary.

Could you keep everything else the same, but make `HomePage` a class again? Thanks!

You're right, having `mapDispatchToProps` actually makes sense, thanks for clearing that up!

app/containers/HomePage/HomePage.react.js
```diff
@@ -6,61 +6,53 @@
...
+      <Button handleRoute= { () => props.onChangeRoute('/readme') } >Setup</Button>
+      <p> Here is {"'"}{ pathname }{"'"}</p>
+    </div>
+  );
 }
 
-// REDUX STUFF
+function mapStateToProps(state) {
+  return {
+    loc: state.routing.location,
```
mxstbr: Could you rename `loc` to `location`? I'm not a big fan of shortening names in the source since UglifyJS will shorten them for us anyway.

app/containers/HomePage/HomePage.react.js
```diff
@@ -6,61 +6,53 @@
...
+  };
+}
 
-// Which props do we want to inject, given the global state?
-function select(state) {
+function mapDispatchToProps(dispatch) {
   return {
-    data: state
+    onChangeRoute: (url) => dispatch(routeActions.push(url)),
+    onChangeProjectName: (val) => dispatch(asyncChangeProjectName(val)),
```
mxstbr: Same here, could you rename `val` to `value`?

app/containers/HomePage/HomePage.react.js
chungchiehlun: Okay , I'll rename these. BTW , The reason why I used shortening names is to give me a hint that `loc` is merely a meaningless variable rather a specific property.

mxstbr: Thanks. That's a nice convention, a beginner might be confused looking at the file and wondering wtf `loc` is – Lines Of Code? which is why I prefer full names.

mxstbr:
**Ping**

Any updates? I'd love to finally merge this PR, really looking forward to redux-simple-router!

app/containers/HomePage/HomePage.react.js
```diff
@@ -37,30 +41,38 @@ class HomePage extends React.Component {
...
         </label>
         <label className={styles.label}>Change to your name:
           <input className={styles.input} type="text"
-            onChange={this.onChangeOwnerName}
+            onChange={ this.onChangeOwnerName }
             defaultValue="mxstbr" value={ownerName}
           />
         </label>
-        <Button route="/readme">Setup</Button>
+        <Button handleRoute= { () => this.onChangeRoute('/readme') } >Setup</Button>
```
chungchiehlun: I need to pass down `handleRoute` **function**, but the lint  cause error when using arrow function. Is there any method to solve this?

app/components/Button/Button.test.js
```diff
@@ -25,9 +25,9 @@ describe('<Button />', () => {
     expect(renderedComponent.find('a').length).toEqual(1);
   });
 
-  it('should render a Link if the route prop is specified', () => {
-    const renderedComponent = shallow(<Button route="/" />);
-    expect(renderedComponent.find('Link').length).toEqual(1);
+  it('should render a button to change route if the handleRoute prop is specified', () => {
+    const renderedComponent = shallow(<Button handleRoute={ ()=> {} } />);
```
chungchiehlun: When I used `shallow` method ,  `handleRoute={ ()=> {} }` and `handleRoute='anything'` both passed the test.  Did shallow rendering ignore the propTypes ?
```js
Button.propTypes = {
  handleRoute: React.PropTypes.func,
};
```

chungchiehlun:
I just updated. But there are two points that I need some help.

app/containers/HomePage/HomePage.react.js
mxstbr: Yeah, by binding onChangeRoute in the constructor:

```JS
constructor() {
    this.changeRoute = this.changeRoute.bind(this);
}
```

and then using that function in the Button:

```JSX
<Button handleRoute={this.changeRoute} />
```

and then have  a changeRoute function:

```JS
changeRoute() {
    this.onChangeRoute('/readme');
}
```

Maybe there's a better way though.

app/components/Button/Button.test.js
mxstbr: Might be, good question. Will have a look at airbnb/enzyme

mxstbr:
Sorry, I didn't look at the full code for the HomePage, I see that you've done what I posted above already. The solution:

Add another function called `changeRouteToReadme`:

```JS
changeRouteToReadme() {
    this.onChangeRoute('/readme');
}
```

and call that from the onclick on the Button:

```JS
<Button onClick={this.changeRouteToReadme} />
```

Not the nicest way, but we should really try to avoid binding anything in the JSX!

chungchiehlun:
Thanks , this correctly solved my problem. I updated my PR and it worked well finally. :)
As you've said , this is not the nicest way. Maybe we can open an new issue to discuss how to enhance this.

The other question is about airbnb/enzyme. Its doc of *shallow render* didn't mention static method , so we must study its soure code for further understanding.

mxstbr:
Thank you so much, I'll look into enzyme! Cheers @chungchiehlun!

chungchiehlun:
You're welcome. I learned much from this repo. Thanks ! :)

PR #99, opened 2016-01-17
Fix typo in Babel Loader for Hot-Reload

amorino:
Brings the correct query so react-transform can hot-reload components properly.


mxstbr:
Thanks so much for sticking with it and catching that bug, so much appreciated! :heart:

Issue #74, opened 2016-01-06
Look into producthunt/chai-enzyme


mxstbr:
[Link](https://github.com/producthunt/chai-enzyme)

vesln:
let me know if you need any assistance

mxstbr:
@vesln thanks, will def do if something comes up!

philihp:
I like it! I'll try to take a stab at it on v3.0.0 sometime this week.

mxstbr:
:+1: nice @philihp!

mxstbr:
Closed as of #104!

Issue #79, opened 2016-01-08
Use Karma to run headless unit tests


mxstbr:
Currently we don't compile our tests with webpack, which means any component we test that imports non-js/non-css files (e.g. images) will not work.

Karma fixes this because it uses webpack and renders against a headless browser, which means if our application works with the webpack config it'll work in the tests.

Example Karma config: https://github.com/jquense/react-widgets/blob/master/karma.conf.js
Run with: `karma start karma.conf.js --single-run=true`

> Note: If you submit a PR, do it against the `v3.0.0` branch please!

mxstbr:
As I've never before working with Karma, this has proven to be harder than I initially thought.

If somebody else wants to take a stab at this please do so, I really want this to happen!

jbinto:
I was able to get the example Karma config linked in the issue working for one of my small projects.

It is not headless: on my OS X dev machine a Chrome window will pop up. It does work on Travis though (via `xvfb`), so that's good enough for me.

I'll send a PR shortly.

mxstbr:
@jbinto I'd love a PR, thank you so much! I thought you could use PhatomJS instead of a browser, and that'd be headless?

jbinto:
From what I've cobbled together from google, it doesn't look like PhantomJS is the way to go anymore. 
PhantomJS is based on old (c.2013) Webkit code, and the engines are different enough to cause trouble.

See e.g. 

* http://www.gizra.com/content/phantomjs-chrome-docker-selenium-standalone/
* http://stackoverflow.com/a/29433038/19779

and my favourite:

* https://twitter.com/stevenlangbroek/status/676445439940513793 
* https://gist.github.com/StevenLangbroek/552f639c95e2cf9f74ad

Many projects have moved to Chrome, e.g. [react-router](https://github.com/rackt/react-router/blame/af6c1ef9856267b223f666474171e62b1229af3b/karma.conf.js#L88). 

I also resisted moving away from jsdom/mocha, since it's so light-weight. But in the end I think Chrome makes more sense. 

--


As for a PR: I'm having some trouble with the webpack config right now. I think for some reason, it's running all of `node_modules` through webpack. 

I'll try to sort it out now.

mxstbr:
Amazing, thanks so much for the insight and the sources. 👏

I had similar problems with webpack, so I'm really looking forward to seeing how you solve them!

jbinto:
Okay, so I have karma working for the most part:

[...]

PR #104, opened 2016-01-19
Show a test using Chai-Enzyme

philihp:
Implemented a test using assertions from chai-enzyme, for #74. Are these `chai` tests significantly better than `expect` tests? I don't feel strongly one way or the other but maybe an example could be crafted that shows how they're better?

cc @vesln


mxstbr:
Thanks for doing this!!

Doesn't seem like a massive difference to be honest, not sure what the advantage would be. Maybe @vesln can shed some light on that?

> I think moving on from Mocha and using Karma to be able to use our webpack config when rendering the components to test them would be a much bigger improvement. See #79

vesln:
The value of `chai-enzyme`, or in general the value of any custom assertions, is that you end up with more expressive syntax, better assertion errors and reusability.

"More expressive syntax" is kinda vague and really depends on your taste. If you are using `expect/should` extensively, I claim that writing `expect(foo).to.have.attr('bar')` will be preferable over `expect(!!foo.attr('bar')).to.be.true`.

Reusability and encapsulation should be kinda obvious. enzyme's `shallow`, `render` and `mount` are different, have different api and in some cases, you will be using different api. `chai-enzyme` encapsulates this for you as much as possible, so you don't have to worry about it much. 

The other obvious benefit is protection against API breaking changes. You will be using the custom assertions, not enzyme's public methods (not always, but mostly), therefore if they change the api, we can make the necessary changes in one place (the assertion library) and if you are leveraging it, you will simply have to update the module vs. updating your entire test suite. 

This might not seem like a huge deal at first, but if you have thousands tests, it actually is.

About the better assertion errors, let's take the following component as an example:

```js
class PostImage extends React.Component {
  render() {
    return (
      <img alt='test' />
    );
  }
}
```

And the following test:

```js
const wrapper = shallow(<PostImage />)
expect(wrapper).to.have.attr('alt', 'other')
```

The exception message we will get will be:

```
AssertionError: expected <PostImage /> to have a 'alt' attribute with the value 'other', but the value was 'test'

     HTML:

     <img alt="test" />
```

As you see, it clearly states the component name, the actual attribute, the expected attribute and it prints the component as HTML (when possible). This gives you much more context than if you wrote:

```js
expect(wrapper.attr('alt')).to.equal('other');

// the exception will be something like:
// expected alt to equal other but it was test
```

Last but not least, `chai-enzyme` is a community driven project specifically created for enzyme. The authors of enzyme and others (like myself :)) participate actively in it and therefore you will benefit from the improvements we introduce "for free".

mxstbr:
Sounds good to me, what do you think @philihp ?

philihp:
Sounds good to me :smile: I agree that it reads better. I tend to reach this point in every test I write where I say "okay, I know what I want, now how do I write an assertion to say that?". I will defer to your experience that it's more intuitive this way. I'll ping you whenever I don't know how to express it.

My next question: can we have chai/chai-enzyme assertions for all of our tests, and remove the expect module?

mxstbr:
Agreed, could you look into that?

I'd like to switch to Karma for testing, see #79 for the reasoning behind that. I'm fine with merging this for now, but when I (or somebody else) gets around to implementing Karma this module won't be necessary since we won't be using enzyme anymore. I'm not sure I feel comfortable with releasing v3 without Karma, as I now know how shitty our testing configuration is. (Doesn't work when anything except `.js` files and `.css` files are imported, e.g. when Images are added Mocha fails)

marr:
Needs rebase

mxstbr:
@marr what do you mean?

marr:
@mxstbr There have been commits to master since this PR was made against it. There are no conflicts but the author should probably rebase it against latest master to avoid any merge commits.

marr:
@mxstbr I rebased it locally and see some warnings (unsure if new or existing):
```
Warning: Failed propType: Invalid prop `handleRoute` of type `string` supplied to `Button`, expected `function`.
```

mxstbr:
@marr could you check if they exist on the `v3.0.0` branch w/o this PR?

marr:
@mxstbr Yes they do exist on the `v3.0.0` branch

mxstbr:
@marr thanks for the report, will fix asap!

mxstbr:
@marr or maybe you want to take a look?

marr:
@mxstbr done https://github.com/mxstbr/react-boilerplate/pull/110

mxstbr:
I'm merging this for now, but I'm not sure if this change will last long. Thanks for having a look @philihp, and thanks very much for the input @vesln! (You might want to add that to the documentation?)

PR #106, opened 2016-01-19
[v3] fix the main js file reference in sw

hosmelq:


mxstbr:
I need to rethink the whole ServiceWorker thing, I'm also not caching any chunk files I just noticed. Do you recon there's a way to automatically add all files generated by webpack to the serviceworker.js file?

mxstbr:
Re: Prev comment, see #107

Thanks for this PR, much appreciated!

hosmelq:
how about this?

https://github.com/GoogleChrome/sw-precache

NekR:
@mxstbr, @hosmelq `sw-precache` is very good and in fact is done by better SW experts than me (creator of SW contribute there). But thing with `sw-precache` is that it's a bit more complicated, it's like using `gulp` for all the tasks webpack can do easily. I mean, if you are using webpack and do not want to configure (I mean wring SW code here) SW on your own, then `offline-plugin` is your choice. If you use something else or want to easily implement your own complicated logic in SW. then should use `sw-precache` since it's also kind of framework for SW.

Hope this helps.

PR #110, opened 2016-01-20
Fix warning for handleRoute prop

marr:
Pass a function instead of a string to avoid warning.


mxstbr:
Thanks so much @marr, this is very appreciated!

PR #111, opened 2016-01-21
Offline automatization [offline-plugin]

### LINKED ISSUE #107: Automatically cache all files in the ServiceWorker
With the chunking happening right now in `v3.0.0`, not all the files that are emitted by webpack get cached in the ServiceWorker.

We should use https://github.com/sporto/assets-webpack-plugin to emit a .json file that has a list of all files, which we can then cache using the ServiceWorker.

NekR:
* Remove AppCacheWebpackPlugin
* Remove hand-written ServiceWorker
* Add OfflinePlugin and use it to generate both ServiceWorker and AppCache
___________________________________

I hope it's okay to remove `AppCacheWebpackPlugin` since it's not needed anymore here. 
`offline-plugin` had couple of issues so I released ver. `2.0.0` to fix them and make it work fine with `HtmlWebpackPlugin` and other things required for `react-boilerplate`. 

I tested it with your prod builds and also tested it on `WebpackDevServer` (with `lazy: false`) and all seems to work. Let me know if you see any issues here.

Closes #107

Thanks!


NekR:
Fixed build. Sorry, copied everything to the end right before commit and forgot to test.

NekR:
Hmm.. it looks that you are testing on Node 0.10 and it doesn't have promises. Do you really need to support it, is there many people on it?

I can add polyfill, no problem. Just didn't saw much reason for it yet.

mxstbr:
Thank you so much, this looks great!

See [this comment](https://github.com/mxstbr/react-boilerplate/pull/83#issuecomment-172326267) as to why I currently test against 0.10. If you can disprove that enough people are still using it, I'd be very happy to drop support!

NekR:
@mxstbr seems that a lot of people are still using it for old projects https://semaphoreci.com/blog/2015/12/15/nodejs-version-usage-in-commercial-projects-2015-edition.html
But as far as I understand, boilerplate is needed only for new projects :-) Anyway, let me add support of promises of 0.10 and then I will commit a fix.

NekR:
Okay, fine now.

mxstbr:
Thanks so much @NekR, much appreciated!

NekR:
Sure, you are welcome!

PR #112, opened 2016-01-21
Fix dependencies

callumlocke:
- all using `^` and updated to latest versions
- changed psi to v2 API
- removed 'history' as it doesn't seem to be in use

npm run build and npm start both seem to work fine after this


mxstbr:
Thanks Callum, this is very appreciated!

PR #117, opened 2016-01-25
Karma, Chrome, Travis setup

### LINKED ISSUE #79: Use Karma to run headless unit tests

jbinto:
* Fixes #79.

* Reverts changes made in #104 until we can figure out how to get `chai-enzyme` to work as a Karma plugin.

* Replaces `rewire.js` with `babel-plugin-rewire`. (As it stands, rewire is incompatible with webpack). This means you can now reach into all ES modules with `__get__`  without having to require specially with `rewire`. This is only available when `NODE_ENV=test`.

* Adds a new `webpack.test.js`:

  * It doesn't inherit from the base config. 
  * Lots of tweaks needed for individual libraries (e.g. enzyme) and to accommodate the fact we're running our testing libraries in the browser.
  * Uses the `null-loader` for CSS (which is the whole reason you wanted off mocha/jsdom :wink:). Not sure if you'd ever want to actually run PostCSS for tests, but I can't see why you'd need it ATM. 
  * `json-loader` is needed to deal with many devDependencies that `require` JSON (since it's expecting a node environment).

* Modifies `.travis.yml`:

  * to only test on Node 5. It doesn't make sense (IMO) to repeat the tests on different Node versions since we're just launching a separate Chrome environment anyway. (Though the final webpack build that you do may differ...)

  * to add `xvfb` configuration so we can run a Chrome/Chromium browser,

  * to use a current Chrome instead of the ancient Chromium that Ubuntu 12.04 still uses.


mxstbr:
Thank you so much for your work, this is amazing. I'm pinging @vesln here again, he might have an idea how to make chai-enzyme work with Karma. (I really like chai-enzyme)

I don't quite understand why you'd not want to transform your CSS with PostCSS - we want to test our components exactly how they are, right?

vesln:
@mxstbr @jbinto thanks for bringing me in on this. Let's unify the discussion here https://github.com/producthunt/chai-enzyme/issues/16 as there might be other folks that would like to contribute to a final solution

mxstbr:
Amazing work @jbinto, I've merged for now. I really hope we can add chai-enzyme back to our roster of libraries again, I miss the assertions already!

Issue #129, opened 2016-01-29
ImmutableJS


mxstbr:
Since the last discussion (#46), I've read a lot and worked on a project with ImmutableJS. Last time I did not know about the performance benefits with `shouldComponentUpdate`. These are significant, because Immutable state equality operations (i.e. checking if the state of a component changed) are much faster.

I'm now reconsidering, and think using ImmutableJS by default is a good idea. With the new application structure (#27) this boilerplate is ready to scale, and ImmutableJS should not be missing in any big react app.

*For small applications it's not entirely necessary, but the syntactic sugar for reducer is a side-benefit that small apps also experience.*

I'd love a PR implementing ImmutableJS and [the PropTypes](https://github.com/HurricaneJames/react-immutable-proptypes)!

----

> Note: If you submit a PR, please do it against the **v3.0.0** branch of this repo!

mxstbr:
Done by #131!

PR #131, opened 2016-01-29
ImmutableJS

mxstbr:
Added an initial implementation of ImmutableJS, see #129.

I haven't worked enough with it to know all the best practices and standards, so I need some people with more experience with Immutable Redux Apps to review this. I managed to get it working with `combineReducers` and `react-router-redux` (prev. `redux-simple-router`) by using [`redux-immutable`](https://github.com/gajus/immutable).

@okonet @nikgraf @zavelevsky @borisd any input on this?


app/containers/HomePage/index.js
```diff
@@ -22,9 +22,15 @@ class HomePage extends React.Component {
     this.onChangeRoute = this.onChangeRoute.bind(this);
     this.changeRouteToReadme = this.changeRouteToReadme.bind(this);
   }
+
+  shouldComponentUpdate(nextState, prevState) {
```
nikgraf: https://facebook.github.io/react/docs/component-specs.html#updating-shouldcomponentupdate gets passed in nextProps and nextState

comparing them takes actually more effort. I recommend to use this https://github.com/gaearon/react-pure-render

nikgraf:
looks good!

app/containers/HomePage/index.js
mxstbr: No idea what my brain was thinking there :+1:

Not sure this can be faster than `(nextProps.ownerName !== this.props.ownerName || nextProps.projectName !== this.props.projectName)`, but then I don't know how it works internally.

mxstbr:
Eslint apparently doesn't like the class property transform in the `react-pure-render`, so I think I'm going to have to use this implementation for now.

nikgraf:
you can import the shallowEqual directly from that repo as well

okonet:
I was going to say the same + I think current implementation of `shouldComponentUpdate` is broken anyway. 

For eslint you should just use `babel-eslint` as the parser

mxstbr:
Implemented :+1: Waiting to hear back from Travis, then merging. Thanks for the help!

willopez:
react-pure-render is no longer maintained, [https://github.com/gaearon/react-pure-render](https://github.com/gaearon/react-pure-render). It is now suggested to use `react-addons-shallow-compare ` or [https://github.com/acdlite/recompose](https://github.com/acdlite/recompose)

Issue #10, opened 2015-12-16
Component/Action/Reducer command line generators


mxstbr:
https://medium.com/@yamalight/building-modular-javascript-applications-in-es6-with-react-webpack-and-babel-538189cd485f#.ujsh7hpzj

mxstbr:
This seems like a good starting point: https://github.com/ironer/este-cli

mxstbr:
This might be useful maybe https://github.com/tj/commander.js

mxstbr:
https://github.com/FormidableLabs/generator-formidable-react-component/blob/master/app/index.js

somus:
Can we use something like this? https://github.com/amwmedia/plop

mxstbr:
That looks very promising, thanks for the link @somus! Have you worked with it before?

somus:
I use this at work. If it is ok, I could do a PR.

mxstbr:
@somus I'd love a PR for that! Could you do it against the `v3.0.0` branch? Thanks!

somus:
Sure.

mxstbr:
Done by the amazing @somus!

PR #152, opened 2016-02-07
Add generators for component, page, selectors and sagas

somus:
PR for issue #10 

I have couple of questions,
1. I'm not sure where to put the templates for the generators, right now it is at '/plop_templates'. I don't feel it is the proper place. Is there are anywhere else we can put it?
2. Can there be sub-components inside either components or containers folder? If that's the case, I can add a directory picker input to find out where to put the component.


mxstbr:
Holy moly @somus, that looks amazing. I'm in the middle of travel preparations right now, and seeing as this quite an important PR it's going to take me a few days to review this, but thanks a lot already!

----

I think we should put the templates in a subfolder called `templates` inside the `webpack` folder, and rename the `webpack` folder to something else. Maybe `config`, but that might imply the users can configure something about their application in that folder, so I'm not a fan of that – any ideas?

----

Yes exactly, `containers` should be their own option which'll generate a component that is `connect`ed to the global state. This connect call can also have an optional `mapStateToProps` and `mapDispatchToProps` function and a selector associated with it. Containers can have a `actions`-`constants`-`reducer` tupel associated with them and might have a selector in them. Components don't, they are only a `index.js` and a `styles.css` file, and not `connect`ed to the state. An example of a container that is only connected to the state:

```JS
// imports here
class Bla extends React.Component {
  // component here
}

export default connect()(Bla);
```

An example with the `mapXtoY` functions:

```JS
// imports here
class Bla2 extends React.Component {
  // component here
}

function mapStateToProps(state) {
  // A part of the state is returned here as an object
  return {
    partOfTheState: state.get['somepart']
  };
};


function mapDispatchToProps(dispatch) {
  // Some actions mapped to a dispatch call are returned as an object here
  return {
    onSomeAction: dispatch => dispatch(someAction())
  }
};

export default connect(mapStateToProps, mapDispatchToProps)(Bla2);
```

One example with a selector:

```JS
import someSelector from 'someSelector';

// component here

export default connect(createSelector(
  someSelector,
  (something) => { something }
))(Component);
```

And this is what the `connect` call would look like when you have both a selector and the `mapXtoY` functions:

```JS
export default connect(mapStateToProps, mapDispatchToProps, createSelector(
  someSelector,
  (something) => { something }
))(Component);
```

> The above examples were not tested and might have typos in them, please be careful when copy & pasting them.Take a look at the [`App`](https://github.com/mxstbr/react-boilerplate/tree/v3.0.0/app/containers/App) container for an example of a container with !

I'm not sure that made any sense at all – if you want to we can jump on a quick Skype call soon and go through all the small intricacies together!

mxstbr:
So, I just tried this and it's absolutely amazing. I realised you've already included the whole `mapXtoY` thing, which is working great! Sorry for not checking before writing my above comment. Very good start, really really nice, I love it! 

A few nitpicks to make this really solid and perfect:

- At the start, one should be able to choose between `container` and `component`, instead of `page` and `component`. It's just a coincidence all the pages are in the `containers` folder, and tbh almost all of them shouldn't be there anyway. Most of those pages should be in the `components` folder. `containers` either have a `connect` call associated with them and/or an actions/constants/reducer tupel.

- When we add actions, we don't need to ask if the user wants a reducer as well – I've yet to encounter a case when I wanted component specific actions but not a reducer.

- Instead of asking for a reducer, ask if they want to add a component specific selector to the `container` (currently `page`). See my last comment for an example of what that might look like. (Adding the `createSelector` function call to the `connect` call should be enough)

- When a new saga is added, it should be added to the object in the `sagas/index.js` file so it's automatically included in the Redux middleware.

- When adding a new route and choosing the path, somebody might add a prefixed slash, i.e. `/about`. We should take care of not prefixing another slash in front of that route, otherwise it becomes `//about`.

- When adding a new component, we add a comment at the top saying `ComponentName.react.js` even though the file is called `index.js` – lets change that to simply say `ComponentName`!

- When adding styles, I love how a class is automatically added to the wrapping div, but right now it says `styles.hello-world` which is invalid JS. Either we change it to say `styles['hello-world']` or change the class that's generated to not-include hyphens, i.e. `helloWorld` and say `styles.helloWorld` in the component. (I'd prefer the second option, but the first is fine too)

Again, those are just a few nitpicks. This is absolutely amazing, thanks so much for this PR!

mxstbr:
Please note, I just moved the routes to their own `routes.js` file, so we need to update the route generator! :+1: See https://github.com/mxstbr/react-boilerplate/commit/ccd4f6cf81d76c360b7b983fc8ef9796c3c07b1a

somus:
Thanks for the feedback. I'll get on to the changes.

After that, we can schedule a skype call to go over it and make it perfect.

somus:
I've made changes based on the feedback.

I've added a new route generator which will pick components from both `components` and `containers` folder with a name ending with **Page** (I'm not sure this is the right way).

For adding selectors to the container, should we show a list of selectors present in `selectors` folder and then add it to the container?

mxstbr:
Amazing work so far, I love love love this!

A few more tiny changes to make this perfect:

- When generating a container/component, the files in the test subfolder should be called `actions.test.js` and `reducer.test.js` instead of `actions.js` and `reducer.js`

- I love the way the route generator works, but I don't think only listing all the components ending in `-Page` is the way to go. People might want to have routes to other components, so I think we should simply let them choose from all components and containers.

- I think it'd make sense to ask people if they want a selector inside a container, and then let them choose/generate one.

- If people enter `nameSaga` as the name for a Saga, can we strip the duplicate saga? Same thing with the selectors and the routes with the prefixed slash.

somus:
Made the changes.

I'm showing all the components for route generation. But, if we show everything it will be difficult to find the component for which you need to generate the route for. This ultimately destroys the purpose. Adding routes manually will be much easier in that case.

mxstbr:
Amazing!! What about manually entering the component name we want the route to? That's still easier than manually making the route.

somus:
That makes sense. We can add validations to check whether the component actually exists in either `components` or `containers` folder.

webpack/templates/component/stateless.js.hbs
```diff
@@ -0,0 +1,25 @@
...
+*
+* {{ properCase name }}
+*
+* {{ sentenceCase description }}
+*/
+
+import React from 'react';
+
+{{#if wantCSS}}
+import styles from './{{ dashCase name }}.scss';
```
mxstbr: This should be `{{ dashCase name }}.css` not `.scss`.

plopfile.js
```diff
@@ -0,0 +1,320 @@
...
+        return 'description is required';
+      }
+    }, {
+      type: 'confirm',
+      name: 'wantCSS',
+      message: 'Do you want to create corresponding CSS file?'
+    }, {
+      type: 'confirm',
+      name: 'connectRedux',
+      message: 'Do you want to connect the page to a redux store/dispatch actions?'
```
mxstbr: We don't need to ask this here, because if they chose to generate a `container` they WANT to be connected to the store – otherwise it's not a `container` but a `component`!

plopfile.js
```diff
@@ -0,0 +1,320 @@
...
+    }, {
+      type: 'confirm',
+      name: 'wantActionsAndReducer',
+      message: 'Do you want to generate corresponding actions/reducer files for the container?',
+      when: answers => answers.connectRedux
+    }, {
+      type: 'confirm',
+      name: 'wantSelector',
+      message: 'Do you want to add selectors to the connect method?',
```
mxstbr: This doesn't work for me, when selecting a selector it doesn't add it to the `createSelector` call in the index.js

webpack/templates/route.hbs
```diff
@@ -0,0 +1,8 @@
+ {
+    path: '/{{ path }}',
```
mxstbr: This should be `path: '{{ path }}'`, the prefixed slash is a "reset" and does not always have to be used!

mxstbr:
I added a bunch of comments for tiny fixes to make this even better than it already is. I don't quite understand why we ask if they want to add a selector when generating a container, when that doesn't do anything. Are you working on this or did it turn out too hard? We could just scrap that question and simply add a `createSelector` call without any specific selector.

plopfile.js
somus: Did you choose `Select from already available selectors` or `Generate new one`?

If you chose the first option, you can select multiple selectors from the list which you want to add to the `createSelector` method.

If you chose the second one, you have to enter name and description, it will be added to `createSelector` method.

I just checked both of them. They are working as expected.

webpack/templates/route.hbs
somus: Fixed in a2605d999b1e67358d3677c5b9201098bcb93770

plopfile.js
somus: Removed in a2605d999b1e67358d3677c5b9201098bcb93770

webpack/templates/component/stateless.js.hbs
somus: Fixed in a2605d999b1e67358d3677c5b9201098bcb93770

mxstbr:
On my phone at the moment, I'll check your fixes out asap! 

Meanwhile, add me on Skype and let's chat there,  I'm `mstoiber05` there.

somus:
The first option from the list lets you add already available selectors to your container. Other one lets you create a new selector and add it to the container.

plopfile.js
```diff
@@ -0,0 +1,319 @@
...
+    }, {
+      type: 'checkbox',
+      name: 'selectors',
+      message: 'Choose the selectors which has to be added (select/deselect selectors by pressing space)',
+      choices: fs.readdirSync('app/selectors').map(dir => ({ name: dir.slice(0, -3), value: dir.slice(0, -3) })),
+      validate: value => {
+        if (value.length > 0) {
+          return true;
+        }
+        return 'atleast one selector must be selector';
```
mxstbr: Typos, this should read: "At least one selector must be selected"

Maybe add a notice that one selects a selector with space as well?

mxstbr:
I think it's perfect now, thanks so much for doing this @somus – so awesome!

PR #153, opened 2016-02-07
Update logs

7rulnik:
I added "✓" symbol for success logs and updated logs for init and pagespeed


mxstbr:
Amazing work, thanks so much, these look absolutely amazing!

Issue #156, opened 2016-02-08
Use scroll-behaviour for RouteTransitions


amorino:
Implement the use [scroll-behaviour](https://github.com/rackt/scroll-behavior) to scroll top when the route change.

mxstbr:
Would love a PR for this!

mxstbr:
Done in #164 thanks @somus!

PR #164, opened 2016-02-13
scroll to top after route transition

somus:
Fixing #156


mxstbr:
This looks great, thanks!

PR #167, opened 2016-02-13
Use "trusty" on Travis for isolated builds

### LINKED ISSUE #166: Upgrade Travis to use trusty beta
https://docs.travis-ci.com/user/trusty-ci-environment/

> Travis CI runs each Trusty build in an isolated Google Compute Engine virtual machine that offer a vanilla build environment for every build.

> This has the advantage that no state persists between builds, offering a clean slate and making sure that your tests run in an environment built from scratch.

In #155, I forgot to include `sinon` in `package.json`. I had it in my `node_modules`, so "it worked on my machine (TM)". 

More importantly, Travis already had it on **it's** machine, so it missed something it should have caught.

With the new Trusty environment builds are entirely independent of each other, so this kind of thing won't happen.

jbinto:
Fixes #166.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: node_js
 sudo: true
+dist: trusty
 node_js:
   - "5.0"
 script: npm run build


mxstbr:
Thanks so much!

Issue #145, opened 2016-02-05
Add Redux DevTools


standino:
I think if  redux-devtools-dock-monitor  is added to the project. It will be very useful.

https://github.com/rackt/redux/blob/master/examples/real-world/containers/DevTools.js

https://github.com/gaearon/redux-devtools-dock-monitor

mxstbr:
Agreed, I'd love a PR against the v3.0.0 branch of this repo!

PR #173, opened 2016-02-15
PR for the issue #145: Please add redux-devtools-dock-monitor. 

standino:


mxstbr:
Alright, after trying this we still have a problem: If we build the application, the DevTools will still be bundled into the application, they just won't be shown. That's problem, because we want the code of them to not be included in production. Any ideas how to solve that?

standino:
Yes, you are right. 

Let's do more research or someone can help on this.

paul42:
inside the redux-saga repo, in the [examples/real-world](https://github.com/yelouafi/redux-saga/tree/master/examples/real-world), the 'root' container has this:
```
if (process.env.NODE_ENV === 'production') {
  module.exports = require('./Root.prod')
} else {
  module.exports = require('./Root.dev')
}
```
could we just segment out the parts from app.js that rely on redux-devtools and try a similar pattern? I don't know how the code-splitting would like this though, I don't know enough.

mxstbr:
Is that worth the increased complexity? We'd have to have two more files to see what actions run and to change the history in your application in a nice ui.

We could include the [redux-logger](https://github.com/fcomb/redux-logger) middleware instead to do the same thing, in the console, much easier.

mxstbr:
Sorry, but I don't think this is worth it – thank you so much for your contribution, but I don't see a good way to do this yet.

PR #187, opened 2016-02-22
Move config files to `package.json`

sotojuan:
Unfortunately for now ESLint does not allow ignoring stuff from `package.json`.


mxstbr:
This is indeed much nicer, appreciate this PR!

PR #189, opened 2016-02-22
Move `.eslintignore` to webpack directory

sotojuan:


mxstbr:
This might break `npm run lint:js:staged` - can you make sure that still works?

sotojuan:
When I run that, nothing happens, so I guess it's not broken?

mxstbr:
While checking this PR out, I noticed that our `.gitignore` is pretty much the same as our `.eslintignore`, at least as far as JS is concerned. This means we can use our `.gitignore` instead of the `.eslintignore`, which saves us a full file.

I went ahead and tried that, and it worked, so I've already pushed the change in be95f59d803124bbfc3d4038b3034beed7bba4ff.

Thanks for kicking off the discussion of removing cruft in our root folder, this has definitely made the boilerplate much nicer! Sorry I'm not merging this one, but I'm sure you'll find something else to do in our issues. Thanks for the contribution!!

Issue #113, opened 2016-01-22
WebpackDevServer breaks on complex nested routes


jongold:
Was running into this for a few days (on master & since upgrading to 3.0.0)

I had some routes like
`/releases`
`/releases/:version`

My `:version` param takes a semver number, like `1.2.3`.

They work perfectly when navigating around the app using `<Link />` etc, and `/releases` works when hard-refreshing the page, but hard refreshing on a url like `localhost:3000/releases/1.2.3` returns an error like this:
<img width="527" alt="messages image 3763230329" src="https://cloud.githubusercontent.com/assets/591643/12516743/f39566f0-c126-11e5-8157-888fae9e2ff5.png">

I confirmed that it's specifically having periods in paths that are broken (by adding a route handler on `/releases/foo`; finally managed to track it down to this issues. The fix proposed by stefanfisk works perfectly when I edited `webpack/server.dev.js`.

https://github.com/rackt/react-router/issues/676#issuecomment-73963306

I think this is a general enough problem to fix; it certainly shouldn't be happening - do you want me to clean up the code and submit a PR to 3.0.0?

mxstbr:
Wow thanks for the in-depth bug report, much appreciated!

This ties into #100 – do you reckon it's worth switching to express.js as the development server? I haven't worked with that at all, so I'd love a PR!

mxstbr:
I think we're fine sticking with `WebpackDevServer` for now, I don't really see any benefit switching to `express.js`. I'd love a PR implementing the fix by Stefan above!

mxstbr:
@jongold any update on this?

jongold:
@mxstbr it's still intermittently breaking for me; I want to find time to figure out why & make it bulletproof before I put code in other people's repos 😁

mxstbr:
Nice, sounds good! :+1:

mxstbr:
If you can't get it to work, a simple `express.js` server implementation like [this one](https://github.com/rackt/react-router/blob/master/docs/guides/basics/Histories.md#configuring-your-server) might solve the problem.

jshultz:
I'm running into the same problem. Once I get to a url like http://localhost/something/somethingelse a hard refresh breaks it. I see this in the console log:

`GET http://localhost:3000/classrooms/js/bundle.js 404 (Not Found)`

mxstbr:
Add the `publicPath` to your `output` option in your `webpack.prod.config.js`:

```JS
output: {
[...]

PR #194, opened 2016-02-24
Express Server

amorino:
Express Server with Webpack-Hot-Middleware and Dev-Middleware


mxstbr:
Amazing work @amorino!

After the first glance I love the new logging!

We now see new builds and errors in the terminal

```Shell
webpack built b38b8acaa89cfa923f49 in 6980ms
webpack building...
webpack built 97f498721ca27fe8bf96 in 717ms
webpack building...
webpack built f91a819d0520f444053a in 691ms
```

and in the browser console we see the names of the modules that were hot reloaded!

```Shell
[React Transform HMR] Patching HomePage
[HMR] Updated modules:
[HMR]  - ./app/containers/HomePage/index.js
[HMR] App is up to date.
```

Nice!

----

I'll CC @jongold here, he was the one who initially reported the bug about the WebpackDevServer – would you mind taking a look at this Jon, does it fix your problem?

webpack/webpack/webpack.dev.babel.js
```diff
@@ -36,6 +33,7 @@ module.exports = require('./webpack.base.babel')({
   // Add hot reloading
   plugins: [
     new webpack.HotModuleReplacementPlugin(), // Tell webpack we want hot reloading
+    new webpack.NoErrorsPlugin(),
```
mxstbr: What does the `NoErrorsPlugin` do and why do we need it? Don't we want to see errors?

mxstbr:
After trying it out, this does not seem to fix the original problem of #113:

screen shot 2016-02-25 at 09 37 24

Any ideas @amorino?

webpack/webpack/webpack.dev.babel.js
amorino: Because of [this](https://github.com/gaearon/react-hot-boilerplate/issues/1):

amorino:
Mhm, with the historyApiFallback it should't happen. I think is because of the dots (1.2.3), with plain routes it also throws an error?

mxstbr:
It works with plain routes, but the original point of switching to express was to fix the issues with the dots. Do you think you can get that to work?

amorino:
Yep, I will search for that. My bet is on the historyapifallback. I will check the library.

mxstbr:
Yep, that's it! See here: https://github.com/expressjs/express/issues/669#issuecomment-188949249

mxstbr:
According to [#5 in connect-history-api-fallback](https://github.com/bripkens/connect-history-api-fallback/issues/5) this should be possible with the [`rewrite` option](https://github.com/bripkens/connect-history-api-fallback#rewrites), which was specifically introduced to fix this issue.

mxstbr:
Any update on this @amorino? I'd love to finally merge this!

webpack/webpack/server.prod.js
```diff
@@ -16,16 +24,25 @@ console.log(
 );
 console.log('Starting server from build folder...');
 
-new WebpackDevServer(webpack(config), { // Start a server
+app.use(historyApi({
+  verbose: false
+}));
+
+app.use(webpackDevMiddleware(compiler, {
```
grabbou: Do we need webpack in production at all?

grabbou:
Great work, going to try that out later today.

Few concerns/questions:
- Do we need Webpack Dev Server in production at all? It feels like we should enable gzip in there and just serve static files that were build upfront
- Can we just make one file, called "server" and based on the NODE_ENV, just add in appropriate middlewares? E.g. if development, add webpack-dev-middleware etc., otherwise, turn on production performance improvements

mxstbr:
It appears there's a misconception here that comes from my bad naming practices - `server.prod.js` should actually not be used in production, ever. It solely exist to make `npm run serve` work and tunnels your local connection with `ngrok` to let you access your local application anywhere on the world – this should not be run in production at all, like, ever!

Thanks for the valuable feedback @grabbou, let's do the following:

- Add a **real, production ready `server.prod.js`**, enabling all performance improvements we can (e.g. `gzip`)
- **Rename `server.prod.js`** to something else (ideas? `server.serve.js` looks weird...) and have it start the production server and then tunnel *that* with `ngrok`.

grabbou:
Yeah, I started thinking I might be mis-using that file to be fair! Maybe let's name it serve.dev.js and serve.prod.js? 

Anyway - I was thinking about adding top-level folder called `server` since in the future, once we move towards adding e.g. server-side rendering, that might look better. In my particular use-case, I have to handle form submission with POST request, so it feels better to live at the root instead of `internals` folder.

mxstbr:
That sounds like a good idea, I like it!

mxstbr:
I prefer @grabbou's #211, as it fits better with this boilerplate. Thanks so much for your work and input though @amorino, if you have any suggestions I'd love to hear them in #211!

ASH-khan:
i am reading about the same issue from last 2 hours and all the solution posted here none is working :  here what i did in server.js file....
const express = require('express');
const path = require('path');
const port = 9990;
const app = express();

app.use(express.static(__dirname));

app.get('*',(req, res) => {
    res.sendFile(path.resolve(__dirname, 'index.html'));
});

app.listen(port);
console.log(`Server is Runing on Port # ${port}`);

//////////
app.get('*',(req, res) => {
    res.sendFile(path.resolve(__dirname, 'index.html'));
});
/////////

just add this function after the app.use() and working beautiful.
:)

Issue #188, opened 2016-02-22
Branding


mxstbr:
The current "Features" section is way outdated. For v3 I think we should focus less on the modules being used, and more on the things that these modules enable.

mxstbr:
Some of the things this boilerplate does (well) or focusses on: (This ties into #161)

- Developer Experience (Hot-reloading, performance focus with automatic code-splitting, PageSpeed Insights right from the CLI,...)
- Generators (in the `npm run generate` sense of the word)
- Offline Compatibility
- Domain Driven Development (Component encapsulation)
- Easily Testing
- Highly Scalable
- Takes care of a lot of details (hot-reloading reducers, code-splitting routes,...)

mxstbr:
Initial efforts at #196 by @oliverturner!

PR #196, opened 2016-02-25
1st draft of a more feature-focused approach. Addresses #188

oliverturner:
Less about the implementation, more about the value that the features add.

I've tried to keep it punchy and minimal with enough detail for the curious. Writing style is always a matter of taste: I'll completely understand if you have different ideas.

The final "but wait..." section is just a stub: I'm a newcomer to this project, so I'm unaware of many of the goodies you've baked in. Please suggest more of the value-adds - linting is one I just saw - I've missed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,19 +17,32 @@ Made with :heart: by [Max Stoiber](https://twitter.com/mxstbr)!
 
 ## Features
 
-- Using [**react-transform-hmr**](https://github.com/gaearon/react-transform-hmr), your changes in the CSS and JS get reflected in the app instantly without refreshing the page. That means that the **current application state persists** even when you change something in the underlying code! For a very good explanation and demo, watch Dan Abramov himself [talking about it at react-europe](https://www.youtube.com/watch?v=xsSnOQynTHs).
+- **Instant feedback** Enjoy the best DX and code your app at the speed of thought! Your saved changes to the CSS and JS are reflected instantaneously without refreshing the page. That means that the **current application state persists** even when you change something in the underlying code!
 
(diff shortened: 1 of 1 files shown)


mxstbr:
Great start, thanks for kicking things off!

I think we can get rid of the _via_s in the README, and mention the specific tools to achieve certain features in `docs/general/features.md`. We should definitely mention our new generators feature (ref #152), which is something that makes this boilerplate a lot nicer to use than others.

In the `but wait...` section, we could drop in "highly scalable" (or maybe "Build for scale" or something like that) and that we take care of a lot of details to make this a solid foundation.

I'll have to write down my thoughts about Domain Driven Development soon, and will add that later on as well!

Again, thanks for kicking things off and helping out!

mxstbr:
Oooh I just noticed you worked on the `master` branch. That branch is ~200 commits behind the `v3.0.0` branch – could you submit a new PR against that one? Thanks!

mxstbr:
I'm going to close this because this is not being merged, but thanks for the effort and please resubmit a PR against the `v3.0.0` branch!

oliverturner:
No problem: will make the needed amends and resubmit on the correct branch :smile:

grabbou:
That entire diff makes me feel like I have little english knowledge :o Looks really impressive!

oliverturner:
@grabbou so about that...

1. Your English is great.
1. Even for a native speaker it's really hard to verbally convey the value of the ideas behind this project and the way they've been structured in a way that's honest, informative and brief.

Thanks for the kind words: freshly inspired to give this another go :)

PR #201, opened 2016-02-29
Fix typo in webpack config

lfittl:
## Scope

`s/build.js/bundle.js/`

Alternative would be to remove those comments, but since I think the idea in the boilerplate is to be more  verbose than normal, I think just fixing the typo should do it :)

### DIFF
--- a/makewebpackconfig.js
+++ b/makewebpackconfig.js
@@ -70,7 +70,7 @@ module.exports = function(options) {
 
   return {
     entry: entry,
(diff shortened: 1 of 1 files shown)


mxstbr:
I'm no longer merging anything into the master branch - take a look at the `v3.0.0` branch instead! 

Still, thanks so much for taking the time to submit a PR, appreciate it!

lfittl:
Ah, that makes sense :)

Issue #197, opened 2016-02-26
Use Redux Devtools


carloscuatin:
I see that the repository uses so I think redux redux-devtools help a lot when watching as stocks soar among other oopciones that these have

mxstbr:
We've thought about implementing it before, and @standino actually went ahead and submitted a great PR, which I sadly had to discard – see #173 for my reasoning behind not implementing redux-devtools.

If you know of a way to solve the issues associated with it, feel free to reopen this issue!

carloscuatin:
yes, solution is with include in new.webpack.DefinePlugin a var _DEVTOOL_ true or false for not render in production

mxstbr:
The problem is not rendering it, but including the code for it in production. I don't want to simply not render it in production, I don't want the code for the DevTools in my bundle in production! Any ideas how to do that?

carloscuatin:
If I use just redux-devtools in development, and not be included in the production code

mxstbr:
How do you exclude it from the production bundle?

carloscuatin:
the CLIENT variable is defined in webpack.dev.js and also using it to create history without having to make a Root.dev.js and Root.prod.js as exponene in sagas or in the same repository redux![redux-devtool](http://i.imgur.com/VqmXLTG.png)

PR #207, opened 2016-03-03
add redux devtools

carloscuatin:
add redux devtools render in development 

### Development
image

### Production
image


carloscuatin:
#197 solution issue

knowbody:
lgtm!

mxstbr:
Hey Carlos, thanks so much for the help, this is very appreciated!

This PR currently has the same problem as #173, which is that you only *hide* the DevTools in production.

screen shot 2016-03-03 at 12 50 07

This is a screenshot of a newly built bundle, fresh from your branch. As you can see, there's still references to the `DevTools` in there. That means that **the code for the DevTools is still included in the production bundle**. This is a no-go, as the code is unused and shouldn't be in the bundle!

Do you have any idea how to not include the code for the DevTools in the production bundle?

carloscuatin:
use files different for compilation provider and store

mxstbr:
Could you implement that @carloscuatin?

carloscuatin:
@mxstbr  ok :+1:

luandro:
I've been using redux devtools for a while. I found that using [redux-devtools-extension](https://github.com/zalmoxisus/redux-devtools-extension) is way easier to implement (a single line and no additional packages), and less obtrusive.

carloscuatin:
@luandro yes exactly is another solution

mxstbr:
That sounds much more reasonable, would you mind implementing that? :+1:

mxstbr:
I'll close this issue because I can't see this being merged (thanks so much though @carloscuatin!), lets add the extension instead!

PR #208, opened 2016-03-04
codesplitting reducers

### LINKED ISSUE #199: Codesplitting reducers
Right now we codesplit routes (see [`app/routes.js`](https://github.com/mxstbr/react-boilerplate/blob/v3.0.0/app/routes.js)) but all reducers are loaded with the main bundle – that doesn't make a lot of sense in large applications. They should be loaded when the corresponding route is visited.

*See [this StackOverflow answer](https://stackoverflow.com/questions/32968016/how-to-dynamically-load-reducers-for-code-splitting-in-a-redux-application) for an example implementation.*

----

> Note: If you want to submit a PR for this, work on the **v3.0.0 branch** please!

somus:
Fixes #199


app/containers/HomePage/reducer.js
```diff
@@ -0,0 +1,57 @@
...
+  }),
+});
+
+function homeReducer(state = initialState, action) {
+  switch (action.type) {
+    case CHANGE_USERNAME:
+      // Delete prefixed '@' from the github username
+      return state
+        .setIn(['userData', 'username'], action.name.replace(/@/gi, ''));
+    case LOAD_REPOS:
```
mxstbr: Can we move the global `LOAD_REPOS`, `LOAD_REPOS_SUCCESS` and `LOAD_REPOS_ERROR` actions and constants back to the `App` component, only leaving the `CHANGE_USERNAME` in this reducer?

mxstbr:
Thanks very much, just one small nitpick! Also, could you add the changes to the templates for the `npm run clean` command?

somus:
When I build locally, eslint `react/prefer-stateless-function` error is not coming. It seems to appear only in Travis. Is that a newly added rule?

mxstbr:
That's weird, lets disable it for now, I've added an issue to track this: #212

I'll do a functionality test now and merge right afterwards, thanks so much for your help @somus!

mxstbr:
This works perfectly, thanks so much @somus!

PR #209, opened 2016-03-05
Upgrade to react-router-redux@4

jwinn:
Upgrade to react-router-redux. Feel free to remove/modify the initialState piece in app/app.js as appropriate. Forgot a semicolon, used to ASI. ;-)


app/app.js
```diff
@@ -43,12 +42,21 @@ import '../node_modules/sanitize.css/dist/sanitize.min.css';
...
-// 2. reduxRouterMiddleware: Syncs the location/URL path to the state
+// 2. routerMiddleware: Syncs the location/URL path to the state
 import rootReducer from './rootReducer';
 import sagas from './sagas';
-const createStoreWithMiddleware = applyMiddleware(reduxRouterMiddleware, sagaMiddleware(...sagas))(createStore);
-const store = createStoreWithMiddleware(rootReducer, fromJS({}));
-reduxRouterMiddleware.listenForReplays(store, (state) => state.get('route').location);
+// Configure history for react-router
+const browserHistory = useRouterHistory(createBrowserHistory)({
+  basename: process.env.BASENAME || '',
```
mxstbr: What do we need this for?

app/app.js
```diff
@@ -43,12 +42,21 @@ import '../node_modules/sanitize.css/dist/sanitize.min.css';
...
+});
+// Create redux store and sync react-router-redux
+const initialState = window.__INITIAL_STATE__ || {};
```
mxstbr: Why is something attached to the `window`, where's that coming from?

app/app.js
```diff
@@ -43,12 +42,21 @@ import '../node_modules/sanitize.css/dist/sanitize.min.css';
...
+const middleware =
+  applyMiddleware(routerMiddleware(browserHistory), sagaMiddleware(...sagas));
+const store = middleware(createStore)(rootReducer, fromJS(initialState));
+const history = syncHistoryWithStore(browserHistory, store, {
+  selectLocationState: (state) => state.get('route').toJS(),
```
mxstbr: Lets factor this out into a selector.

app/routeReducer.js
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * merge route state into immutable
+ *
+ * @param {object} state = initialState
+ * @param {object} action
+ * @returns {object}
+ */
+export function routeReducer(state = initialState, action) {
+  if (action.type === LOCATION_CHANGE) {
```
mxstbr: Can you make this a `switch` statement to stay in line with the other reducers?

mxstbr:
I've added a few question inline, please also add all the updates to the files we replace the example with when we run `npm run clean`. They files are in `internals/templates`!

Thank you so much for doing this, it's very much appreciated!

jwinn:
Went ahead and removed the initial state and browser history using a basename. Created a selectLocationSelector which I thought I'd separate from routeSelector, even though right now it just unwraps the route Immutable. I updated the clean process to not remove the new selector, but everything else in that folder; clean adds the routeReducer and I updated the other relevant files in internals/templates. Changed from if to switch statement in the reducer to match existing style.

I wasn't sure if travis tests `npm run setup && npm run clean`, so I manually tested this as well.
Also, for future PR, what's this project's preferred method of handling pull requests with several commits and amendments?

internals/scripts/clean.js
```diff
@@ -37,16 +37,32 @@ cp('internals/templates/homePage.js', 'app/components/HomePage/index.js');
...
+  'app/selectors/tests/routeSelector.test.js',
+  'app/selectors/tests/usernameSelector.test.js',
+  'app/selectors/currentUserSelector.js',
+  'app/selectors/errorSelector.js',
+  'app/selectors/globalSelector.js',
+  'app/selectors/loadingSelector.js',
+  'app/selectors/reposSelector.js',
+  'app/selectors/routeSelector.js',
+  'app/selectors/usernameSelector.js',
+]);
```
mxstbr: This is probably a bad idea because people might play with the example and add new selectors, which won't be deleted this way. I think deleting the whole folder and having a template for the location selector (and whatever else selector you need) is a better way to go here!

mxstbr:
Sorry for the late reply, I read this and then totally forgot about it... :flushed: This looks almost perfect, just one tiny amendment and we're done!

mxstbr:
Also, don't worry about the commits, I'd rather have the changes with explanations in there rather than squashing them!

jwinn:
Haha. I should have realized `npm run clean` is more like a reset than cleaning up folder structure. This should be good to go as I tested `npm run setup && npm run clean` to make sure only selectLocationSelector(.test).js are copied.

mxstbr:
Yeah, need to document that a bit better – will do soon, thanks for the reminder!

Thanks for cleaning up and doing this btw, I can't wait to get this merged!

jwinn:
I apologize for this taking a bit to get in. Work and wanting to determine root cause to why injectAsyncReducer in the Route.getComponent was causing a RangeError (infinite loop). My guess is that the new (in 4.0.0) syncHistoryWithStore and changing the store in the route causes the loop. If anyone else reads this and has intimate knowledge please weigh in. :-) For now, it will only inject (in getComponent) if the reducer key is not there. Also, store.js was not copied in `npm run clean` and the HomePage and NotFoundPage were being copied to the components and not containers folder which I changed back.

mxstbr:
If I understand you correctly we basically have to wrap each route with a reducer in `if (!store.asyncReducers.reducername)`?

mxstbr:
What about this? https://github.com/reactjs/react-router-redux/issues/314#issuecomment-190678756

jwinn:
Agreed. It's not ideal, which is why I asked for help in further research. I've been, swamped with work the last week and haven't had the available time to properly determine root cause and a correct fix. I will look into this further today to see if #314 is the root cause and apply a fix.

jwinn:
Was a relatively straightforward fix, just had to use Iterable.equals for a value equality check instead of !==

mxstbr:
Awesome, this looks much better!

jwinn:
I can make the change if you'd like. :-)

mxstbr:
Please do!

internals/scripts/clean.js
```diff
@@ -27,27 +27,35 @@ rm('-rf', 'app/components/*');
 // Cleanup containers folder
 rm('-rf', 'app/containers/*');
 mkdir('app/containers/App');
-mkdir('app/components/NotFoundPage');
-mkdir('app/components/HomePage');
+mkdir('app/containers/NotFoundPage');
+mkdir('app/containers/HomePage');
```
mxstbr: These should actually be in the components folder, as they don't have any associated state with them!

internals/scripts/clean.js
```diff
@@ -27,27 +27,35 @@ rm('-rf', 'app/components/*');
...
-cp('internals/templates/homePage.js', 'app/components/HomePage/index.js');
+cp('internals/templates/notFoundPage.js', 'app/containers/NotFoundPage/index.js');
+cp('internals/templates/homePage.js', 'app/containers/HomePage/index.js');
 
 // Cleanup sagas folder
 rm('-rf', 'app/sagas/*');
 cp('internals/templates/sagas.js', 'app/sagas/index.js');
 
-// Cleanup selectors folder
+// Cleanup selectors folder, leave selectLocationSelection(.test).js
```
mxstbr: This comment is no longer true, you can remove the ", leave selectLocationSelection(.test).js" part again! Thanks!

jwinn: To clarify, the comment is no longer necessary because we want clean to remove all selectors--which is problematic as selectLocationSelector is required for how we've setup react-router-redux, or just remove the extra info and keep the cp?

mxstbr: No, because we *are* deleting all selectors and then copying the `selectLocationSelector` from the `templates/` :+1:

internals/templates/app.js
```diff
@@ -15,15 +15,29 @@ import 'file?name=[name].[ext]!./.htaccess';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { Provider } from 'react-redux';
-import { Router } from 'react-router';
-import { browserHistory } from 'react-router';
+import { Router, browserHistory } from 'react-router';
```
mxstbr: Good catch with the two imports by the way!

PR #211, opened 2016-03-05
Alternative take on Express.js

grabbou:
Hey!

This is my alternative take on Express with modularised middleware-style approach.

Main difference is that it now does not have serve.prod and serve.dev and it's fully compliant with environments like Heroku (see my issue).

The trickiest thing was to use express handler (because historyApi is not working with dots). Fortunately, it's possible to access memory-fs used by webpackDevMiddleware so that's solved.

In production - for now, we have compression() (gzip) turned on by default.


mxstbr:
Awesome, thanks very much! This looks good at a first glance, I'll try it asap. One tiny thing, could you move the server to `internals/`?

grabbou:
Sure, moved - the only thing I left at root was `server.js` file - mainly because I expect some people (like me) might want to extend it by adding more stuff, e.g. POST handler. So internals now contain everything boilerplate-related.

mxstbr:
I'd still like `server.js` to also move to `internals` – you can still tweak it to your hearts content, but the root folder is meant to be as clean as possible!

The tests are only failing because of a missing return statement, run `npm run lint` to see the errors for yourself!

grabbou:
Yeah, I know - having linter in Atom, for some reason it wasn't showing earlier. 

Fair enough, you are the boss here :8ball: Let's just ignore that extraneous file for now - will reorganise that tomorrow properly in `internals/server`

mxstbr:
Awesome, thanks so much, I appreciate the help a lot!

To explain the reasoning: I don't think *a majority* of people are going to change the server, so it clutters the root folder for some/many. If they want to change it, they'll find it and change it anyway!

grabbou:
Sure! That is reasonable. All in all it's React.js focused and probably better for it to stay backend agnostic when possible.

mxstbr:
@oliverturner would love to get your input here, how hard would it be to swap out the production server for hapi (or anything really) and what could be done to make that easier?

grabbou:
FYI I am going to rebase that today & move the server.js into internals

oliverturner:
Chiming in as requested on the Express issue...

I _personally_ agree with the original structure from @grabbou that placed the `server` dir as a sibling of `js` in the project root: I get the argument for keeping the root clean, but I'd balance it with the aim of making _how_ the project works as self-evident as possible. More than that, if using the bundled server in production I feel as though it has stopped being an `internal` feature and become a first-class element of the application.

My main concern with the decision to move away from `webpack-dev-server` as a mere utility towards a full-fledged Express implementation is that it will _de facto_ make this an Express-focused boilerplate. The more work imposed on developers to remove / alter portions of it (in this instance updating deps, scripts, routes, etc), the less it will happen.

mxstbr:
Thanks for chiming in!

Fair enough, I can see that – lets put it back in the root directory. (sorry @grabbou!)

The general aim has definitely been to be the best *frontend baseline* as I don't have that much experience with node.js. This *will change* with v4 and universal rendering (as that is very important for bigger apps), so this needs to be the best solution we could possibly achieve to still be as baseline-y and modular as possible – how would you rather do it @oliverturner?

oliverturner:
Good to know where you're heading with the project.

The truth is that it really isn't easy to be particularly modular where SSR becomes involved: it's simply not possible to remain framework-agnostic at that point: the syntax of routing, dependency injection, etc. becomes necessarily specific.

My honest feeling is that you _should_ embrace Express fully at this point if that's the decision, but that you should do so understanding that this is the case.

mxstbr:
Makes sense. Why should we NOT embrace express fully?

oliverturner:
> Why should we NOT embrace express fully?

No particular reason not to: it's very popular and has a large ecosystem of supporting plugins. 

The counter-arguments would mainly be: 

- Uncertainty about its development: https://github.com/expressjs/express/issues/2844
- Developer preference: Koa and Hapi are worthy alternatives... but now I'm recapitulating my earlier point :)

mxstbr:
I'm not sure uncertainty about it's development is still true since it got [accepted into the Node.js Foundation](https://github.com/nodejs/TSC/pull/39), but I'm not very much into that scene so I don't really know!

From my standpoint, express seems like the best solution. We could possibly add example SSR implementations on top of Koa/Hapi/whatever else so one can easily switch to the preferred backend system. Does that sound reasonable?

oliverturner:
+1

mxstbr:
Alright, lets get this merged for now and then we'll take a look at the other implementations! Sorry for polluting this issue @grabbou, thanks for the great work!

mxstbr:
Could you resolve the conflicts so I can get this merged @grabbou?

grabbou:
Sure, I'll do it today. Just busy at work with other things atm.

grabbou:
@mxstbr done - is everything working for you as expected? ;)

mxstbr:
Awesome, I'll check it out later today - thanks so much!

mxstbr:
I love it! From a first look I can only see one small feature change and one small code change:

- `npm run serve` should run the production server and tunnel that to `ngrok`. `npm start` should run the development environment, which shouldn't be tunneling. (it makes no sense to tunnel that anywhere)

- Why do we have a `frontendMiddleware` if we don't have a `backendMiddleware`? It would make more sense to have a `middleware.dev.js` which exports what is now called the `addDevMiddlewares` as `devMiddleware`, and the same thing for the production middleware.

server/index.js
```diff
@@ -0,0 +1,43 @@
...
+const frontend = require('./middlewares/frontendMiddleware');
+const isDev = process.env.NODE_ENV !== 'production';
+
+const app = express();
+
+// If you are developing more than just front-end, like POST form
+// handling or API, add your custom middleware here
+// app.use('/api', myApi);
+
+// Initialize frontend middleware that will server your JS app
```
mxstbr: Small typo: `serve` instead of `server`

server/index.js
```diff
@@ -0,0 +1,43 @@
...
+const express = require('express');
+const logger = require('./logger');
+const ngrok = require('ngrok');
+
+const frontend = require('./middlewares/frontendMiddleware');
+const isDev = process.env.NODE_ENV !== 'production';
+
+const app = express();
+
+// If you are developing more than just front-end, like POST form
```
mxstbr: Lets change this comment to 

> If you need a backend, e.g. an API, add your custom backend-specific middleware here
> app.use('/api', myApi);

server/index.js
```diff
@@ -0,0 +1,43 @@
...
+// handling or API, add your custom middleware here
+// app.use('/api', myApi);
+
+// Initialize frontend middleware that will server your JS app
+const webpackConfig = isDev
+  ? require('../internals/webpack/webpack.dev.babel')
```
mxstbr: I'm not a fan of the ternary operator here, but can't come up with a nicer way to do this. Generally, ternary operators are very confusing for beginners, so if you can think of a better way...

server/index.js
```diff
@@ -0,0 +1,43 @@
...
+  : require('../internals/webpack/webpack.prod.babel');
+
+app.use(frontend(webpackConfig));
```
mxstbr: This API is kind of weird, why do we pass the config to our middleware – shouldn't we pass that to our application directly? (From a user perspective)

server/logger.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+module.exports = {
+
+  // Called whenever there's an error on the server we want to print
+  onError: err => {
+    console.log(chalk.red(err));
+  },
+
+  // Called when express.js app starts on given port w/o errors
+  onAppStarted: port => {
```
mxstbr: Not a fan of calling this `onAppStarted`, I think calling `logger.appStarted` makes more sense. Same thing for `onTunnelStarted` -> `logger.tunnelStarted`

server/middlewares/frontendMiddleware.js
```diff
@@ -0,0 +1,55 @@
...
+  // artifacts, we use it instead
+  const fs = middleware.fileSystem;
+
+  app.get('*', (req, res) => {
+    const file = fs.readFileSync(path.join(compiler.outputPath, 'index.html'));
+    res.send(file.toString());
+  });
+};
+
+const addProdMiddlewares = (app, options) => {
```
mxstbr: Is there anything else this middleware could/should be doing in a real production situation other than compressing?

server/middlewares/frontendMiddleware.js
```diff
@@ -0,0 +1,55 @@
...
+  const compression = require('compression');
+
+  app.use(compression());
```
mxstbr: Lets add a comment here explaining what this does.

server/middlewares/frontendMiddleware.js
```diff
@@ -0,0 +1,55 @@
+const express = require('express');
+const path = require('path');
+
+// Dev middleware
+const addDevMiddlewares = (app, options) => {
+  const webpackDevMiddleware = require('webpack-dev-middleware');
+  const webpackHotMiddleware = require('webpack-hot-middleware');
```
mxstbr: I generally keep all requires at a top level at the very beginning of the file for a quick overview of what is included, and I think it'd make sense to do the same thing here.

grabbou: This is included in a function to avoid webpack-dev requires in PROD environment when it might be not installed.

mxstbr: In what case would that not be installed? How could that happen?

server/index.js
grabbou: In this case, it probably does not really matter, but wanted to make it webpack independent (so it's location nor the place it's coming from does not mater - likely a candidate for a separate repo / folder - following up our previous discussion of making it agnostic).

server/index.js
grabbou: Ha, true!

grabbou:
@mxstbr thanks for the feedback! I'll try to get back to you on that later today / tomorrow. Some of the `weird` things like `why do we have but there's no` are probably because of me stripping out few files from internal infra based on this boilerplate! Will take a look once again to see how that looks in overall.

mxstbr:
Sounds good, can't wait to get this merged!

server/middlewares/frontendMiddleware.js
grabbou: `npm install --production` to skip dev dependencies in prod (e.g. after deploying).

server/index.js
grabbou: Maybe in this case `if (isDev) { webpackConfig = ''; }` would work? (as far as beginners are concerned)

server/logger.js
grabbou: True! That would've fit better if it was inside event emitter :D

server/middlewares/frontendMiddleware.js
grabbou: Well, in the scope of this middleware, I don't think so. We could add stuff like 500 page for uncaught exceptions, but it's more a fit for `server.js`.

server/middlewares/frontendMiddleware.js
grabbou: Maybe a link to Express.js docs would help as well? http://expressjs.com/en/advanced/best-practice-performance.html#use-gzip-compression

server/index.js
mxstbr: Nevermind that, I don't think many beginners will look into this file!

server/middlewares/frontendMiddleware.js
mxstbr: Ah that makes sense!

server/middlewares/frontendMiddleware.js
mxstbr: Yeah, very good idea! Here's a shortlink to that URL: `http://mxs.is/googmy`

server/middlewares/frontendMiddleware.js
mxstbr: Ahh, didn't even know `--production` exists, that's hella annoying, might actually bite us in other places? Good point of thinking about that though!

grabbou: Yeah, that's true. Mainly because we require webpack.prod.babel.js file in index.js. We need to extract common config of webpack to a file that does not have any requires, so we don't have to install it in production (e.g. publicPath). That's one of the possibilities.

grabbou:
> npm run serve should run the production server and tunnel that to ngrok. npm start should run the development environment, which shouldn't be tunneling. (it makes no sense to tunnel that anywhere)

Why do we want to tunnel production server? To me, it makes sense to have `staging` in that case. Since production is meant to run on destination server, it does not makes sense to me to tunnel it at the same time.

I mainly tunnel my development work to others, but that might be my very specific use-case.

mxstbr:
Ah yeah, I forgot we now have a truly in production usable server. Lets change that command then (`npm run prod`?) and have `npm run serve` run the development server and tunnel it.

Then we would have `npm start` for development, `npm run serve` for tunnelling and `npm run prod` (?) for production, does that make sense?

oliverturner:
Just a suggestion, but a task naming strategy that has worked well for my teams is `npm start`, `npm run start:prod`, `npm run start:whatevs`, etc.

Makes task variants easier to remember.

mxstbr:
Ooh that's nice! What about `start`, `start:tunnel` and `start:prod`?

grabbou:
I think npm start could just be a proxy to npm run serve npm run prod based on NODE_ENV. Otherwise we will have to add in another file to repo ( -1 ) to tell e.g. Heroku to execute it instead.

mxstbr:
Sounds good to me, and the commands too! :+1

mxstbr:
@grabbou any updates on this? I'd love to land this asap and finally get v3 out of the door!

grabbou:
Yeah, I was actually checking notifications yesterday and wondering when this one gets refreshed! Since I've confirmed the HMR issue to not be because of this PR but some other internal things I've made - I am happy to rebase and just merge.

I will send the updates later today on this one.

mxstbr:
Awesome, sounds good!

Could you also add [the link to the express.js docs](https://github.com/mxstbr/react-boilerplate/pull/211#discussion_r55729254) and the explanation [why there are non-top-level requires](https://github.com/mxstbr/react-boilerplate/pull/211#discussion_r55716526)? Thanks!

grabbou:
@mxstbr sure! Sorry for the delay - Easter ;) On my list for tmrw.

grabbou:
That moment when you spent few hours sorting out why `merge master` produced so many conflicts...

grabbou:
Regarding note about dynamic requires, I have to update webpack config to make that actually run, so let's wait with this till tomorrow.

grabbou:
@mxstbr I think we are ready now :)

mxstbr:
Awesome!! Great work, I'll leave this in this PR for now and add documentation about the new commands and the folder, but this looks perfect. (with the workaround being gone soon)

Thanks so much!

mxstbr:
I made the logging a bit nicer and added documentation about the new commands! This looks very solid and is pretty much ready to go, I'll just wait for the CI to finish!

mxstbr:
Boom, here we go! :tada:

Thanks so much @grabbou, great work!

PR #216, opened 2016-03-06
#207 Adding Redux Devtools Extension and fixing a small bug in sanitize.css

luandro:
Added Redux Devtools Extension support #207 and fixed small sanitize.css bug.

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -34,7 +34,7 @@ openSansObserver.check().then(() => {
 });
 
 // Import the CSS reset, which HtmlWebpackPlugin transfers to the build folder
-import '../node_modules/sanitize.css/dist/sanitize.min.css';
+import '../node_modules/sanitize.css/sanitize.css';
 
 const store = configureStore();
 
--- a/app/store.js
+++ b/app/store.js
@@ -1,4 +1,4 @@
-import { createStore, applyMiddleware } from 'redux';
+import { createStore, applyMiddleware, compose } from 'redux';
 import { fromJS } from 'immutable';
 import sagaMiddleware from 'redux-saga';
 import { syncHistory } from 'react-router-redux';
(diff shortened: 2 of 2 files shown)


app/app.js
```diff
@@ -34,7 +34,7 @@ openSansObserver.check().then(() => {
 });
 
 // Import the CSS reset, which HtmlWebpackPlugin transfers to the build folder
-import '../node_modules/sanitize.css/dist/sanitize.min.css';
+import '../node_modules/sanitize.css/sanitize.css';
```
mxstbr: What does this change, why would we import the non-minified version?

grabbou: To compile from source? :) Sanitize.css big +1 over normalize in my opinion is the fact it's highly configurable with css vars (or whatever).

However - this setup will require us to change the webpack settings. After I made that change I've noticed lots of broken CSS in my code. It turned out I had to enable postcss loader for node_modules as well, so it can parse css vars and stuff. I am also using it to compile basscss from source with custom settings.

mxstbr: We don't really want to do that though, we just want to add it to our stylesheet. If you want to compile it from source, I think you should add it manually, not from node_modules...

That problem was recently fixed, see: https://github.com/mxstbr/react-boilerplate/blob/v3.0.0/internals/webpack/webpack.base.babel.js#L26-L33

grabbou: What you mean by "I think you should add it manually" ?

Yeah, I am just saying the solution was to add `postcss` here https://github.com/mxstbr/react-boilerplate/blob/v3.0.0/internals/webpack/webpack.base.babel.js#L33 and they are still in global scope.

mxstbr: Download it and unzip it in your root component. :grin:

@okonet wrote that, have to ask him!

grabbou: Ha, fair enough. For now, I just added a yet another blacklisted folder called `assets` where I define global style with imports (that is processed with post-css which allows this stuff). If the current approach works well - I am happy with leaving it as is. Just something to be aware of for the future.

luandro: Hey just woke up. I think npm is installing a newer version of sanitize, where the dist folder doesn't exist anymore. So there's not a minified version of it included anymore. Clone the project and do a fresh npm install to see.

mxstbr: Damn, good catch! Thanks so much for fixing that!

PR #217, opened 2016-03-07
Fix typo in README

aried3r:
"safe fonts" instead of "save fonts".

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,13 +93,13 @@ For a full explanation of all files in the root folder, see [`docs/FILES.md`](do
 
 If you simply use web fonts in your project, the page will stay blank until these fonts are downloaded. That means a lot of waiting time in which users could already read the content.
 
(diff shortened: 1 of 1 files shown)


mxstbr:
I'm no longer merging anything into the `master` branch since version three is very far advanced! Thanks very much though for submitting this PR, it is very appreciated, and have a look at the `v3.0.0` branch!

aried3r:
Ah, I didn't see that mentioned anywhere. Thanks for pointing it out!

mxstbr:
It isn't actually mentioned anywhere. We should really add a notice to the master README that the `v3.0.0` branch exists and is a lot better... Do you maybe want to do that?

aried3r:
If I can include the typo fix, sure ;)

mxstbr:
Hah yeah definitely!

PR #219, opened 2016-03-07
Readme Updates

oliverturner:
Finally carved out the time to get to this...


mxstbr:
Awesome work, this is great! Thanks so much Oliver!

Two small things before I merge:

- Lets not mention the universal rendering just yet, it's going to take a bit more time before that is launched. (I want to make sure it's very solid and useable with different systems)

- Lets mention the generators, either as its own point or in some description. (I think they're worthy of their own point though) When users run `$ npm run generate` in the terminal, they can quickly add new components, containers (+ reducers, actions, constants), routes, sagas and selectors without having to remember to import things or setting up the boilerplate code.

Thank you so much, this is great start!

oliverturner:
Thanks for the feedback: the generators sound awesome & well worth highlighting!

I thought that letting folks know that SSR is at least under active discussion might be a good thing (since no deadline is mentioned), but I get your rationale. I'll be making an exploratory PR for that shortly :)

mxstbr:
Awesome, looking forward!

mxstbr:
I'll merge it for now, we can do incremental updates while it's live anyway. Awesome work @oliverturner, thanks for the help & discussions!

Issue #27, opened 2015-12-17
Application Structure


mxstbr:
Changing the application structure might be worth thinking about, right now the thing might seem quite bloated and randomly thrown together for newcomers.

mxstbr:
Maybe we should have a folder called `app`, which is the folder you write your application in and only contains the application files and folders, and a config folder for all the webpack stuff.

I think that would be an improvement, as right now build tools/application files/general config are quite jumbled together and hard to seperate if you haven't worked with some of those tools yet.

The structure of the boilerplate would then be as follows:

```
app/
    js/
    css/
    img/    <--- maybe rename this to assets
    .htaccess
    index.html
    manifest.json
    serviceworker.js
config/    <--- maybe this should be called webpack
    makewebpackconfig.js
    server.js
    webpack.dev.config.js
    webpack.prod.config.js
docs/
test/
.babelrc
.eslintignore
.eslintrc
.gitattributes
.gitignore
LICENSE.md
README.md
package.json
```

mxstbr:
I believe having the `test` (`tests`) folder inside the `app` folder makes sense. That way the separation of code-the-user-has-to-write and tooling is complete, and we can tell users they should only care about the `app` folder and let the boilerplate handle the rest.

The more I think about this structure, the more I like it. It makes using this boilerplate much easier and newcomer friendly.

mxstbr:
Maybe `config` is the wrong name for the folder with boilerplate things. That might lead devs to think that they can configure their application in there, which the can't.

Maybe `tooling` is a better name? I don't want to call it `webpack`, because there might be other config we want to put somewhere later on.

swastik:
I think calling it `webpack` is a good idea because it's very clear. `tooling` is somewhat ambiguous. You can always change the name later.

`img` should probably be changed to assets, yeah, because you could add webfonts, etc. that way.

swastik:
Also, I don't think tests should go inside app. They aren't the "app", technically... nothing there would transpile to the final build. A separate tests folder is fine... it's also probably more visible.

mxstbr:
Good points. I've implemented this in the [v3.0.0 branch](https://github.com/mxstbr/react-boilerplate/tree/v3.0.0), check it out and report any problems you come across!

swastik:
Looks pretty good!

hampusohlsson:
[...]

PR #229, opened 2016-03-14
Container Generator requires CSS file that is not created

wnstn:
In a default install of v3.0.0 (cloned about 15 minutes ago), generating a new Container and then attempting to require it on the homepage throws the following error (assuming the Container is named `LoginForm`).

```
./app/containers/LoginForm/index.js
Module not found: Error: Cannot resolve 'file' or 'directory' ./login-form.css in [localpath]/app/containers/LoginForm
resolve file
  [localpath]/app/containers/LoginForm/login-form.css doesn't exist
  [localpath]/app/containers/LoginForm/login-form.css.js doesn't exist
  [localpath]/app/containers/LoginForm/login-form.css.jsx doesn't exist
  [localpath]/app/containers/LoginForm/login-form.css.react.js doesn't exist
resolve directory
  [localpath]/app/containers/LoginForm/login-form.css/package.json doesn't exist (directory description file)
  [localpath]/app/containers/LoginForm/login-form.css doesn't exist (directory default file)
[[localpath]/app/containers/LoginForm/login-form.css]
[[localpath]/app/containers/LoginForm/login-form.css.js]
[[localpath]/app/containers/LoginForm/login-form.css.jsx]
[[localpath]/app/containers/LoginForm/login-form.css.react.js]
 @ ./app/containers/LoginForm/index.js 33:17-44
```

This is because the Container generator has a hardcoded CSS file path of `styles.css`:

```
path: '../../app/containers/{{properCase name}}/styles.css',
```

But the created component imports a module with the `dashCase` version of the component name:

```js
import styles from './{{ dashCase name }}.css';
```

This PR updates the generator to output a `dashCase` version of the css file inside the container directory, so that no error is thrown and also so an app full of components is also not full of `styles.css` files.

This resolves the error and allows the app to execute normally.

### DIFF
--- a/internals/generators/container/index.js
+++ b/internals/generators/container/index.js
@@ -89,7 +89,7 @@ module.exports = {
     if (data.wantCSS) {
       actions.push({
         type: 'add',
-        path: '../../app/containers/{{properCase name}}/styles.css',
+        path: '../../app/containers/{{properCase name}}/{{dashCase name}}.css',
         templateFile: './container/styles.css.hbs',
         abortOnFail: true,
       });
--- a/internals/generators/container/index.js
+++ b/internals/generators/container/index.js
@@ -89,7 +89,7 @@ module.exports = {
     if (data.wantCSS) {
       actions.push({
         type: 'add',
-        path: '../../app/containers/{{properCase name}}/{{dashCase name}}.css',
+        path: '../../app/containers/{{properCase name}}/styles.css',
         templateFile: './container/styles.css.hbs',
         abortOnFail: true,
       });
--- a/internals/generators/container/index.js.hbs
+++ b/internals/generators/container/index.js.hbs
@@ -19,7 +19,7 @@ import {{this}} from '{{this}}';
 {{/if}}
 
 {{#if wantCSS}}
-import styles from './{{ dashCase name }}.css';
+import styles from './styles.css';
 {{/if}}
 
 class {{ properCase name }} extends React.Component {


mxstbr:
That's a pretty good catch, thanks for trying out `v3.0.0`!

It is actually intended to have a common structure, so having a `styles.css` file is the fix- would you mind to swap that around? 

Thanks very much for the help!

wnstn:
Absolutely, love the goals of this project. Will update the PR tonight and double check all other generators.

One question just for clarification; what's the motivation for wanting to have a common structure? The result of this is that filenames give you no help in fuzzy-finding or grepping or in many other scenarios, instead you have to pay attention to both the parent director and the filename. Happy to amend the PR to whatever your goals are for the project, but personally I will probably change this on my project to be like this PR.

formula349:
I tend to agree. I'm not having trouble right now with the common file names (index.js, etc.), but as the project grows it seems like it would make it hard to navigate the project.

mxstbr:
There was a huge discussion about that in #27, read through it if you're interested in the why's. The result is the current application structure, and we are going to stick with it for now! Would love an update to this PR, thanks a lot!

wnstn:
Updated and ready to go once the CI approves!

mxstbr:
Awesome, lgtm!

Thanks so much for catching this @wnstn, let me know if anything else comes up! (I'd love feedback on v3 if you have any)

PR #232, opened 2016-03-15
Windows compatibility for v3

mxstbr:


mxstbr:
Can we have Travis CI run the tests on Windows too?

app/utils/tests/request.test.js
```diff
@@ -6,6 +6,8 @@ import request from '../request';
 import sinon from 'sinon';
 import expect from 'expect';
 
+import 'whatwg-fetch';
```
mxstbr: I'm pretty sure that isn't needed, we're stubbing fetch further down anyway

appveyor.yml
```diff
@@ -0,0 +1,46 @@
...
+
+# Set build version format here instead of in the admin panel
+version: "{build}"
+
+# Which branches to build
+branches:
+
+  # Blacklist
+  except:
+    - master
```
mxstbr: We actually want to build master, I've disabled builds for branches that don't have an appveyor.yml so we can safely remove this I think.

appveyor.yml
```diff
@@ -0,0 +1,46 @@
...
+
+# Do not build on gh tags
+skip_tags: true
+ 
+# Test against these versions of Node.js
+environment:
+
+  matrix:
+    # Node versions to run
+    - nodejs_version: "5.0"
```
mxstbr: We should run on the same versions here as we run on travis

app/utils/tests/request.test.js
jwinn: wanted to be consistent from a testing perspective

appveyor.yml
```diff
@@ -0,0 +1,46 @@
...
+# Disable automatic builds
+build: off
+
+# Post-install test scripts
+test_script:
+  # Output debugging info
+  - node --version
+  - npm --version
+  # run tests
+  - npm test
```
mxstbr: We should run `npm run build` here, which runs `npm test` anyway

app/utils/tests/request.test.js
mxstbr: Still, lets remove it so we don't confuse anybody.

mxstbr:
Awesome, awesome work, thanks for fighting through this, very appreciated!

w01fgang:
Now on Mac `npm run setup` fails with an error `sh: cross-env: command not found`
I think we need to add `cross-env` in the prerequisites section of the documentation or in installation instructions.

mxstbr:
Oh good catch, is it not in the dependencies?

mxstbr:
It is, so that *should* work...

mxstbr:
Damn I can replicate, it doesn't work. Give me a second.

mxstbr:
Right, the problem is that the `npm run setup` script uses cross-env, which isn't installed at that point because `npm run setup` installs the dependencies

w01fgang:
yes, we must to check first if `cross-env` is installed

mxstbr:
Fixed in 2d4c95d

mxstbr:
Thanks for notifying me of this, good catch!

w01fgang:
glad to help :+1:

PR #236, opened 2016-03-17
Create .nginx.conf

sergeycooper:


.nginx.conf
```diff
@@ -0,0 +1,66 @@
...
+##
+# HTTPS configurations
+##
+
+server {
+
+  listen 443;
+
+# Type your domain name below
+  server_name example.com;
```
mxstbr: Instead of repeating the `server_name` variable, can we put it at the top so people only had to change it once?

.nginx.conf
```diff
@@ -0,0 +1,66 @@
...
+    # Set path
+    root /var/www/;
+    try_files $uri /index.html;
+  }
+
+##
+# If you want to use Node/Rails/etc. API server
+# on the same port (443) config Nginx as a reverse proxy.
+# For security reasons use a firewall like ufw in Ubuntu
+# and deny port 3000/tcp.
```
mxstbr: That's an awesome and very useful addition! :+1:

.nginx.conf
sergeycooper: Unfortunately no.. Here why:
http://stackoverflow.com/questions/15416957/using-variables-in-nginx-location-rules

.nginx.conf
sergeycooper: Thanks! :)

.nginx.conf
mxstbr: Damn, alright. Thanks!

PR #243, opened 2016-03-23
Use depth=1 on the git clone

peterbe:
If you're going to delete the `.git` directory there's no point having to download the history. 
And I thought it was unclear which `.git` you should delete when you've cloned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,9 +38,9 @@ Version 3.0 is actively being developed on the [v3.0.0 branch], please check it 
 
 ## Getting started
 
-1. Clone this repo using `git clone https://github.com/mxstbr/react-boilerplate.git`.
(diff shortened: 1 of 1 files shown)


knowbody:
more about it: https://github.com/git/git/blob/9cfde9ee8f6dc82365f1e977935960f34108076b/Documentation/user-manual.txt#L2131-L2149

mxstbr:
This sounds like a reasonable thing to do, thanks so much for taking the time to submit this PR, we appreciate it!

Version 3.0 is actively being developed on the `v3.0.0` branch, and we are thus no longer merging PRs into master. Thank you so much for doing this, and sorry we won't merge it – definitely check out v3 and tell us what you think!

PR #250, opened 2016-03-29
Update .editorconfig

rikkert:
Reflecting CSS to 2 spaces change.


mxstbr:
Don't we need to change it to this?

```
indent_size = 2
```

rikkert:
Not needed because the [*] already covers the indent_size = 2

mxstbr:
Ah, makes sense. Thanks so much for sending in this PR, very appreciated!

Issue #249, opened 2016-03-29
3.0.0: npm run test fails with linting errors


maxmantz:
I'm getting an error message when running `npm run test` on Windows 10. 
Everything apart from the tests works fine (start, build, serve etc.)...

maxmantz:
Since attaching the log file didn't work here it is in copy & paste:

0 info it worked if it ends with ok
1 verbose cli [ 'C:\\Program Files (x86)\\nodejs\\node.exe',
1 verbose cli   'C:\\Users\\Maxe\\AppData\\Roaming\\npm\\node_modules\\npm\\bin\\npm-cli.js',
1 verbose cli   'run',
1 verbose cli   'test' ]
2 info using npm@3.8.1
3 info using node@v4.4.0
4 verbose run-script [ 'pretest', 'test', 'posttest' ]
5 info lifecycle reactjs-boilerplate@2.4.0~pretest: reactjs-boilerplate@2.4.0
6 verbose lifecycle reactjs-boilerplate@2.4.0~pretest: unsafe-perm in lifecycle true
[...]

PR #252, opened 2016-03-30
Added line ending settings for *.js files in .gitattributes

maxmantz:
Added `eol=lf` to *.js files in .gitattributes, see issue #249


mxstbr:
Thank you so much @maxmantz!

PR #258, opened 2016-04-06
Fixes #223: `build:clean` task added `prebuild` task

### LINKED ISSUE #223: Explain how to remove stuff
Keeping a focus on staying flexible to changes at the foundational level, as long as one doesn't compromise on anything else, is a really important trait this boilerplate has/should have. To do that, we should add more documentation on how to **remove** stuff.

I've added it in a few places, but it's scattered, incomplete and hard to find. We need to elaborate a bit more on the JS documentation, right now it's a lot less detailed than everything else. Splitting it up into the different parts of the JS like the CSS docs (e.g. a `reselect` doc, an `immutablejs` doc,...) and then adding sections to each of the parts on how to remove the stuff would be really helpful.

We could then maybe add a doc to the `general/` folder which is a collection of links to sections on how to remove parts of the boilerplate.

oliverturner:
Uses rimraf for cross-platform consistency

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
   },
   "scripts": {
     "prebuild": "npm run test && npm run build:clean",
-    "build:clean": "rimraf ./build",
+    "build:clean": "rimraf ./build/*",
     "build": "cross-env NODE_ENV=production webpack --config internals/webpack/webpack.prod.babel.js --colors -p",
     "start": "npm run start:tunnel",
     "start:tunnel": "NODE_ENV=development node server",
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "postcss-reporter": "^1.3.0",
     "psi": "^2.0.2",
     "react-addons-test-utils": "^0.14.6",
+    "rimraf": "^2.5.2",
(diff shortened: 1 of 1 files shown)


mxstbr:
Thanks so much! I like the `build:clean` and `prebuild` script approach, very clean :+1:

This works fine, but isn't very nice – the build folder disappears completely while the application is building.

The way I would love to see this working, but don't know if that's possible, is that the build folder stays as-is until the build is complete. When the build is complete, the build folder is emptied (but the folder persists) and the new files are put in immediately afterwards.

That is imo the nicest UX, as it seems as if the files were simply swapped out, and if the build fails, the old, working files are still there.

Since I don't know how one would go about telling webpack to do that, deleting the *contents* of the build folder before building is also okay, but I'd prefer the first approach. (and better than having the entire folder disappear while we're building)

oliverturner:
Since the tests run first and `build:clean` only runs if they pass, is that an acceptable compromise?

(I've changed the path arg so that the build folder itself is preserved across builds)

mxstbr:
Ah yes, that's a good point! This looks good to me, thanks!

PR #261, opened 2016-04-06
Add npm run analyze command

oliverturner:
Not as elegant as it might be, but generates a `stats.json` from a productionised build and opens a browser to [the webpack analyser](http://webpack.github.io/analyse/) ready for the user to upload. 

Needs some instructions output to the terminal on completion...


mxstbr:
This is good first step, but I'd make this a script and log some output after finishing the build. Can take a look at this!

mxstbr:
Just took a look, is there a reason we aren't using https://github.com/chrisbateman/webpack-visualizer?

Since it's a webpack plugin, instead of having to upload a huge JSON file somewhere it generates a visualization `.html` file one can simply open locally in the browser. I think that's preferable, especially for bigger applications, but I might be missing something?

oliverturner:
Well, I'm always looking for ways to avoid thrusting yet another dependency on users, and it's not the actually the case that you upload anything (since the code runs client-side) but mainly it's that it doesn't give the level of insight that the Webpack analyzer gives.

webdesserts:
I will say that I've had to setup the project multiple times in the last few days and the `npm install`s sometimes run ~5minutes at times. Since this isn't exactly a need to have feature, reduced dependencies would be a nice bonus from my point of view.

mxstbr:
Fair enough, I'm convinced! I'll add a bit of documentation about this feature in the commands.md and merge it in!

mxstbr:
Alright, I think this is ready to merge – I'd appreciate if somebody could try running it on Windows (we should highlight our amazing OS compatibility in the README btw @oliverturner) but I'm 99% sure it works. I gave it nicer output and instead of opening the link in the default browser (which I think only works on OSX) I output the link like so:

screen shot 2016-04-13 at 11 20 09

*I'm aware the keyboard-shortcut-trick only works on OSX in the default terminal, but it's CMD-Click in iTerm so that'll work too and on Windows I'm pretty sure people will figure it out themselves*

PR #267, opened 2016-04-07
Feature/docs3

oliverturner:
Moar docs


mxstbr:
LOVE IT :heart:

The changelog is A+, and the files explanation is so much better than before. Huge :+1:

PR #270, opened 2016-04-08
Yet moar tweaks

oliverturner:
- Added a note on stylelint 
- Made the argument against Sass a bit more coherent
- Moved CHANGELOG.md into the root

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,20 +4,20 @@
 
 React Boilerplate (RBP) v3.0.0 is out, and it's a _complete_ rewrite! :tada:
 
-We've focussed on becoming a rock-solid foundation to start your next project
+We've focused on becoming a rock-solid foundation to start your next project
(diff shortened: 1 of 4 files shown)


mxstbr:
I think CHANGELOG should stay in the docs, probably in the docs root folder. Having that in the general root folder isn't necessary since nobody upgrades a boilerplate. :+1: 

Other than that love it!

oliverturner:
Couple of points there...
- [most](https://github.com/facebook/react) [projects](https://github.com/stylelint/stylelint) [have](https://github.com/postcss/postcss) [it](https://github.com/reactjs/redux) [there](https://github.com/babel/babel)
- It may be true that folks don't upgrade an installed boilerplate, but people who are taking a second look definitely will

mxstbr:
Fair enough, I'll take it. Awesome work :sparkles:

Issue #275, opened 2016-04-11
[v3.0.0] Page does not maintain scroll postion after css changes


webdesserts:
One thing that I noticed while working with the boilerplate's hot reloading is that when editing anything in the css the page always scrolls back to the top. This may seem like a minor annoyance, but when working on the footer of a long website (think mobile site where everything is vertical) this will severely prolong the feedback loop.

oliverturner:
Sounds like a job for https://github.com/taion/scroll-behavior

mxstbr:
Can you try removing the `useScroll` call here and see if that fixes the issue? https://github.com/mxstbr/react-boilerplate/blob/v3.0.0/app/app.js#L66

webdesserts:
@mxstbr I used that on the react-boilerplate-site and it fixed the issue 👍

mxstbr:
Awesome!

webdesserts:
@mxstbr Do you want me to make a PR for this or does that function still serve some use?

PR #277, opened 2016-04-12
Use Standard Scroll

webdesserts:
Swapped `useScrollToTop` to `useStandardScroll` in response to #275. Feel free to close if the current behavior is desired. Was there a particular reason we were scrolling to top each time before?

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -20,7 +20,7 @@ import { Provider } from 'react-redux';
 import { Router, browserHistory } from 'react-router';
 import { syncHistoryWithStore } from 'react-router-redux';
 import FontFaceObserver from 'fontfaceobserver';
-import useScroll from 'scroll-behavior/lib/useScrollToTop';
+import useScroll from 'scroll-behavior/lib/useStandardScroll';
 import configureStore from './store';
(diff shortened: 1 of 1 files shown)


mxstbr:
See #164 and the related issue #156. The reason that is in there is because otherwise the scroll position persists when we open a new route, which is highly undesirable, but I'm not sure how we can disable that for **just** hot reloaded changes. Any ideas?

webdesserts:
@mxstbr So I was thinking about this yesterday and I agree, from a user standpoint this is undesirable. What I don't understand is why a history change would be triggered if the route doesn't change. If there was a way to test whether the route changed, and only scroll to top when it does, that would be ideal I think.

mxstbr:
Maybe something like this could work:

```JS
history.listen(() => { scrollToTop() }); // scrollToTop being a function that scrolls the body to the top instantly. Left as an exercise to the reader. :wink:
```

mxstbr:
I'll close this for now since it seems like a minor bug and I'm not willing to invest the time to investigate, I'd still love to get it fixed eventually!

PR #279, opened 2016-04-13
Describe how to use syncHistoryWithStore without reselect

JCB-K:
as discussed in https://github.com/mxstbr/react-boilerplate/issues/272#issuecomment-209386771


mxstbr:
Awesome, thanks so much!

PR #286, opened 2016-04-18
True should be a keyword, not a variable

JCB-K:
title says it all


mxstbr:
Well, you'll need to fix the test too... (which is now failing :wink:)

JCB-K:
@mxstbr what are you talking about? 😇 (Fixed now, amended my commit!)

mxstbr:
Awesome, thanks so much!!

PR #287, opened 2016-04-19
Centralising addition of sanitize.css via webpack config

oliverturner:
If we want santitize.css to be a centralised addition then this is an alternative approach to consider.

### DIFF
--- a/app/containers/App/index.js
+++ b/app/containers/App/index.js
@@ -8,9 +8,6 @@
 
 import React from 'react';
 
-// Import the CSS reset, which HtmlWebpackPlugin transfers to the build folder
-import 'sanitize.css/sanitize.css';
-
 import Img from 'Img';
 import Footer from 'Footer';
 import Banner from './banner-metal.jpg';
--- a/internals/webpack/webpack.dev.babel.js
+++ b/internals/webpack/webpack.dev.babel.js
@@ -16,6 +16,7 @@ module.exports = require('./webpack.base.babel')({
   entry: [
     'eventsource-polyfill', // Necessary for hot reloading with IE
     'webpack-hot-middleware/client',
+    'sanitize.css/sanitize.css', // Baseline styles
     path.join(process.cwd(), 'app/app.js'), // Start with js/app.js
   ],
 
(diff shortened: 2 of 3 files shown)


novascreen:
I'm playing around with sanitize.css at the moment as well. I might be missing something really obvious, but i was simply trying to change the background color of `:root` in my `App/styles.css` and `:root` from sanitize takes precedence. So i though i can just change the `--root-background-color` property, but in the end i only got that working by using postcss-import and doing something like this in my `App/styles.css`:
```
/* Note that i'm loading the version from lib that still has the custom properties */
@import 'sanitize.css/lib/sanitize.css';

:root {
  --root-background-color: #262628;
}
```
Sanitize.css clearly deals in global styles, so where would be a good place to change the variables it uses and how to achieve this in a clean way?

oliverturner:
Hey @novascreen let me get right back to you: this is something we're actively working on right now

novascreen:
Great, i'll leave this for the moment then, thanks @oliverturner 
Working on my first React/Redux project at the moment and this boilerplate is just great. I'm kind of rolling a mix of 2.x and 3.x to not make the first project too overwhelming.

grabbou:
+1 to @novascreen, I am doing the same with my sanitize.css - by importing already compiled version we kinda resign from all its benefits automatically

oliverturner:
I'm not speaking for the project here, but here's how I'm doing things on a project that uses sanitize:

1) Import sanitize via webpack: this means that any additional entrypoints will automatically inherit its rules and any shared styles will be extracted into `common.css` on compilation
```javascript
// webpack.config.babel.js
module.exports = {  
  entry: [
    'sanitize.css/lib/sanitize.css',
    './src/index.jsx'
  ],
  ...
}
```

2) Override the sanitize.css defaults. Here, `--root-font-family` and `--root-background-color`:
```css
/* src/styles.pcss */
:root {
  font-family:      "Helvetica Neue", Helvetica, Arial, sans-serif;
  background-color: #999;
}
```

3) Import the main CSS for the entrypoint
```javascript
// src/index.jsx
import './styles.pcss'
```

mxstbr:
I'll close this PR since we won't merge it. It's an awesome idea, but not suited to be included by default—a very nice reference implementation though!

PR #292, opened 2016-04-26
Import PropTypes from react

Lesuk:
Import PropTypes from react like in A component.

### DIFF
--- a/app/components/Button/index.js
+++ b/app/components/Button/index.js
@@ -6,7 +6,7 @@
  * otherwise it'll render a link with an onclick
  */
 
-import React from 'react';
+import React, { PropTypes } from 'react';
 
 import styles from './styles.css';
 
@@ -33,10 +33,10 @@ function Button(props) {
 }
 
 Button.propTypes = {
-  className: React.PropTypes.string,
-  handleRoute: React.PropTypes.func,
-  href: React.PropTypes.string,
-  onClick: React.PropTypes.func,
+  className: PropTypes.string,
+  handleRoute: PropTypes.func,
+  href: PropTypes.string,
+  onClick: PropTypes.func,
 };
 
 export default Button;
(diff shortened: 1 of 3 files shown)


mxstbr:
It appears that this change (for some unknown reason) decreases test coverage, which is not desireable. Any ideas how to fix that? (I like your style a lot more!)

Lesuk:
No, I just saw the difference

maxmantz:
I guess has to do with the fact that the prop types aren't covered in tests. You would have to test each components prop types like this:

instead of `Button.propTypes` try the following:

    export const propTypes = {
       // prop types here...
    };

    Button.propTypes = propTypes;

Then in tests:

    import { propTypes } from './index';

    // ...
    expect(renderedComponent.propTypes).toEqual(propTypes);

justingreenberg:
alternatively, proptypes can be destructured.... ie

```js
import React from 'react'

const Img = ({ className, src, alt }) =>
  <img className={className} src={src} alt={alt} />

const { PropTypes: { string } } = React

Img.propTypes = {
  src: string.isRequired,
  alt: string.isRequired,
  className: string,
}

export default Img
```

allows clean namespace import of react and makes adding new props painless

can also be used for React.Component...

```js
const { PropTypes: { string }, Component } = React

class Img extends Component {
  static propTypes = {
      src: string.isRequired,
      //...
  }
}
```

oliverturner:
2 questions: 
1. Doesn't this simply highlight the arbitrary nature of the coverage tests?
2. I like the cleanliness of the approach, but is it worth investing time in managing proptypes when Flow will inevitably replace them? If it is, isn't something like https://www.npmjs.com/package/babel-plugin-flow-react-proptypes a more forward-compatible approach?

mxstbr:
Of course coverage is completely arbitrary, I was mostly kidding. Sorry for not merging this earlier, I got sidetracked with other things.

Would you mind resolving the conflicts @Lesuk and then I'll land this? I really like this pattern!

mxstbr:
Realised I could just as well solve that tiny merge conflict myself. Thanks so much @Lesuk! :tada:

PR #296, opened 2016-05-02
Fix run clean script and template

xoxoxo:
Atm after running npm run clean boilerplate is broken. This pr fixes it.

### DIFF
--- a/internals/scripts/clean.js
+++ b/internals/scripts/clean.js
@@ -27,10 +27,10 @@ rm('-rf', 'app/components/*');
 // Cleanup containers folder
 rm('-rf', 'app/containers/*');
 mkdir('app/containers/App');
-mkdir('app/components/NotFoundPage');
+mkdir('app/containers/NotFoundPage');
 mkdir('app/components/HomePage');
 cp('internals/templates/appContainer.js', 'app/containers/App/index.js');
-cp('internals/templates/notFoundPage.js', 'app/components/NotFoundPage/index.js');
+cp('internals/templates/notFoundPage.js', 'app/containers/NotFoundPage/index.js');
 cp('internals/templates/homePage.js', 'app/components/HomePage/index.js');
 
 // Cleanup sagas folder
--- a/internals/templates/routes.js
+++ b/internals/templates/routes.js
@@ -23,14 +23,14 @@ export default function createRoutes(store) { // eslint-disable-line
     {
       path: '/',
       getComponent(location, cb) {
-        System.import('HomePage')
+        System.import('containers/App')
           .then(loadModule(cb))
           .catch(errorLoading);
       },
     }, {
       path: '*',
       getComponent(location, cb) {
-        System.import('NotFoundPage')
+        System.import('containers/NotFoundPage')
           .then(loadModule(cb))
           .catch(errorLoading);
       },


mxstbr:
Very good catch, thanks so much!

sbraconnier:
Since this merge it seems that after the example app is deleted (npm run clean), the HomePage container is no longer used. Not sure why:
`System.import('HomePage')`
has been fixed to
`System.import('containers/App')`
instead of
`System.import('components/HomePage')`
in the internals/templates/routes.js file.

Talking about the HomePage. Is there a reason why the HomeHage is copied into the components directory while the NotFoundPage is copied into the containers directory ? In the example app, the HomePage lives into the containers directory.

Thus, the fix would be:
`System.import('containers/HomePage')`
in the internals/templates/routes.js file and
```
mkdir('app/containers/HomePage');
...
cp('internals/templates/homePage.js', 'app/containers/HomePage/index.js');
```
in the internals/scripts/clean.js script.

Thanks.

mxstbr:
Good catch, would you mind submitting a pull request?

xoxoxo:
Yes I missed that. My question would be if we even need HompePage component/container since we use App as a main route? This should be dev version and I guess if someone needs a hompage can easily generate one.

sbraconnier:
It would be an honor to submit my first pull request ever to this great project :) Just let me know if you prefer to just remove the HomePage or to fix the route in the internals/templates/routes.js file. And if the HomePage stays, should it be as a container, or a component ? I suspect what the answer is but I still wear diapers with all this stuff (react, redux, webpack, node, etc., even github) so a confimation would be welcome.

mxstbr:
@xoxoxo the App component is a wrapper around all pages, the HomePage has to be it's own component!

@sbraconnier The HomePage should be a component in the cleaned version, since it is not connected to the state!

xoxoxo:
Oh yes you are right my bad but have and idea maybe we should call it something like Layout? @mxstbr

sbraconnier:
@mxstbr  see, I didn't suspect the right answer :). I knew that in the cleaned version, the HomePage doesn't know about redux but I though that it was assumed that it would be connected to the state someday. Following this path, isn't the same for the NotFoundPage, which is a container in the cleaned version? Sorry if I missed something. Maybe the NotFoundPage should be a component too ?

mxstbr:
Exactly, the `NotFoundPage` component should be component too! :+1:

xoxoxo:
Why its not a component in example app then? I think that it could confuse some people as we have 2 different setups (it did me).

Correct me but if aim wrong
App - works as a Layout/wrapper
HomePage - is for example our index page
NotFound is just another page

this is how it looks like atm in example app, so why make it different? In components we have only things like H1, A, Img etc. I ma new to react too, so just trying to get most out of it.

mxstbr:
If you take a look, it's connected to the state: https://github.com/mxstbr/react-boilerplate/blob/v3.0.0/app/containers/NotFoundPage/index.js#L41

Which the cleaned version isn't, so it's not a container :blush:

xoxoxo:
Cheers! React ain't for dummies :D

mxstbr:
No worries!

sbraconnier:
Great it is clear now. NotFoundPage and HomePage are dumb components in the cleaned version and smart components in the example app.

I'll read all the necessary doc in order to contribute to the project and submit the pull request later today :)

Issue #281, opened 2016-04-13
3.0.0 Remove a generated resource


mariusleu:
Using `npm run generate <resource>` to generate components, containers, routes etc. is a  very great and welcome feature.

But how can we remove a generated resource after we generate it? Maybe we generate it misspelled or things like this.

Kuzmin:
Most resources are isolated to their own directory, so you should be able to just delete them with `rm`. If you're using Git it's also possible to reset to `HEAD` which will have the side effect of deleting all code changes since the last commit. 

I think these two should be able to provide you with what you need, since the tool used by `npm run generate`, (https://github.com/amwmedia/plop)[plop], doesn't support undo.

mxstbr:
Exactly what @Kuzmin said, thanks!

PR #299, opened 2016-05-04
V3.0.0 - Async sagas

justingreenberg:
also a little cleanup ie assignment destructuring promise.all, named routes, etc...

edit: not sure if this is something you were interested in adding, but figured i'd open the PR :)


app/routes.js
```diff
@@ -16,27 +16,36 @@ function loadReducer(store, name) {
...
         Promise.all([
           System.import('containers/HomePage/reducer'),
           System.import('containers/HomePage'),
-        ]).then(modules => {
-          loadReducer(store, 'home')(modules[0]);
-          loadModule(cb)(modules[1]);
+          System.import('sagas/getGithubData.saga'),
+        ]).then(([reducer, component, saga]) => {
+          loadReducer(store, 'home')(reducer);
+          loadModule(cb)(component);
```
fubhy: Load the Saga first, then run the Callback after loading the module

justingreenberg: good catch

justingreenberg:
maybe colocate async sagas with containers, similar to reducers?

```js
import { getGithubDataSaga } from `containers/HomePage/sagas.js`
```

sagas.js:

```js
export function* getGithubDataSaga() { //...
```

mxstbr:
Yes!! This is absolutely amazing, A+ works! Let's get rid of that saga folder and have sage per container too! :+1:

internals/scripts/clean.js
```diff
@@ -27,10 +27,10 @@ rm('-rf', 'app/components/*');
 // Cleanup containers folder
 rm('-rf', 'app/containers/*');
 mkdir('app/containers/App');
-mkdir('app/components/NotFoundPage');
+mkdir('app/containers/NotFoundPage');
 mkdir('app/components/HomePage');
 cp('internals/templates/appContainer.js', 'app/containers/App/index.js');
-cp('internals/templates/notFoundPage.js', 'app/components/NotFoundPage/index.js');
+cp('internals/templates/notFoundPage.js', 'app/containers/NotFoundPage/index.js');
```
mxstbr: This was fixed in #281, we can revert this change!

mxstbr:
We'll also need to expand the route generator to generate the saga loading bit if a saga is present, see https://github.com/mxstbr/react-boilerplate/pull/294/files

app/routes.js
```diff
@@ -16,27 +16,36 @@ function loadReducer(store, name) {
   return (module) => injectAsyncReducer(store, name, module.default);
 }
 
+function loadSaga(store) {
+  return (module) => injectAsyncSaga(store, module.default);
```
justingreenberg: also, it may be a good idea to change this name since `module` is a global in node..

justingreenberg:
@mxstbr rebased, let me know what you think! i updated the syntax to be a little more declarative since this is new to most people... i also left some scaffolding in root saga folder, since most people will want at least one root daemon. i am swamped with work so i wouldn't have the time to update the corresponding generator until next week, if anybody else is willing to jump in sooner that would be awesome :)

ralomucho:
Amazing !

mxstbr:
This looks amazing from a first glance, sorry for the late reply. I'll check this out asap, thanks so much!

ralomucho:
Just a few remarks

-  In store.js, why putting a hook on store to run saga ? Doing that, you're changing normal store API and you can still directly use sagaMiddleware.run rather than store.runSaga in injectAsyncSagas
- Looked at whatcher.js. I think it's more appropriate tu use a dedicated middleware (like [redux logger](https://github.com/theaqua/redux-logger)) which directly catch any action call. I don't think catching calls to be considered as an async operation.

By the way, thank you for your proposal. All of this splitting stuff and async loads is very interesting.

ralomucho:
Just tried your work by removing you store reference in injectAsyncSagas. Seems to work ok.

ralomucho:
Trying to do a updated generator

ralomucho:
@justingreenberg made a new PR in your repo with new routeWithReducer.hbs

ralomucho:
Sorry but this new template doesn't use your  store reference in injectAsyncSagas, so add it if you really sure to keep it.

justingreenberg:
@ralomucho thank you!! to address some of your questions...

> In store.js, why putting a hook on store to run saga ? Doing that, you're changing normal store API and you can still directly use sagaMiddleware.run rather than store.runSaga in injectAsyncSagas

we are not changing the store api, we are just enhancing it! the reason for creating a hook is so that we have access to saga middleware directly on store for sync use (ie for on server, middleware, etc)... previously runSaga was exposed directly, so this pattern is a result of recent changes to the redux-saga api... best practices are definitely still emerging, but you can see this being used in the official examples:

note this example also enhances the store with a `close` method for use in SSR....
https://github.com/yelouafi/redux-saga/blob/master/examples/real-world/store/configureStore.dev.js#L31

basic sync runSaga usage...
https://github.com/yelouafi/redux-saga/blob/master/examples/async/src/main.js#L10

for our use case (async injection) having a facade to access hook on store is the cleanest solution imo

> Looked at whatcher.js. I think it's more appropriate tu use a dedicated middleware (like redux logger) which directly catch any action call. I don't think catching calls to be considered as an async operation.

not async but it's an example of isolating side-effects, which i think is a very important concept... obviously in this case the watcher isn't actually doing anything, but i left it there as a reference for setting up root sagas and basic concurrency (fork)... @mxstbr let me know what you think when you have a chance to review 😄

justingreenberg:
@ralomucho also, just to clarify regarding the watcher, it wasn't meant to be a real logger at all! just a demo of running a root saga with basic concurrency

mxstbr:
Alright, finally got the time to take a look at this! It looks amazingly solid, very good work @justingreenberg and great critique @ralomucho 

1. I don't think having the `saga/` folder in the root `app/` folder is necessary by default. I agree that it's useful for the watching use case, but not many applications will have a need for that, so I'd rather not include it by default. (re-including it isn't that big of an effort anyway)
This also means that the saga generator should be removed again for now, we can look into generating sagas for a specific container down the line.

2. I love the way you handle asynchronous saga injection, big :clap: 

3. When generating a new container, the question if they want to generate a new saga should be separate. One can have a saga for a container without having an actions/constants/reducer tupel, and the other way around too.

Again, this is an amazing PR, thanks so much for taking the time to submit it! :+1:

*Ping @neuling, who uses the saga-per-container pattern in his app. Would you mind taking a look at this?*

app/store.js
```diff
@@ -6,27 +6,34 @@ import { createStore, applyMiddleware, compose } from 'redux';
...
+  const enhancers = [
+    applyMiddleware(...middlewares),
+    devtools(),
+  ];
+
+  const store = createStore(
+    createReducer(),
+    fromJS(initialState),
+    compose(...enhancers)
+  );
```
justingreenberg: refactored to use latest syntax

app/routes.js
```diff
@@ -13,6 +13,9 @@ const loadModule = (cb) => (componentModule) => {
 };
 
 export default function createRoutes(store) {
+  // create reusable async injectors using getHooks factory
+  const { injectReducer, injectSagas } = getHooks(store);
```
justingreenberg: curried helpers remove the need to pass store to helpers in every callback

justingreenberg:
ok, i have no idea why this CI is failing—tests work locally and the test that is failing is [HomeView error state](https://travis-ci.org/mxstbr/react-boilerplate/builds/129913759#L1915) on freakin IE11... 

```
FAILED TESTS:
  <HomePage />
    × should render an error if loading failed
      IE 11.0.0 (Windows 8.1 0.0.0)
    Error: ShallowWrapper::contains() can only be called with ReactElement (or array of them), string or number as argument.
       at contains (C:/projects/react-boilerplate/internals/testing/test-bundler.js:54483:12 <- webpack:///~/enzyme/build/ShallowWrapper.js:284:0)
       at Anonymous function (C:/projects/react-boilerplate/internals/testing/test-bundler.js:70296:50 <- webpack:///app/containers/HomePage/index.test.js:9:2700)
```

very weird, i tried a million things to debug. please take a look if you guys have a sec. i'm trying not to let this PR get stale :) i will investigate more tomorrow.

mxstbr:
Thanks, I sadly don't have a Windows machine around…

mxstbr:
@justingreenberg lets get this merged! Update the version of enzyme in the `package.json` to `2.3.0`, and change the failing test in `HomePage/index.test.js` to:

```JS
it('should render an error if loading failed', () => {
    const renderedComponent = mount(
      <HomePage
        loading={false}
        error={{ message: 'Loading failed!' }}
      />
    );
    expect(
      renderedComponent
        .text()
        .indexOf('Something went wrong, please try again!')
      ).toBeGreaterThan(-1);
  });
```

That should fix the issue. (You'll also have to remove a few imports to fix linting errors)

mxstbr:
Fixed this on the main branch, could you rebase to integrate the newest changes, then we'll merge this!

justingreenberg:
@mxstbr no problem, just got out of bed... coffee and rebase! give me about 15 mins :)

mxstbr:
It's fine, I've merged and rebased locally! Give me five minutes and this'll be merged!

mxstbr:
Merged, the eagle has landed! Thanks for the amazing work @justingreenberg!! :tada:

justingreenberg:
thank you!!

PR #298, opened 2016-05-04
Apply changes required in order to make the cleaned version works.

sbraconnier:
My first contribution ever! I tested it and the cleaned app show the HomePage on startup.

PR #300, opened 2016-05-05
Load the HomePage when default route is taken

TheElf:
I think this is what you meant, otherwise the App container gets rendered twice on the homepage


mxstbr:
Didn't #298 fix this?

mxstbr:
Yeah, #298 fixed this, still thanks so much for taking the time to submit this PR, much appreciated!

PR #304, opened 2016-05-08
Update offline-plugin to v3

NekR:
Just released it with bunch of updates and fixes.
See changelog here: https://github.com/NekR/offline-plugin/blob/master/CHANGELOG.md


mxstbr:
Thanks!

PR #305, opened 2016-05-15
RFC: Colocate selectors and reducers

mxstbr:
This RFC gets rid of the `selectors/` folder and adds `selectors.js`/`selectors.test.js` files to the two containers with reducers, `containers/App/` and `containers/HomePage`.

---- 

Inspired by a [Dan Abramov tweetstorm](https://twitter.com/dan_abramov/status/730933179511640064), I think that co-locating selectors with reducers makes a lot of sense.
I don't think that putting them in the same file is necessary, since that file will explode in size and become unmanageable (see: ducks), but having them right next to the reducer makes so much more sense imo than having them in a `selectors/` folder.

Also note that with #299, we would have neither a `sagas/` nor a `selectors/` folder, and I think that's amazing.

**What do you think?**
I like this *a lot*, but I haven't properly worked with it. One thing that's a bit weird is the `locationSelector` that now lives in `containers/App/selectors.js`, but is only used in the main `app.js`. Any ideas?

*Todo before merging: Fix generators/templates, just wanted to get this RFC out.*


ralomucho:
In my own App I changed selector names : for instance, loadingSelector becomes  selectLoading which sounds like a getter while still keeping the idea of selectors which can be composed. Just my personnal feeling.

oliverturner:
Solid thumbs-up from me.

nikgraf:
Nice change! 
I really like the idea of 'selectSomething' compared to 'somethingSelector'.

In the apps that I built we had selectors that have been used by multiple containers/selectors. Where would you recommend to put shared selectors?

ralomucho:
about locationSelector:
routing is something done globally so putting  locationSelector in App would not bothers me. App could have two reducers : one for the global domain (if needed) and one for the routing domain.
The fact is that in the example app, App container deals with too much business. Maybe it would have been better to put  everything apart the nav bar in HomePage  and to put HomePage content in a HomePage Child. Like this, App would only deal with global bahaviour.

mxstbr:
@nikgraf one can always create a root `selectors/` folder if one has complex selectors which combine state between different reducers.

@ralomucho I think the App container is fine, I agree that we should actually put the `routeReducer` there too instead of `reducers.js`! Good point!

----

Since there seems to be a consensus that this is a net-positive change, I'll go ahead and implement the generators anew and merge this asap.

ralomucho:
@nikgraf : in my app, I make the distinction between two kinds of selectors :
- directSelector which directly selects a single part of the state and doesn't need all of this boilerplate with createSelector and which can still be composed with other selectors. For instance  I use

```
selectError = () => state.getIn['global', 'error']
```
rather than 

```
const errorSelector = () => createSelector(
    globalSelector(),
    (globalState) => globalState.get('error')
)
```

- computeSelector which composes selectors with `createSelector` to benefit from memoization

These to kinds of selectors have their own generator. The fact is we need sometimes selectors from several containers :

- if  the new selector is only used by one container, we can just make proper imports inside this container directory
- if it is used by several containers, it's clear we nead a shared directory. `sharedSelectors`would be more appropriate to make the distinction with containers own selectors.

mxstbr:
The reasoning is that each selector selects a part of the state, right? So the selector should be right next to the reducer that has the part of the state it selects. This means that when you change the state shape, all you have to change is two files while the rest of the files stay the exact same.

A `sharedSelector/` folder is basically what we just got rid of, it's only useful if you have a multitude of very complex selectors that mix state between reducers but otherwise almost all selectors should be right next to a reducer!

wmertens:
So, about the exploding ducks, I don't see how adding a reducer to a file with actions makes it grow that much bigger. Am I missing something?

mxstbr:
@wmertens I'm not a big fan of the ducks approach, see my comment [here](https://github.com/mxstbr/react-boilerplate/issues/27#issuecomment-168247634) and the followup [here](https://github.com/mxstbr/react-boilerplate/issues/27#issuecomment-168932177).

PR #312, opened 2016-05-17
Added missing loaders for fonts files and svg

jptissot:
While trying to integrate Semantic UI to this boilerplate, I noticed that the webpack config was missing file loaders for font files and svg files used by semantic-ui. I thought it could be usefull to others so I thought I'd share it. 

Found the fix in this thread https://github.com/Semantic-Org/Semantic-UI/issues/3533

Not included in the PR since it's very specific to semantic-ui : 

One other thing I had to do to make it work was to include / exclude semantic css in the `webpack.base.babel.js` file like so : 

```javascript
 {
      // Transform our own .css files with PostCSS and CSS-modules
      test: /\.css$/,
      exclude: [/node_modules/, /semantic/],
      loader: options.cssLoaders,
    }, {
      test: /\.css$/,
      include: [/node_modules/, /semantic/],
      loaders: ['style-loader', 'css-loader'],
    }, 
```

Then in my app.js I imported semantic.css `import '../semantic/dist/semantic.css';`

### DIFF
--- a/internals/webpack/webpack.base.babel.js
+++ b/internals/webpack/webpack.base.babel.js
@@ -34,6 +34,21 @@ module.exports = (options) => ({
     }, {
       test: /\.jpe?g$|\.gif$|\.png$/i,
       loader: 'url-loader?limit=10000',
+    }, { 
+      test: /\.woff(\?v=\d+\.\d+\.\d+)?$/, 
+      loader: 'file?name=fonts/[name].[hash].[ext]&mimetype=application/font-woff' 
+    }, { 
+      test: /\.woff2(\?v=\d+\.\d+\.\d+)?$/, 
+      loader: 'file?name=fonts/[name].[hash].[ext]&mimetype=application/font-woff'
+    }, { 
+      test: /\.ttf(\?v=\d+\.\d+\.\d+)?$/, 
+      loader: 'file?name=fonts/[name].[hash].[ext]&mimetype=application/octet-stream' 
+    }, { 
+      test: /\.eot(\?v=\d+\.\d+\.\d+)?$/, 
+      loader: 'file?name=fonts/[name].[hash].[ext]' 
+    }, { 
+      test: /\.svg(\?v=\d+\.\d+\.\d+)?$/, 
+      loader: 'file?name=images/[name].[hash].[ext]&mimetype=image/svg+xml' 
     }, {
       test: /\.html$/,
       loader: 'html-loader',
--- a/internals/webpack/webpack.base.babel.js
+++ b/internals/webpack/webpack.base.babel.js
@@ -32,23 +32,20 @@ module.exports = (options) => ({
       include: /node_modules/,
       loaders: ['style-loader', 'css-loader'],
     }, {
-      test: /\.jpe?g$|\.gif$|\.png$/i,
+      test: /\.jpe?g$|\.gif$|\.png$|\.svg$/i,
       loader: 'url-loader?limit=10000',
-    }, { 
-      test: /\.woff(\?v=\d+\.\d+\.\d+)?$/, 
-      loader: 'file?name=fonts/[name].[hash].[ext]&mimetype=application/font-woff' 
-    }, { 
-      test: /\.woff2(\?v=\d+\.\d+\.\d+)?$/, 
-      loader: 'file?name=fonts/[name].[hash].[ext]&mimetype=application/font-woff'
-    }, { 
-      test: /\.ttf(\?v=\d+\.\d+\.\d+)?$/, 
-      loader: 'file?name=fonts/[name].[hash].[ext]&mimetype=application/octet-stream' 
-    }, { 
-      test: /\.eot(\?v=\d+\.\d+\.\d+)?$/, 
-      loader: 'file?name=fonts/[name].[hash].[ext]' 
-    }, { 
-      test: /\.svg(\?v=\d+\.\d+\.\d+)?$/, 
-      loader: 'file?name=images/[name].[hash].[ext]&mimetype=image/svg+xml' 
+    }, {
+      test: /\.woff(\?v=\d+\.\d+\.\d+)?$/,
+      loader: 'file?name=fonts/[name].[hash].[ext]&mimetype=application/font-woff',
+    }, {
+      test: /\.woff2(\?v=\d+\.\d+\.\d+)?$/,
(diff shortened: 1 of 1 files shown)


internals/webpack/webpack.base.babel.js
```diff
@@ -34,6 +34,21 @@ module.exports = (options) => ({
...
+      loader: 'file?name=fonts/[name].[hash].[ext]&mimetype=application/font-woff'
+    }, { 
+      test: /\.ttf(\?v=\d+\.\d+\.\d+)?$/, 
+      loader: 'file?name=fonts/[name].[hash].[ext]&mimetype=application/octet-stream' 
+    }, { 
+      test: /\.eot(\?v=\d+\.\d+\.\d+)?$/, 
+      loader: 'file?name=fonts/[name].[hash].[ext]' 
+    }, { 
+      test: /\.svg(\?v=\d+\.\d+\.\d+)?$/, 
+      loader: 'file?name=images/[name].[hash].[ext]&mimetype=image/svg+xml'
```
mxstbr: Instead of already using the file loader, add `.svg` the the `.jpg|.gif|.png` loader above so it also uses the `url-loader`!

mxstbr:
This is a very useful addition, good catch! Would you mind fixing the one small nitpick and the linting errors? Thanks so much!

jptissot:
Done. 

Can you think of a cleaner way of excluding / including the semantic css from the css loaders ? I wonder if we could add some kind of folder where global styles can live in ?

mxstbr:
Awesome, thanks so much!

mxstbr:
> Can you think of a cleaner way of excluding / including the semantic css from the css loaders ?

No, you did it perfectly right. :+1: 

> I wonder if we could add some kind of folder where global styles can live in ?

You can definitely do that, but I don't think the boilerplate should have one by default since CSS modules makes that redundant! :grin:

LukeAskew:
I'm not sure this actually works.

I added some fonts to `app/fonts` then updated `app/containers/App/styles.css` to include:

```css
@font-face {
  font-family: Qanelas;
  src: url('/fonts/QanelasSoft-ExtraBold.eot');
  src: url('/fonts/QanelasSoft-ExtraBold.eot?#iefix') format('embedded-opentype');
  src: url('/fonts/QanelasSoft-ExtraBold.woff2') format('woff2');
  src: url('/fonts/QanelasSoft-ExtraBold.woff') format('woff');
  src: url('/fonts/QanelasSoft-ExtraBold.ttf') format('truetype');
  font-weight: 800;
  font-style: normal;
}
```

The fonts don't get picked up by webpack. They are not bundled to `build` on production builds, and I get the following error in dev:

```
Failed to decode downloaded font: http://localhost:3000/
OTS parsing error: invalid version tag
```

mxstbr:
Hmm, I've never seen that error before. Possibly because we're adding a hash to it?

LukeAskew:
It's an issue with css-loader https://github.com/webpack/css-loader/issues/232

Because styles with sourcemaps are loaded as a blob, the urls resolve relative to the `blob://` url.

It works fine with sourcemaps turned off.

Still looking for a viable solution.

kencanak:
not sure if anyone encounter this, but for custom local font usage, when we run npm run build, it will throw error for missing loaders for font files. turns out, the loaders for font files need to be added into webpack.test.babel.js as well

KarandikarMihir:
I'm getting an error for *.woff2 files. Other files (ttf, eot, woff) are loaded just fine. How do I fix this?

gihrig:
> I'm getting an error for *.woff2 files.

@KarandikarMihir Post the error please.

KarandikarMihir:
@gihrig `https://wpd-web/e6cf7c6ec7c2d6f670ae9d762604cb0b.woff2 Failed to load resource: the server responded with a status of 404 (Not Found)`

The aforementioned error occurs in production only, disappears on dev server. The file exists in the build folder, with same name.

gihrig:
Just a wild guess, but what happens if you rename .woff2 to .woff, Same error?

KarandikarMihir:
@gihrig error was gone after renaming. But does that solve the issue? I'm not sure.. I don't see any visible change because there has to be some fallback mechanism in the css framework.

gihrig:
> there has to be some fallback mechanism in the css framework.

I think you're onto something here. 

.woff in place of .woff2 might be a work around, but definitely not  a fix.

I searched the project for `woff` and `woff2`. The only place woff2 is not present with woff is in `.gitattributes` which is related to [git diff](https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes#Binary-Files). So I'm thinking it might be something in webpack's `file-loader`. See `internals/webpack/webpack.base.babel.js` line 35.

Beyond that, I'm out of clues 😞 

Perhaps @mxstbr knows more?

KarandikarMihir:
@gihrig Thanks for skimming through the files! I'm looking if there's any issue with the server or firewall while serving this file. That could also be the reason, right?

Also, after googling a bit, I found that woff2 has a mime type that is not supported by IIS server, by default. So I explicitly added that mime-type in the web.config file. But alas, no joy. Thanks for the help again. I'll comment again if I find anything.

prateekrastogi:
@jptissot  Thanks for your contribution. It saved a lot of pain for me.

jeanmichelcote:
Using latest version of React-Boilerplate,  can't load local fonts:
`You may need an appropriate loader to handle this file type.`

Its a `ttf` type loaded in `container/App/styles.css` like so:
```
@font-face {
  font-family: 'PTC55F';
  src: url('assets/fonts/PTC55F.ttf') format('truetype');
  font-weight: normal;
  font-style: normal;
}

@font-face {
  font-family: 'PTC75F';
  src: url('assets/fonts/PTC75F.ttf') format('truetype');
  font-weight: bold;
  font-style: bold;
}
```

Loader is :
```
{
      test: /\.(eot|svg|ttf|woff|woff2)$/,
      loader: 'file-loader',
    }
```

Anybody has any solution for this?

fabienheureux:
Did you try adding your fonts to app/global-styles.js ?

karna41317:
@fabienheureux Yes, Its in app/global-styles.js ?, Still have issue.

fabienheureux:
@karna41317 do you have some informations about what you are trying to achieve ? It doesn't seem that you provide details about your setup/specific issue in this thread

PR #314, opened 2016-05-18
Add meaning of "DX". Fix #313

### LINKED ISSUE #313: [Docs] What is DX?
The term "DX" is used in the README, but it is not explained anywhere. Perhaps draw from how this project explains it https://github.com/skidding/cosmos

curran:
The term "DX" is used in the README, but it is not explained anywhere. This PR adds "DX (Developer eXperience)", so readers unfamiliar with the term will immediately understand its meaning.

https://github.com/mxstbr/react-boilerplate/issues/313

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@
   <dd>Create components, containers, routes, selectors and sagas - and their tests - right from the CLI!</dd>
 
   <dt>Instant feedback</dt>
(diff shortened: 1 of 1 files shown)


mxstbr:
Ah, good catch, awesome, thanks!

curran:
Thanks!

PR #323, opened 2016-05-19
Fix injector method names in route generator

justingreenberg:


mxstbr:
Ah, good catch Justin, thanks!

PR #326, opened 2016-05-19
Adds "Linux" to line 18 of docs/general/faq.md 

brianboyko:
Because the process for killing an existing server is identical in Linux and OSX, this change adds " / Linux" to the subheader in this section.

(A small change, but I felt it was worth a note.)


mxstbr:
Yeah, nice one!

PR #327, opened 2016-05-19
When generate component, move index.test.js to tests/ folder

lyuehh:


mxstbr:
Awesome, good catch, thanks!

PR #330, opened 2016-05-19
Update base webpack config with Json loader

ralomucho:
I had a problem with some packages (same error as [here](https://github.com/webpack/webpack/issues/592)). I got to do this fix to get everything work. Maybe It could help others.


mxstbr:
Looks good to me, thanks @ralomucho!

zanedev:
I seem to be still getting these errors after pulling master and doing a fresh npm install.  The json loader is in base webpack config so not sure why it's not pulling it in.  My app seems to work fine without it for now..

PR #352, opened 2016-05-21
Fix promise Error when creating route to a container

n4cr:


mxstbr:
Awesome, thanks so much @n4cr!

PR #359, opened 2016-05-21
Users don't need to download whole git history

alexeyraspopov:


mxstbr:
This currently brings 0.9MB of savings:

```Sh
➜  rb-without-depth git:(master) du -sh .git
2.2M	.git
➜  rb-with-depth git:(master) du -sh .git
1.3M	.git
```

I'm not yet sure I'm a fan of this, as it might potentially confuse newcomers to the whole git thing. ("What is that weird `--depth` thing, I want the whole repo not just the root folders!") What do you think?

alexeyraspopov:
> This currently brings 0.9MB of savings:

And clean history for those who'll probably push the result to the different origin.

> I'm not yet sure I'm a fan of this, as it might potentially confuse newcomers to the whole git thing. ("What is that weird --depth thing, I want the whole repo not just the root folders!") What do you think?

I was assuming that people who know about React know about git. `--depth` in context of `clone` seems to be obvious, doesn't it?

mxstbr:
> And clean history for those who'll probably push the result to the different origin.

Just like `npm run setup` :wink:

> --depth in context of clone seems to be obvious, doesn't it?

I distinctly remember being confused the first time I saw it, but maybe that's just me. I'll leave this open for a few hours, if nobody else has that problem I'll merge. 

Thanks for submitting the PR btw!

mxstbr:
Thanks!

PR #360, opened 2016-05-21
Use babel-preset-react-optimize in production

### LINKED ISSUE #351: Implement babel-react-optimize
See https://github.com/thejameskyle/babel-react-optimize and #350!

Aweary:
Resolves #351

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,11 +49,7 @@
     "env": {
       "production": {
         "only": ["app"],
-        "plugins": [
-          "transform-react-remove-prop-types",
-          "transform-react-constant-elements",
-          "transform-react-inline-elements"
-        ]
+        "presets": ["react-optimize"]
       }
     }
   },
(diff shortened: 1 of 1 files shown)


Aweary:
Hmm, this didn't fail locally when I tested. I wish there was a better error message

mxstbr:
This gives an error code 2, which according to the [Node.js docs](https://nodejs.org/api/process.html#process_exit_codes) means "2 - Unused (reserved by Bash for builtin misuse)". 

What the hell is that supposed to mean?

----

Unrelated to the issue, we should remove the old babel transforms from the dependencies!

mxstbr:
Doesn't seem like there's a solution for this yet. 😕 Annoying, as I'd really like to use the preset!

mxstbr:
Closing this for now since it doesn't seem like this is going to be fixable by us. Thanks so much for taking the time to submit this though!

mxstbr:
Ah oops, this actually will be fixed with https://github.com/mxstbr/react-boilerplate/pull/391#issuecomment-222334520, reopening, my bad!

justingreenberg:
@mxstbr @gihrig this can probably be closed, `babel-optimize` has been a source of headaches for many other projects. imo optimizations should really be hand selected based on project.

gihrig:
Agree with @justingreenberg (close).

Github project [babel-preset-react-optimize](https://github.com/thejamekyle/babel-preset-react-optimize/blob/master/packages) is returning 404.

There's [babel-react-optimize](https://github.com/thejameskyle/babel-react-optimize) a very young project by the same author...

Not looking good 👎

mxstbr:
Alright, fair enough. Thanks for chiming in!

Issue #337, opened 2016-05-19
Promise Error when creating route to a container


aslight21:
I'm new to react and I decide to use this boilerplate for my first project. I don't know if this is a bug or if I need to link some things manually, but when I create a container and a route to this new container with the generator, I get an error when I try to access this container.

The error is the following:

```
es6.promise.js:119 Unhandled promise rejection ReferenceError: injectAsyncReducer is not defined
```

EducatedTard:
same here

mxstbr:
That's highly likely a bug with @justingreenberg's latest change. Lets see if we can fix it…

mxstbr:
Ah nevermind, that was actually fixed in #323!

Change the `internals/templates/route/routeWithReducer.hbs` file, replacing `injectAsyncReducer` with `injectReducer` and `injectAsyncSaga` with `injectSaga` and it's going to work again—sorry for the troubles!

aslight21:
This do not correct the problem entirely.

I got injectReducer is not define after the change. I saw that some line in the reducer.js where commented so I delete the // before these lines. This lead me to store is not define. I add store as a parameters to the create routes function and now I have two new errors:

Unhandle promise rejection type error: cannot read proprety equals of undefined 
and
Uncaught type error: reducer is not a function

...

justingreenberg:
@aslight21 do not pass `store` to injectReducer—it is initialized with store in `getHooks`

aslight21:
I don't, I change

export default function createRoutes() { ...
to
export default function createRoutes(store) { ..

justingreenberg:
i'm sorry i'm not sure what you're talking about. with regard to the first error, please see:

https://github.com/mxstbr/react-boilerplate/blob/master/app/routes.js#L17
https://github.com/mxstbr/react-boilerplate/blob/master/app/routes.js#L33

mxstbr:
I don't really know what the problem here is either, but it doesn't seem to be a problem with the boilerplate so I'll close this for now!

mxstbr:
Ooops, didn't notice it was already closed. 🙊

EducatedTard:
I have the same problem as @aslight21 ... When you create an empty project (with npm run clean), generate a container, generate a route, then assign the container to the route, the error pops up in the browser console when you are on the route. Now, I've made all the corrections you've mention above and I still have an error: es6.promise.js:119 Unhandled promise rejection TypeError: Cannot read property 'apply' of undefined(…)

[...]

PR #361, opened 2016-05-21
Update container and selectors templates 

ralomucho:
Following discussion in #337, I rewrote container and selectors templates :

- added comments to explain all of these selectors, mapStateToProps stuff
- simplify code in container template to use the default selector defined in selector.js, inside connect. All selector stuff is now in selectors.js

I also fixed an issue in reducer.test.js

### DIFF
--- a/internals/generators/container/index.js.hbs
+++ b/internals/generators/container/index.js.hbs
@@ -11,8 +11,6 @@ import select{{properCase name}} from './selectors';
 import styles from './styles.css';
 {{/if}}
 
-// Underlying component
-// named export so that it can be tested
 export class {{ properCase name }} extends React.Component { // eslint-disable-line react/prefer-stateless-function
   render() {
     return (
@@ -27,17 +25,12 @@ export class {{ properCase name }} extends React.Component { // eslint-disable-l
   }
 }
 
-// mapStateToProps selects parts of the state to be used by the underlying component
-// Selection is made by the default selector declared in selectors.js
 const mapStateToProps = select{{properCase name}}();
 
-// mapDispatchToProps() allows to pass dispatch store method and our own
-// handlers to the underlying component
 function mapDispatchToProps(dispatch) {
   return {
     dispatch,
   };
 }
 
-// The container (smart component) to be exported by default
 export default connect(mapStateToProps, mapDispatchToProps)({{ properCase name }});
--- a/internals/generators/container/index.js.hbs
+++ b/internals/generators/container/index.js.hbs
@@ -6,19 +6,14 @@
 
 import React from 'react';
 import { connect } from 'react-redux';
-
-{{#if wantActionsAndReducer}}
-import { createSelector } from 'reselect';
-import {
-  select{{properCase name}},
-} from './selectors';
-{{/if}}
-
+import select{{properCase name}} from './selectors';
 {{#if wantCSS}}
 import styles from './styles.css';
 {{/if}}
 
-class {{ properCase name }} extends React.Component {
+// Underlying component
+// named export so that it can be tested
+export class {{ properCase name }} extends React.Component { // eslint-disable-line react/prefer-stateless-function
   render() {
     return (
       {{#if wantCSS}}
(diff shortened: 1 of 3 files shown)


internals/generators/container/selectors.js.hbs
```diff
@@ -1,7 +1,44 @@
...
+/**
+ * direct selector to the {{ camelCase name }} state domain
+ */
+const select{{ properCase name }}Domain = () => state => state.get('{{ camelCase name }}');
 
+/**
+ * other specific selectors
+ */
+
+// exemples
```
mxstbr: Nit: Change to "Examples"

internals/generators/container/selectors.js.hbs
```diff
@@ -1,7 +1,44 @@
 import { createSelector } from 'reselect';
 
-const select{{ properCase name }} = () => state => state.get('{{ properCase name }}')
+/**
+ * direct selector to the {{ camelCase name }} state domain
+ */
+const select{{ properCase name }}Domain = () => state => state.get('{{ camelCase name }}');
 
+/**
+ * other specific selectors
```
mxstbr: Nit: Capitalize first letter, i.e. should be "Other specific selectors"

internals/generators/container/selectors.js.hbs
```diff
@@ -1,7 +1,44 @@
 import { createSelector } from 'reselect';
 
-const select{{ properCase name }} = () => state => state.get('{{ properCase name }}')
+/**
+ * direct selector to the {{ camelCase name }} state domain
```
mxstbr: Nit: Capitalize first letter, i.e. should be "Direct selector…"

internals/generators/container/index.js.hbs
```diff
@@ -6,47 +6,38 @@
...
-{{/if}}
-
+import select{{properCase name}} from './selectors';
 {{#if wantCSS}}
 import styles from './styles.css';
 {{/if}}
 
-class {{ properCase name }} extends React.Component {
+// Underlying component
+// named export so that it can be tested
```
mxstbr: Nit: Change to "Exported separately for testing"

mxstbr:
A few nitpicks, but otherwise this looks good.

----

From a usage perspective though, I definitely don't want those comments in every container I ever generate, since after the first time I should know what's going on.

I'm not sure including them in every single generated container is a good way to go…

ralomucho:
Removed comments.
Maybe adding something in docs?

mxstbr:
This looks amazing, and that would definitely be a case of adding to the docs!

One small question, shouldn't we only include a selector if we generate a reducer? I.e. if `wantActionsAndReducer` is true?

Since that basically signifies that we want to add a new part of our state, we should only really select that if it exists, otherwise that doesn't make much sense, right?

ralomucho:
I don't see the reason to know if there is a reducer or not. you choose the selector form you want in the default selector in `selector.js` : a selector created by `createSelector`, a direct access to a part of the state, a plain object...
`selectSomething()` is just a function which takes in state and give back something. I think it's more cohesive to systematically pass  `selectMyContainer()` to `connect` (through `mapStateToProps`) in `MyContainer` container, and define `selectMyContainer` in `selector.js`, even if it's only does something like `state => state.get('myProperty')`. A selector is just a getter (which can memoize if it was created with `createdSelector`)
If there is no reducer in our container, there is no new domain in the combined reducer. The container will take pieces of state somewhere else (in other domains). A container is a statefull component so it must get something from the state.

mxstbr:
> A container is a stateful component so it must get something from the state.

I absolutely agree with everything you said, but that "something" might also just be the `dispatch` function, in which case a selector isn't necessary, right?

I generally agree though, I don't think many people will use containers _just_ to inject `dispatch`. Let me have a think about this, also pinging @oliverturner who might have an amazing insight. ✨

mxstbr:
Okay, this looks good to me, you're right @ralomucho! Thanks so much for taking the time to submit this PR, much appreciated! :+1:

PR #363, opened 2016-05-21
Update route.hbs

Bigbrd:
create routes that point to components in the components folder


mxstbr:
Awesome, thanks Bryan! :tada:

PR #366, opened 2016-05-22
Fix the first command in the overview section

ishg:
changed 'npm setup' to 'npm run setup'


mxstbr:
Ah, good catch! Thanks very much!

PR #368, opened 2016-05-22
Fix component test generator

ralomucho:
wrong path to component to be tested


mxstbr:
Good catch, thanks!

----

_Can I reach you somewhere a bit more private? Would love to chat!_

PR #380, opened 2016-05-24
Update stylelint-config-standard to 8.0.0

glennreyes:
.. with couple of [new rules](https://github.com/stylelint/stylelint-config-standard/blob/master/CHANGELOG.md#800) :)


mxstbr:
Nice one! :+1:

PR #388, opened 2016-05-25
Updated docs for removing reselect

robbue:
Changed example code to convert state object to raw js with toJS()


mxstbr:
Awesome!! Great job, thanks so much!

PR #390, opened 2016-05-25
Prevent duplication of sagas when re-entering routes.

### LINKED ISSUE #384: github request for each time you navigate to the home page
Each time you navigate back to the home page, there is another fetch done for each github search.

1. Run sample app.
2. Open Redux dev tools.
3. Navigate to Features page.
4. Click Back or Home.
5. Search for github user.
6. See two `LOAD_REPOS_SUCCESS` actions, and 2 requests in the network tab.

I assume it is because `runSaga` is called each time you navigate, but I'm pretty new to redux-saga.

kraffslol:
Prevents the addition of duplicate sagas when returning to routes that has already been loaded. 
Fixes #384. 
I'm sure this isn't the best way to fix this bug, but it's the best I could think of at the moment. Suggestions on improving this are more than welcome.

### DIFF
--- a/app/routes.js
+++ b/app/routes.js
@@ -31,7 +31,7 @@ export default function createRoutes(store) {
 
         importModules.then(([reducer, sagas, component]) => {
           injectReducer('home', reducer.default);
-          injectSagas(sagas.default);
+          injectSagas('home', sagas.default);
 
           renderRoute(component);
         });
--- a/app/store.js
+++ b/app/store.js
@@ -34,6 +34,7 @@ export default function configureStore(initialState = {}, history) {
   // Extensions
   store.runSaga = sagaMiddleware.run;
   store.asyncReducers = {}; // Async reducer registry
+  store.asyncSagas = {}; // Async saga registry
 
   // Make reducers hot reloadable, see http://mxs.is/googmo
   /* istanbul ignore next */
--- a/app/utils/hooks.js
+++ b/app/utils/hooks.js
@@ -14,7 +14,13 @@ export function injectAsyncReducer(store) {
  * Inject an asynchronously loaded saga
  */
 export function injectAsyncSagas(store) {
-  return (sagas) => sagas.map(store.runSaga);
+  const prevSagas = [];
+  return (sagas) => sagas.forEach((saga) => {
+    if (!prevSagas.includes(saga)) {
+      store.runSaga(saga);
+      prevSagas.push(saga);
+    }
+  });
 }
 
 /**
--- a/app/utils/hooks.js
+++ b/app/utils/hooks.js
@@ -11,16 +11,17 @@ export function injectAsyncReducer(store) {
 }
 
 /**
- * Inject an asynchronously loaded saga
+ * Inject an asynchronously loaded saga if it's not already started.
+ * If the saga is previously loaded it will prevent starting duplicate instances
+ * of the same saga. See http://mxs.is/googng.
  */
 export function injectAsyncSagas(store) {
-  const prevSagas = [];
-  return (sagas) => sagas.forEach((saga) => {
-    if (!prevSagas.includes(saga)) {
-      store.runSaga(saga);
-      prevSagas.push(saga);
+  return (name, sagas) => {
+    if (!store.asyncSagas[name]) {
+      store.asyncSagas[name] = sagas; // eslint-disable-line
+      sagas.forEach(store.runSaga);
     }
-  });
+  };
 }
 
 /**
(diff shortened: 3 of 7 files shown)


mxstbr:
Awesome, love it! Honestly can't think of a better solution for this. :+1:

A small comment explaining what that piece of code does and maybe linking to the shortlink of the issue (`mxs.is/googng`) would be awesome, then we'll get this merged! Thanks @kraffslol :fire:

jwinn:
My suggestion, for consistency, would be to make it like `injectAsyncReducer` using a hash key. :-)

kraffslol:
@jwinn I made a similar solution to injectAsyncReducer which seems to work but realised that this will break all existing react-boilerplate projects since it will now require a name for injectSagas (like injectReducer currently does). Do you still believe it's worth it, or should I go with the original solution?

jwinn:
@kraffslol I'm going to leave the choice up to @mxstbr, but I don't think it's such a problem to add another key/prop to `store.js` `store.asyncSagas` or something similar. Again, I leave it up to @mxstbr. Thanks for doing this though btw. 👍

mxstbr:
I think it'd be good to have both the same way for consistency, but I don't particularly mind which way. I think it's easier without the name?

kraffslol:
I just committed changes that makes me more like the current injectAsyncReducer function.

This will most likely break existing installations of react-boilerplate if they're pulling from git.

To fix this look in routes.js and change lines like these:
```javascript
injectSagas(sagas.default);
```
To:
```javascript
injectSagas('COMPONENTNAME', sagas.default);
```

mxstbr:
Now that I think about it, it's a bit of unnecessary overhead to pass the name of the component into both `injectAsyncReducer` and `injectAsyncSaga`, right?

Since there's not difference between having the name and using @kraffslol's approach, I'd rather go with that and not have to remember to do that—or am I missing something?

(Sorry for the bikeshedding, this just popped into my mind now…)

jwinn:
There is a difference on performance for read/writes, plus equality of function checks in JS isn't, how shall we say, bulletproof. However, I'm game for the simpler solution and optimize if issues arise later.

mxstbr:
> equality of function checks in JS isn't, how shall we say, bulletproof

Could you elaborate a little bit? I've never ran into that.

jwinn:
Well equality checks in general can be quite complex, depending upon the type of equality one is checking for (e.g. reference, value, syntactic, (ex/in)tensional, etc.) but that's for another discussion altogether. In this case, we're using generator objects so the `Array.prototype.includes` equality check would be by reference and should work as expected. I just wanted to note (for future posterity?) that relying too heavily on builtin equality comparisons isn't bulletproof nor a panacea solution.

TLDR; I was just trying to say there is a difference between the two approaches on a couple of levels, but for our uses I see those differences as not really an issue.

mxstbr:
Thanks for the explanation, that's very interesting!

Lets go with the non-naming approach for now then, since it significantly makes the using developers lives easier.

justingreenberg:
i would just point out that the injector isn't really meant to manage saga lifecycle, but rather as a means to hook into saga middleware. this solution does a good job managing "daemon" style sagas, but what if the saga is supposed to be hooked into route lifecycle, ie loading data? or what if a saga is canceled but still exists in the array, preventing it from re-running? in other words, this solution is tailored to the example saga but i think it ignores a lot of real world scenarios

ie to prevent duplication of example saga, we just have to remove the `while(true)` wrapper

mxstbr:
I have to agree with @justingreenberg, what he pointed out is something I hadn't even considered. Sagas themselves should manage their lifecycle, not our custom `injectAsyncSaga` method.

I'm inclined to merge #399 to fix this issue, it seems like the more reasonable way of doing this. Any oppositions?

kraffslol:
Sounds good to me. While I feel that it could be a bit of a hassle using race every time you want to daemonize a saga, I do believe it's a much better solution than mine.

mxstbr:
Thanks so much for kicking this whole thing off though @kraffslol, glad we found a good solution for this! Sorry that we won't merge this PR 😢

rizedr:
@mxstbr I think it would be a good idea to reflect this practice in the [docs](https://github.com/mxstbr/react-boilerplate/blob/master/docs/js/redux-saga.md) as well.

justingreenberg:
this should be resolved by #1746 — feedback is welcome!

Dattaya:
@cidevant, the bug was fixed in 3.5 and I recommend to migrate to it as soon as the new version released.

cidevant:
@Dattaya 3.5 isn't released yet? Where can I find it?

vicbell:
@cidevant  If you use `dev`, the code is all in there.

PR #392, opened 2016-05-25
Fix wrong command on CLI readme for pagespeed

brousalis:
Readme should be `npm run start` instead of `npm run serve`


mxstbr:
It should actually be `npm run start:prod`, since you want to check the performance of the production version! Would you mind updating the PR? Then we can merge it! :+1:

brousalis:
Updated!

mxstbr:
Awesome, thanks so much!

PR #399, opened 2016-05-26
Fix example saga

### LINKED ISSUE #384: github request for each time you navigate to the home page

justingreenberg:
Alternative to #390, resolves #384

### DIFF
--- a/app/containers/HomePage/sagas.js
+++ b/app/containers/HomePage/sagas.js
@@ -19,20 +19,18 @@ export default [
 
 // Individual exports for testing
 export function* getGithubData() {
-  while (true) {
-    yield take(LOAD_REPOS);
-    const username = yield select(selectUsername());
-    const requestURL = `https://api.github.com/users/${username}/repos?type=all&sort=updated`;
+  yield take(LOAD_REPOS);
+  const username = yield select(selectUsername());
+  const requestURL = `https://api.github.com/users/${username}/repos?type=all&sort=updated`;
 
-    // Use call from redux-saga for easier testing
-    const repos = yield call(request, requestURL);
+  // Use call from redux-saga for easier testing
+  const repos = yield call(request, requestURL);
 
-    // We return an object in a specific format, see utils/request.js for more information
-    if (repos.err === undefined || repos.err === null) {
-      yield put(reposLoaded(repos.data, username));
-    } else {
-      console.log(repos.err.response); // eslint-disable-line no-console
-      yield put(repoLoadingError(repos.err));
-    }
+  // We return an object in a specific format, see utils/request.js for more information
+  if (repos.err === undefined || repos.err === null) {
+    yield put(reposLoaded(repos.data, username));
+  } else {
+    console.log(repos.err.response); // eslint-disable-line no-console
+    yield put(repoLoadingError(repos.err));
   }
 }
--- a/app/containers/HomePage/sagas.js
+++ b/app/containers/HomePage/sagas.js
@@ -4,7 +4,9 @@
 
 /* eslint-disable no-constant-condition */
 
-import { take, call, put, select } from 'redux-saga/effects';
+import { take, call, put, select, race } from 'redux-saga/effects';
+
+import { LOCATION_CHANGE } from 'react-router-redux';
 
 import { LOAD_REPOS } from 'containers/App/constants';
 import { reposLoaded, repoLoadingError } from 'containers/App/actions';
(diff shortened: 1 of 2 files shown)


mxstbr:
I can't try at the moment, but wouldn't this change mean you couldn't reload new repositories once a user's were loaded?

justingreenberg:
@mxstbr sorry stepped away for a few mins! yeah `getGithubData` would only receive first request after loading... i updated PR to leave as watcher using race effect to [cancel on location change](https://github.com/mxstbr/react-boilerplate/pull/399/files#diff-3958a24e8f3cadcf34c019f8bb7e75b8R27)

the idea really is just that sagas are flexible enough to manage themselves so `injectSaga` helper doesn't really need to be responsible for managing their lifecycle

mxstbr:
Awesome, thanks for chiming in @justingreenberg, this LGTM!

holoiii:
Just running into the duplicate saga execution issue myself today and I'm looking at the various issues posted around them.

Is the idea here that upon changing routes, the saga will 'unbind' itself from the store? That way when the route is hit again, re-injecting the same saga will not cause any duplication?

Also, does this mean that we can no longer use saga helpers like `takeEvery` or `takeLatest` in route-specific sagas? Since we'll always need to use the `race` function now.

mxstbr:
> Is the idea here that upon changing routes, the saga will 'unbind' itself from the store? That way when the route is hit again, re-injecting the same saga will not cause any duplication, since the saga is completed already?

See @justingreenberg's very true comment in #390: 

> I would just point out that the injector isn't really meant to manage saga lifecycle... this solution will manage "daemon" style sagas, but what if the saga is intentionally hooked into route lifecycle, ie data loading? maintaining state will also be an issue since helpers are curried in outer scope of router. in other words, this solution is tailored to the example saga but i think it ignores a lot of real world scenarios

----

> Also, does this mean that we can no longer use saga helpers like takeEvery or takeLatest in route-specific sagas? Since we'll always need to use the race function now.

You only need to use race if your saga needs to run again and again and again. For us, not using a while loop was a problem since we need to refetch data when a new username is entered, otherwise it wouldn't have been a problem!

holoiii:
Wouldn't you still need to use `race` even for one-off sagas? Otherwise, if you visit the same route again, the same saga gets reinjected but the old one is hanging around.

For example, wouldn't you need something like this:
```javascript
function* saga() {
  const watcher = yield race({
    action: take(ACTION),
    stop: take(LOCATION_CHANGE)
  });
  if (watcher.stop) return;
  ...
}
```

Sorry if the questions are basic, I'm learning a lot of the concepts used in react-bootstrap for the first time (Great framework btw! Really enjoying reading through the code and seeing how everything is tied together nicely).

justingreenberg:
sagas are just meant for managing control flow.. for example, if we wanted to load the github repos on initial load, we could break into multiple sagas: 

reducer.js
```js
//...
const initialState = fromJS({
  username: 'mxstbr', // set a default
});
```

sagas.js
```js
// ...
// Bootstrap sagas
export default [
  main,
];

function* main() {
  // create variable for watcher
  const watcherInstance = yield fork(getGithubDataWatcher);

  // load data on initial route load
  // ** note: must set username on reducer initialState ! **
  yield fork(getGithubData);

  // if location changes, continuation
  yield take(LOCATION_CHANGE);
  yield cancel(watcherInstance);// cancel task instance
}

function* getGithubDataWatcher() {
  try {
    while (yield take(LOAD_REPOS)) {
      yield call(getGithubData);
    }
  } finally {
    if (yield cancelled()) console.log('watcher cancelled')
  }
}

function* getGithubData() {
  const username = yield select(selectUsername());
  const requestURL = `https://api.github.com/users/${username}/repos?type=all&sort=updated`;

  // Use call from redux-saga for easier testing
  const repos = yield call(request, requestURL);

  // We return an object in a specific format, see utils/request.js for more information
  if (repos.err === undefined || repos.err === null) {
    yield put(reposLoaded(repos.data, username));
  } else {
    console.log(repos.err.response); // eslint-disable-line no-console
    yield put(repoLoadingError(repos.err));
  }
}
```

i don't really have a better answer other than to say this is a super flexible setup :)

matrixl:
can the main function manage multiple watchers? or i need to create three functions for every async call?

PR #401, opened 2016-05-27
fix stacking of async sagas execution on url change

EugeneHlushko:
I had a problem when Saga was lunched several times if you navigate through the app and return to container with saga who would send request to server.

So basically my request ran as many times as i entered this specific url and its saga was passed to injectAsyncSagas.

### DIFF
--- a/app/utils/hooks.js
+++ b/app/utils/hooks.js
@@ -1,5 +1,8 @@
 import createReducer from '../reducers';
 
+// const runnedSagas = [];
+const runnedSagas = [];
+
 /**
  * Inject an asynchronously loaded reducer
  */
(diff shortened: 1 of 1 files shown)


mxstbr:
This is the same solution as @kraffslol suggested in #390, which we didn't merge because @justingreenberg's #399 made more sense. This is fixed if you pull right now!

Thanks so much for taking the time to submit this PR though!

holoiii:
Just running into the duplicate saga execution issue myself today and I'm looking at the various issues posted around them.

Is the idea here that upon changing routes, the saga will 'unbind' itself from the store? That way when the route is hit again, re-injecting the same saga will not cause any duplication?

PR #404, opened 2016-05-27
Fix selectors generator to not show an error message by default

holoiii:
Related to this issue: https://github.com/mxstbr/react-boilerplate/issues/337


mxstbr:
Thanks @holoiii, awesome work! :tada:

Issue #379, opened 2016-05-23
Hot reloading of async reducers


tsm91:
Hey @mxstbr,

does this works for you?

`react-boilerplate/app/containers/HomePage/reducer.js`  and change

```
const initialState = fromJS({
  username: '', // to username: 'whatever'
});
```

in the console i see

```
[HMR] bundle rebuilding
[HMR] bundle rebuilt in 731ms
[HMR] Checking for updates on the server...
[HMR] Updated modules:
[HMR]  - ./app/containers/HomePage/reducer.js
[HMR]  - ./app/routes.js
[HMR]  - ./app/app.js
[HMR]  - multi main
[HMR] App is up to date.
```

but nothing happens, the state won't change

mxstbr:
It's possible that it doesn't work… I can't check right now, would love somebody investigating further!

saschatimme:
I think there is an error in the store config that causes the error.

The current config is:
```js
  if (module.hot) {
    module.hot.accept('./reducers', () => {
      const nextRootReducer = require('./reducers').default; // eslint-disable-line global-require
      store.replaceReducer(nextRootReducer);
    });
  }
```

But I think ist should be

```js
if (module.hot) {
        module.hot.accept("./reducers", () => {
            const nextRootReducer = require("./reducers").default; // eslint-disable-line global-require
            store.replaceReducer(nextRootReducer(store.asyncReducers));
        });
    }
```

since the signature of ```require("./reducers").default``` is ```asyncReducers => reducer```.

mxstbr:
Ah yeah, that sounds reasonable! Have you tried implementing this, and did it fix the problem?

tsm91:
@mxstbr @saschatimme that doesn't solve the issue

AJIh:
@tsm91 
async reducers should be re-injected to store after hot reloading. I rewrote `injectReducer` and route actions to achieve this.

```javascript
export function injectAsyncReducer(store, isValid) {
  return function injectReducer(name, asyncReducer, force = false) {
    if (!isValid) checkStore(store);

    invariant(
      isString(name) && !isEmpty(name) && isFunction(asyncReducer),
      '(app/utils...) injectAsyncReducer: Expected `asyncReducer` to be a reducer function'
    );

    if (!force && Reflect.has(store.asyncReducers, name)) return;

    store.asyncReducers[name] = asyncReducer; // eslint-disable-line no-param-reassign
[...]

Issue #331, opened 2016-05-19
Get rid of static reducers


ralomucho:
In container generator, some plop actions update reducers.js when you add a reducer to the container.
In the same time, if you assign a route to this container, the reducer will be dynamically imported so the former reference in reducers.js  is no more relevant.

I think there must be situations when containers  are used without being linked to a route, and so "static" (don't know if it's the appriopriate word) reducer are still usefull. So we might make the distinction between the two kinds of reducer in container generator (by asking user if the container must be linked to a page). I can do this if you agree.

What do you think?

mxstbr:
Ah, that sounds like a good call actually! Would you mind submitting a PR with that so I can take a closer look?

ralomucho:
ok, I'll do it.

ralomucho:
I don't have time to do it for the moment. I'll do it next week.

mxstbr:
I think there's a case to be made that we do need static reducers, (e.g. for routing, i18n,…) but we should not be adding the newly generated, code-split reducers to the static reducers list!

That's a pretty bad bug we should fix asap, I'd appreciate a PR changing the container generator to no longer do that!

mxstbr:
Done by the great @ralomucho in #509! Thanks!! :tada:

PR #406, opened 2016-05-27
Fix HMR for sync reducers

justingreenberg:
regarding #379 - 

- fixes issue pointed out by @saschatimme
- switches from legacy require.ensure to system loader

however i was not able to get this working with async reducers (yet)... i have several projects working with vanilla webpack HMR (and async reducers), but they not using react-transform or webpack v2

basically, with async imports the dependency graph requires instrumenting module.hot at the root to accept routes.js to rebuild... as @tsm91 pointed out in his console log:

```
[HMR] Updated modules:
[HMR]  - ./app/containers/HomePage/reducer.js
[HMR]  - ./app/routes.js
[HMR]  - ./app/app.js    <--- needs to capture update and re-render routes
```

can be resolved using vanilla hmr, but local component state will be lost on reloads. this also might be solved with react-hot-loader v3, but requires more investigation.

### DIFF
--- a/app/store.js
+++ b/app/store.js
@@ -39,8 +39,12 @@ export default function configureStore(initialState = {}, history) {
   /* istanbul ignore next */
   if (module.hot) {
     module.hot.accept('./reducers', () => {
-      const nextRootReducer = require('./reducers').default; // eslint-disable-line global-require
-      store.replaceReducer(nextRootReducer);
+      System.import('./reducers').then((reducerModule) => {
+        const createReducers = reducerModule.default;
+        const nextReducers = createReducers(store.asyncReducers);
+
+        store.replaceReducer(nextReducers);
+      });
     });
   }
 
--- a/internals/templates/store.js
+++ b/internals/templates/store.js
@@ -37,9 +37,11 @@ export default function configureStore(initialState = {}, history) {
   // Make reducers hot reloadable, see http://mxs.is/googmo
   /* istanbul ignore next */
   if (module.hot) {
-    module.hot.accept('./reducers', () => {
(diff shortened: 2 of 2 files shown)


mxstbr:
Hmm, since we're getting rid of static reducers hopefully (ref #331), will this still be needed?

justingreenberg:
@mxstbr totally up to you! actually i didn't realize you guys were planning on removing sync reducers... i think there's a case to be made for leaving them ie for core ui, locale and auth state

mxstbr:
> i think there's a case to be made for leaving them ie for core ui, locale and auth state

Yeah, makes sense, we'll have to discuss that a little bit more!

Lets merge this PR, could you copy and paste the changes to `internals/templates/store.js`? Thanks! :+1:

ralomucho:
I didn't want to  get rid off static reducers but to make the difference inside  generators between static and async ones !

mxstbr:
Yeah, I misunderstood that, totally agree! Lets get this updated and merged @justingreenberg :shipit:

justingreenberg:
@mxstbr updated 👍

mxstbr:
Awesome, thanks Justin!

petemill:
I'm pretty sure this doesn't work. Not only is the code in the template file from store.js different from that in the sample file for store.js, once implementing the `module.hot.accept('./reducers')` syntax, none of the static or async reducers are reoloadable.

justingreenberg:
nice catch -- fixed in #1154

i'm not sure why HMR is not working for reducers, since this has been a solid pattern forever. it could be related to recent webpack updates.. i'm way too busy with work at the moment to investigate but please let me know if you figure it out :)

PR #410, opened 2016-05-28
when HomePage initial state username is not null, submit the form to …

suhaotian:
…load repos

### DIFF
--- a/app/containers/HomePage/index.js
+++ b/app/containers/HomePage/index.js
@@ -38,7 +38,7 @@ export class HomePage extends React.Component {
    * when initial state username is not null, submit the form to load repos
    */
   componentDidMount() {
-    if (this.props.username.trim().length > 0) {
+    if (this.props.username && this.props.username.trim().length > 0) {
       this.props.onSubmitForm();
     }
   }
--- a/app/containers/HomePage/index.js
+++ b/app/containers/HomePage/index.js
@@ -52,6 +52,15 @@ export class HomePage extends React.Component {
     this.openRoute('/features');
   };
 
+  /**
+   * when initial state username is not null, submit the form to load repos
+   */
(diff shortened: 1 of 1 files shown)


mxstbr:
When does this happen?

suhaotian:
@mxstbr 
**app/containers/HomePage/reducer.js**
```js
// The initial state of the App
const initialState = fromJS({
  username: 'mxstbr',
});
```

when username have a default value and not null, reload page doesn't load repos.

suhaotian:
Maybe we can rewrite it like **redux-saga/examples/async** code

app/containers/HomePage/index.js
```diff
@@ -123,7 +132,8 @@ function mapDispatchToProps(dispatch) {
     onChangeUsername: (evt) => dispatch(changeUsername(evt.target.value)),
     changeRoute: (url) => dispatch(push(url)),
     onSubmitForm: (evt) => {
-      evt.preventDefault();
+      /* eslint no-unused-expressions: ["error", { "allowShortCircuit": true }] */
```
mxstbr: Change this to `// eslint-disable-line no-unused-expressions`!

mxstbr:
Ah, good catch, I like this solution! :+1: One small nitpick, then we can merge this fix—thanks so much!

suhaotian:
change it, but **npm run lint** not pass:
```bash
  136:7  error  Expected an assignment or function call and instead saw an expression  no-unused-expressions

✖ 1 problem (1 error, 0 warnings)
```

mxstbr:
Yeah, I think it needs to be on the same line, but this looks fine. Thanks @suhaotian!

suhaotian:
@mxstbr My pleasure!

PR #411, opened 2016-05-28
Update to more performant source mapping

justingreenberg:
rationale: since inline-source-map generates 1:1 line-column mappings it is much more expensive to generate... dev mode can use cheap-module-eval-source-map algorithm as it provides huge performance gains by only mapping the line numbers from original source

@mxstbr this change has been removed from intl PR :)


mxstbr:
Awesome, thanks so much @justingreenberg!

PR #413, opened 2016-05-29
npm script to check the npm version before install

rogr:
If the npm version is lower to 3, then `npm install` doesn't start and show an error that says:
`[ERROR: React Boilerplate] You need npm version @>=3`

Otherwise if the version is major or equal to 3 `npm install` executes


mxstbr:
Oooh I like this, nice one @rogr!

rogr:
Damn Travis is failing, I will check why.

PR #414, opened 2016-05-29
Don't let Webpack spam the terminal on errors

clessg:
When there's an error in development such as an invalid import, Webpack likes to spam the terminal with a bunch of (as far as I can tell) useless stuff:

This pull request causes the useless spam to be excluded.

### DIFF
--- a/server/middlewares/frontendMiddleware.js
+++ b/server/middlewares/frontendMiddleware.js
@@ -12,6 +12,7 @@ const addDevMiddlewares = (app, options) => {
     noInfo: true,
     publicPath: options.output.publicPath,
     silent: true,
+    stats: 'errors-only',
   });
 
   app.use(middleware);


mxstbr:
This is a lot nicer, thanks @clessg!

hamishtaplin:
I'm still seeing this happen when a test fails due to a dodgy `import`.

clessg:
It probably involves changing [this part](https://github.com/mxstbr/react-boilerplate/blob/master/internals/testing/karma.conf.js#L33) and adding the following line:

`stats: 'errors-only',`

I don't currently have the time to test this. (No pun intended?)

hamishtaplin:
@clessg That did it, cheers.

clessg:
Great. I'll make a PR later.

Issue #412, opened 2016-05-28
Fail when re-install dependencies


tylerle:
Hi guys,

I have a problem when reinstalling dependencies as follows:

npm run setup -- worked
npm run start -- worked
rm -rf node_modules -- worked
npm install -- not work

When I run npm install, I get the following error:

npm ERR! node v4.4.4
npm ERR! npm  v2.15.1
npm ERR! code EPEERINVALID

npm ERR! peerinvalid The package webpack@2.1.0-beta.7 does not satisfy its siblings' peerDependencies requirements!
npm ERR! peerinvalid Peer babel-loader@6.2.4 wants webpack@1 || ^2.1.0-beta
npm ERR! peerinvalid Peer extract-text-webpack-plugin@1.0.1 wants webpack@^1.9.11
npm ERR! peerinvalid Peer karma-webpack@1.7.0 wants webpack@^1.4.0
npm ERR! peerinvalid Peer offline-plugin@3.2.1 wants webpack@>=1.0.0
npm ERR! peerinvalid Peer webpack-dev-middleware@1.6.1 wants webpack@1 || ^2.1.0-beta

rogr:
You need to upgrade to `npm@>=3`.

mxstbr:
What @rogr said!

Issue #385, opened 2016-05-25
Cannot setup package


Kinnza:
npm install fails because of webpack version

```
npm ERR! Darwin 15.4.0
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "install"
npm ERR! node v4.4.0
npm ERR! npm  v2.14.20
npm ERR! code EPEERINVALID

npm ERR! peerinvalid The package webpack@2.1.0-beta.7 does not satisfy its siblings' peerDependencies requirements!
npm ERR! peerinvalid Peer babel-loader@6.2.4 wants webpack@1 || ^2.1.0-beta
npm ERR! peerinvalid Peer extract-text-webpack-plugin@1.0.1 wants webpack@^1.9.11
npm ERR! peerinvalid Peer karma-webpack@1.7.0 wants webpack@^1.4.0
npm ERR! peerinvalid Peer offline-plugin@3.2.1 wants webpack@>=1.0.0
npm ERR! peerinvalid Peer webpack-dev-middleware@1.6.1 wants webpack@1 || ^2.1.0-beta
```

rogr:
you need at least [node > 5](https://github.com/mxstbr/react-boilerplate/blob/master/.travis.yml#L5)

mxstbr:
Yeah, this is fixed by an upgrade to `npm@>=3`.

Sadly, this is out of our hands and in the hands of package authors, there's nothing we can do here! Sorry!

Kinnza:
Ok thanks. So upgrading npm to 3 should fix it? i'll try :)

mxstbr:
Yeah, that's hopefully going to fix that! Let us know how it goes!

Kinnza:
Ok now the setup passed. Do i have to have node 5?

mxstbr:
Well, you have to have `npm@>=3`. If you can figure out how to use that with `node@<5`, that's absolutely fine!

Kinnza:
👍

PR #415, opened 2016-05-29
Script to check the npm version before install

rogr:
If the npm version is lower to 3, then `npm install` doesn't start and show an error that says:
[`[ERROR: React Boilerplate] You need npm version @>=3`
](https://travis-ci.org/rogr/react-boilerplate/jobs/133679428#L276)
Otherwise if the version is major or equal to 3 `npm install` runs.

This helps with issues like #412, #385, ...etc.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ sudo: true
 dist: trusty
 node_js:
   - "5.0"
-  - "0.12"
 script: npm run build
 before_install:
   - export CHROME_BIN=/usr/bin/google-chrome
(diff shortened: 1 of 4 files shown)


rogr:
Wait a moment I think I'm just complicated my self.
I only need to exec `npm -v`

mxstbr:
Hmm, not 100% convinced this is a good idea.

We won't need `npm@>=3` forever, just until the react v15 and the webpack v2 update mania have died down and peer dependencies are normal again…

rogr:
I can delete it when webpack v2 is updated.
What is your concern the extra file on `internals` ?

mxstbr:
You're right! :grin: I like the solution you found, thanks @rogr!

Issue #348, opened 2016-05-20
Migrate from ESLint + JSCS to just ESLint


sscotth:
http://eslint.org/blog/2016/04/welcoming-jscs-to-eslint

MisterFantastic:
Please let us know why you are removing JSCS as it is good linter.

mxstbr:
The JSCS team released 3.0 and announced they were merging the teams with ESLint, see the blogpost posted in the OP.

Agree @sscotth, we should do this sometime. It's imo not super high priority since it works, but I'd merge a PR making this happen!

sscotth:
Agreed on not being high priority, but with JSCS not being maintained any more a path forward is necessary.

> We will continue to support JSCS for three months fixing bugs but will not be implementing any new features.

https://medium.com/@markelog/jscs-end-of-the-line-bc9bf0b3fdb2

mxstbr:
Done by @glennreyes in #417! :fire:

PR #417, opened 2016-05-29
Clean up JS linting

glennreyes:
Will solve #348 as well

### DIFF
--- a/server/logger.js
+++ b/server/logger.js
@@ -24,15 +24,13 @@ const logger = {
       console.log(`Tunnel initialised ${chalk.green('✓')}`);
     }
 
-    console.log(
-      chalk.bold(`
-Access URLs:${divider}
+    console.log(`
+${chalk.bold('Access URLs:')}${divider}
 Localhost: ${chalk.magenta(`http://localhost:${port}`)}
-      LAN: ${chalk.magenta(`http://${ip.address()}:${port}`)}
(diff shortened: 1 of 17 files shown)


mxstbr:
This looks really good, thanks Glen! :fire:

server/logger.js
```diff
@@ -12,26 +12,26 @@ const logger = {
...
 
     console.log(
-      chalk.bold('\nAccess URLs:') +
-      divider +
-      '\nLocalhost: ' + chalk.magenta('http://localhost:' + port) +
-      '\n      LAN: ' + chalk.magenta('http://' + ip.address() + ':' + port) +
-      (tunnelStarted ? '\n    Proxy: ' + chalk.magenta(tunnelStarted) : '') +
-      divider,
-      chalk.blue('\nPress ' + chalk.italic('CTRL-C') + ' to stop\n')
+      chalk.bold(`
```
Deveritas: Won't this bold the whole string now instead of just `Access URLs:`? Or am I missing something?

mxstbr: Yeah, good catch! I haven't tried it, but that sounds like a likely error–does that happen @glennreyes?

glennreyes: Yeah that was dumb, I just commited a fix of that! Thanks @Deveritas 🙌

PR #424, opened 2016-05-30
Js linter config updates

gangster:
Introduces `.eslintrc` and `.jscsrc` files for explicitly configuring javascript linters.   Also sets up some helpful starter rules, which invite refinement.   

Adds a few new npm commands for autofixing javascript linting errors.


rogr:
the linter and the jscs config are at the [package.json](https://github.com/mxstbr/react-boilerplate/blob/master/package.json#L65), I like the `fix` option.

gangster:
Oh duh.   Ok, I'll drop the config files and keep the fix commands.

PR #425, opened 2016-05-30
Add lint:eslint:fix, lint:jscs:fix and lint:js:fix npm commands

gangster:


SilentCicero:
@gangster I like this a lot. Having a specific suffix/option for it makes sense.

mxstbr:
Yeah, having that seems useful since the `--` syntax to pass arguments to the commands isn't too widely known!

I just merged #417 which removes JSCS and moved everything over to ESLint (ref #348), would you mind updating this PR and removing the `npm run lint:jscs:fix` command? Thanks!

mxstbr:
I'll close this PR for now since there hasn't been any activity in 7 days.

I'd really appreciate an update @gangster, this seems like a very useful addition indeed! Let me know when you find the time to make this happen :+1:

PR #431, opened 2016-05-31
re install pre-commit after setup

### LINKED ISSUE #371: pre-commit needs to be re-installed after the clean command
Once you run the command that cleans the git repository and creates a new one, you have to install again `pre-commit` to make it work.

Maybe this should be done by the clean task itself?

FezVrasta:
fixes #371


mxstbr:
LGTM, thanks!

PR #437, opened 2016-06-01
Full paths

raksonibs:
Still needs work.


mxstbr:
Just to tell you why we don't need that, it's the webpack modulesDirectories option: https://webpack.github.io/docs/configuration.html#resolve-modulesdirectories It automatically checks if the modules exist in the `app/` folder! :+1:

raksonibs:
Ha, yeah I noticed that after. I ran into some issues with heroku though complaining, but then made sure the modulesDirectories was in dependencies. I am writing that heroku doc soon. :)

PR #470, opened 2016-06-05
Implement Webpack DLL Plugin

datapimp:
**This is a WIP. Not ready to be merged.**

Here is a proof of concept for the Webpack DLL Plugin mentioned in #319

In a development environment, it replaces the webpack commons chunk plugin with the Webpack DLL Reference Plugin, and adds scripts that need to be run in order to build the DLL that this plugin will reference. 

The primary benefit of this approach is an improvement in server start times, and improvements in hot module reload times.

- [x] Proof of Concept
- [x] Script to build DLL
- [ ] Configuration UI
- [x] Determine best approach for production config
- [ ] Documentation

PR #472, opened 2016-06-05
Fix code splitting

tsm91:
What it does?

Fixes code splitting of shared vendor dependencies.

file sizes before in dev environment:

```
main.js - 4.0MB
common.js.js - 23.6KB
0.chunk.js -  619KB
1.chunk.js - 518KB
2.chunk.js - 80.5KB
4.chunk.js - 5.3KB
```

file sizes after in dev environment:
```
vendor.js - 3.7MB
app.js - 375KB (main.js)
1.chunk.js -  614KB
2.chunk.js - 518KB
4.chunk.js - 19KB
5.chunk.js - 5.3KB
```

Why this still sucks? If you inspect the chunks and search for unique patterns, for example `props.handleRoute` from `components/Buttons/index.js` you will see the CommonChunks plugin didn't extract this. (This is also an issue with the current solution.)

What i think we should do to make code splitting work is to create another file container (f.e.: common.js) and put all shared codebase related stuff there like this:

```
// Shared Components
import 'components/A';
import 'components/Button';
import 'components/Footer';
import 'components/H1';
import 'components/H2';
import 'components/H3';
import 'components/Img';
import 'components/LoadingIndicator';
```

Pros:
 * this way the chunk size will be very small

Cons:
 * the browser needs to download this file whenever a shared component gets modified (This is also an issue now with chunks, because if the `Button` component gets modified all chunks will change. So it wouldn't say this is a very big problem)

cc @mxstbr

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -5,9 +5,6 @@
  * code.
  */
 
-// Needed for redux-saga es6 generator support
-import 'babel-polyfill';
-
 // Load the favicon, the manifest.json file and the .htaccess file
 import 'file?name=[name].[ext]!./favicon.ico';
 import 'file?name=[name].[ext]!./manifest.json';
--- a/app/vendor.js
+++ b/app/vendor.js
@@ -1,0 +1,22 @@
+// For vendors (f.e Lodash) just import them here unless
+// you plan on chunking vendor files for async loading. You would need to import the
+// async loaded vendors at the entry point of the async loaded file.
+
+// Polyfills
+import 'babel-polyfill';
+import 'whatwg-fetch';
+
+// Vendor
+import 'history';
+import 'immutable';
+import 'react';
+import 'react-dom';
+import 'react-pure-render';
+import 'react-redux';
+import 'react-router';
+import 'react-router-redux';
+import 'react-router-scroll';
+import 'redux';
+import 'redux-immutable';
+import 'redux-saga';
+import 'reselect';
--- a/internals/webpack/webpack.base.babel.js
+++ b/internals/webpack/webpack.base.babel.js
@@ -46,12 +46,6 @@ module.exports = (options) => ({
     }],
   },
   plugins: options.plugins.concat([
-    new webpack.optimize.CommonsChunkPlugin('common.js'),
-    new webpack.ProvidePlugin({
-      // make fetch available
-      fetch: 'exports?self.fetch!whatwg-fetch',
-    }),
-
     // Always expose NODE_ENV to webpack, in order to use `process.env.NODE_ENV`
     // inside your code for any environment checks; UglifyJS will automatically
     // drop any unreachable code.
--- a/internals/webpack/webpack.dev.babel.js
+++ b/internals/webpack/webpack.dev.babel.js
@@ -18,8 +18,8 @@ module.exports = require('./webpack.base.babel')({
     app: [
       'eventsource-polyfill', // Necessary for hot reloading with IE
       'webpack-hot-middleware/client',
-      path.join(process.cwd(), 'app/app.js') // Start with js/app.js
-    ]
+      path.join(process.cwd(), 'app/app.js'), // Start with js/app.js
+    ],
   },
 
   // Don't use hashes in dev mode for better performance
@@ -44,7 +44,12 @@ module.exports = require('./webpack.base.babel')({
 
   // Add hot reloading
   plugins: [
-    new webpack.optimize.CommonsChunkPlugin({ name: 'vendor', filename: 'vendor.js' }),
+    new webpack.optimize.CommonsChunkPlugin({
+      name: 'app',
+      children: true,
+      async: true,
+      minChunks: 3,
+    }),
     new webpack.HotModuleReplacementPlugin(), // Tell webpack we want hot reloading
     new webpack.NoErrorsPlugin(),
     new HtmlWebpackPlugin({
--- a/internals/webpack/webpack.dev.babel.js
+++ b/internals/webpack/webpack.dev.babel.js
@@ -13,11 +13,14 @@ const postcssReporter = require('postcss-reporter');
 
 module.exports = require('./webpack.base.babel')({
   // Add hot reloading in development
-  entry: [
-    'eventsource-polyfill', // Necessary for hot reloading with IE
-    'webpack-hot-middleware/client',
-    path.join(process.cwd(), 'app/app.js'), // Start with js/app.js
-  ],
+  entry: {
+    vendor: path.join(process.cwd(), 'app/vendor.js'),
+    app: [
+      'eventsource-polyfill', // Necessary for hot reloading with IE
+      'webpack-hot-middleware/client',
+      path.join(process.cwd(), 'app/app.js') // Start with js/app.js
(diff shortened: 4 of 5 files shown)


mxstbr:
This looks really promising @tsm91! 

> What i think we should do to make code splitting work is to create another file container (f.e.: common.js) and put all shared codebase related stuff there like this:

What about creating `components/index.js` which imports all the components. The component generator could then be modified to add to that long list of components?

tsm91:
@mxstbr 

> What about creating components/index.js which imports all the components. The component generator could then be modified to add to that long list of components?

Yeah the place could be `app/components/index.js`. 

What still bugs me is why is this happening? CommonChunks plugin should be responsible for just that we are doing now manually. I wonder if this has anything to do with on demand code loading (the router) or the beta version of webpack.

I am not a webpack expert, but maybe we should ask somebody who is.

mxstbr:
@jantimon @sokra do you by chance have any input to the above? We must be running against some webpack CommonsChunkPlugin limitations.

tsm91:
i had a feeling that CommonsChunkPlugin plugin wont trace on demand file load dependency graph
and we need to tell webpack about our pages, on this way it might work.

So i tried to include all page as a separate entry point: 

```js
  entry: {
    vendor: path.join(process.cwd(), 'app/vendor.js'),
    app: path.join(process.cwd(), 'app/app.js'),
    homePage: path.join(process.cwd(), 'app/containers/HomePage/index.js'),
    featurePage: path.join(process.cwd(), 'app/containers/FeaturePage/index.js'),
    notFoundPage: path.join(process.cwd(), 'app/containers/NotFoundPage/index.js'),
  },
```

Didn't work, that unique pattern in the Button component is still redundant in the chunks.

So either v2 of webpack, or something with CommonsChunkPlugin. 

Maybe this module resolution thing also matters `resolve: { modules: ['app', 'node_modules'] }`, no idea.

justingreenberg:
note if we are going to be creating DLLs, common chunks plugin will not be necessary for vendor deps in dev... from what i'm seeing in #470, it should address this issue

datapimp:
As far as I can tell in development, we won't need to use CommonsChunk plugin at all.  I am in the process of writing up a more detailed explanation of what is going on with the DLL in #470 , but I definitely see no reason to prefer CommonsChunk over DLL. 

In Production though, I still use it.

I think to really get the most value out of The CommonsChunk plugin you will want to have a configuration with multiple entry points.  For example when you split out each of your pages into their own entries, instead of just having a single entry for `main` or `app` like we do.

In these cases specifying a `minChunks` value of 3 for example, and explicitly listing the dependencies will give you the desired effect  

```javascript
{ 
  entry: {
    adminApp: 'containers/AdminApp',
    bozosApp: 'containers/BozosApp',    
    vendor: [ 
      'react', 
      'react-dom', 
      'react-router', 
      'redux', 
      'react-redux',
    ],
  },
  plugins: [
    new webpack.CommonsChunkPlugin({
      minChunks: 3,
      name: 'vendor'
    })
  ]
}

```

I am not quite sure how the current way containers are being lazy loaded in this project effects the CommonsChunk behavior.  I'm guessing it just works? but I still need  to look into it more 


Below are some blog posts I learned a few things from, but TBH it is all still magic to me

- http://jonathancreamer.com/advanced-webpack-part-1-the-commonschunk-plugin/ 

- http://henleyedition.com/implicit-code-splitting-with-react-router-and-webpack/

sokra:
> i had a feeling that CommonsChunkPlugin plugin wont trace on demand file load dependency graph

yes that's true, by default the CommonsChunkPlugin will only process entry chunk.

But there is an async mode for the CommonsChunkPlugin which use on-demand chunks: http://webpack.github.io/docs/list-of-plugins.html#4-extra-async-commons-chunk

``` js
new CommonsChunkPlugin({
  name: "app",
  async: "async-commons"
});
```

It'll load on-demand when any on-demand chunk is requested. (Both load parallel)

datapimp:
@sokra by on demand chunk do you mean chunks that get generated by `require.ensure` ?

tsm91:
@sokra thx, next time we should read the fucking manual :F

Basically this options hash to CommonsChunkPlugin solves everything: 

```js
    new webpack.optimize.CommonsChunkPlugin({
      name: 'app', // will extracts common chunks of code to app.js
      children: true,
      async: true, // follow on demand require (this is the important bit)
      minChunks: 3,
    }),
```

We don't need to specify every page to the entry point, which is very good.

```js
  entry: {
    // ...
    // the below code to manually keep track of pages is not needed
    homePage: path.join(process.cwd(), 'app/containers/HomePage/index.js'),
    featurePage: path.join(process.cwd(), 'app/containers/FeaturePage/index.js'),
    notFoundPage: path.join(process.cwd(), 'app/containers/NotFoundPage/index.js'),
  },
```


cc @datapimp actually i have a modified fork of this boilerplate and i use CommonsChunkPlugin, i might check out this DLL solution.

You are right about the dev environment, it just slows down recompilation time.

mxstbr:
> Basically this options hash to CommonsChunkPlugin solves everything:

If I understand that correctly, the only thing we need to do to fix chunking would be to add `async: true` to the `CommonsChunkPlugin` options? And then the optimal solution would be to use DLL in development and the above in production?

tsm91:
@mxstbr nope, i guess @datapimp plans to do this ddl stuff for prod. But i haven't read anything about the dll solution yet.

We don't need CommonsChunkPlugin for dev, only to prod builds.

datapimp:
hey @tsm91 sorry for the confusion, I added the proper `CommonsChunkPlugin` config here:

https://github.com/mxstbr/react-boilerplate/pull/470/commits/c02ea4095a468f60356c547f73bc1b88dc20ed0c

I think we should use that in production. DLL is a development only config.

tsm91:
@datapimp ahh okay, cool

so we can discard this pr and merge the dll one?

also you could add the vendor entry [8e1f273](https://github.com/mxstbr/react-boilerplate/pull/472/commits/8e1f273bebef1e4c8ad40922d9d48895900d6b2f), the 1st commit of this pr

justingreenberg:
> DLL is a development only config.

while this implementation may be for dev only, i would just point out that DLLs can also be extremely useful for breaking large apps into "sub-apps" (ie with different dev teams, repos) as well as many other production use cases, ie targeting caching (tracking asset hashes in linked manifests) or a/b testing

datapimp:
@justingreenberg thanks - I suspected as much I just haven't had a chance to dig deeper into it and the js file that gets generated seems too big and I didn't wanna be responsible for putting that on people's production sites

Would love to learn more from you about this

datapimp:
@tsm91 is it necessary? i was hoping Webpack automatically handled that for us

tsm91:
@datapimp hmm, i guess you are right

mxstbr:
I'm not sure I'm following the discussion here, I think we can close this PR?

datapimp:
These config changes have been made in the DLL PR.  Safe to close.

mxstbr:
Thanks for kicking this off @tsm91, even though it didn't get merged!

PR #475, opened 2016-06-05
Remove prefixed CSS from our source code.

christianvuerings:
Since we use cssnext [1], we don't need to write these prefixes
ourselves. Cssnext includes autoprefixer [2], which will automatically
add the necessary prefixes.

[1] http://cssnext.io/
[2] http://cssnext.io/features/#automatic-vendor-prefixes


mxstbr:
Solid, good catch @christianvuerings, thanks!

Issue #174, opened 2016-02-15
Universal rendering


mxstbr:
Adding server-side rendering to this boilerplate is probably the n1 request I get, and it does look quite useful.

Me personally, I haven't played with that at all, so I'm not yet sure if I want to include it by default or not. What do you think?

https://gist.github.com/rgrove/3e612aa366541845161c

luandro:
Hey, I've done quite a lot of Universal rendering setup in the past year. You can check the approach I've been using [here](https://github.com/luandro/hapi-universal-redux), which is mainly inspired by [react-isomorphic-staterkit](https://github.com/RickWong/react-isomorphic-starterkit).

I guess these are the challenges for SSR:
- server-side live-reloading
- server-side pre-fetching with Redux

I've seen the server reloading problem being solved in a few ways:
- use [webpack-dev-middleware]() and [webpack-hot-middleware]() if in `development` as seen [here](https://github.com/jaredpalmer/react-production-starter), or
- use [concurrently](https://github.com/kimmobrunfeldt/concurrently) and [just-wait](https://github.com/download/just-wait) and different webpack config files, as seen [here](https://github.com/RickWong/react-isomorphic-starterkit)

The best api for the pre-fetching problem I've seen has been with [redial](https://github.com/markdalgleish/redial).

Those two options seem to solve all the problems discussed on the [gist](https://gist.github.com/rgrove/3e612aa366541845161c). Am I missing something, or are these to naive?

I'd be glad to give it a try, let me know what you think.

mxstbr:
We're really close to the [v3 release](https://github.com/mxstbr/react-boilerplate/issues?q=is%3Aopen+is%3Aissue+milestone%3Av3.0) with just a few bugfixes left. I expect we officially release and launch it in (at the very latest) a month. This version has seen a major architectural overhaul and massive improvements all around.

I don't think introducing another overhaul for v3 is worth it, but server side rendering is something we definitely need to at least try, and probably release with v4. (?) I'd personally **love to see a first implementation**, just don't expect your PR to be merged immediately!

> Note: If you're going to try this, work on the **v3.0.0 branch** please!

luandro:
I'm gonna try to start working on this after #211 settles.

mxstbr:
As per #211, we should provide alternative implementations in other backend systems. @oliverturner might be able to do a Hapi one, but before we start we need a clear set of specs that all systems need to cover.

[...]

PR #479, opened 2016-06-06
Make ngrok usage opt-in instead of default

datapimp:
This makes the ngrok tunnel feature opt-in instead of the default behavior.

To enable the tunnel you can `npm run start:tunnel` as usual.

Arguments could be made against

- overloading ngrok which is a free service
- against opening up people's development machines to outside traffic 
  without them explicitly asking and being aware.

Although I'm not really passionate about these.

My motivation is mainly that I don't use it often enough and I am not sure most 
people would need this feature at the boilerplate phase of their project, and can easily
enable it in the future if and when they do need to use it.

If this gets merged, we can make ngrok an optional dependency.


mxstbr:
Hmm, not quite sure we don't want to make this a default. It's an interesting suggestion for sure, but I fear it might hide the feature behind a large wall of commands…

tsm91:
i think sometimes less is more, this is one of the features i immediately removed from my fork of react-boilerplate

furthermore i also removed the whole express and webpack-middleware machinery, i got rid of a bunch of code and devDependency

also, no one will serve static content via express.js, they will do it via nginx, one less layer and much more efficient (they will run the nodes as a reverse proxy)

the whole development server is [6 sloc](https://github.com/tsm91/react-seed/blob/master/tools/webpack/webpack.dev.babel.js#L52-L57) in the fork

my opinioon: we got the whole implementation lets write a wiki page about it and let it go

mxstbr:
> also, no one will serve static content via express.js, they will do it via nginx, one less layer and much more efficient (they will run the nodes as a reverse proxy)

So, the thing is, we introduced express.js because it was the first step towards server-side rendering (ref #174), which is the next big topic we'll be looking into.
I love to hear how easy it was for you to rip all of the stuff out you don't need, and we need to work on keeping that as easy because that's really important, but the express implementation is definitely staying.

> My motivation is mainly that I don't use it often enough and I am not sure most 
people would need this feature at the boilerplate phase of their project, and can easily
enable it in the future if and when they do need to use it.

I agree with you. I'll let this sit for a few more hours and think about it, but I'm positive this is a good change! Thanks @datapimp!

tsm91:
@mxstbr ahh ok

mxstbr:
Thanks @datapimp for keeping us real, love it!

datapimp:
@frixid can you share details?

current master:

screenshot 2016-06-09 04 14 30

you will need to `npm run start:tunnel` directly as it is no longer the default when running `npm start`.

I'll make sure to update any missing documentation

frixid:
@datapimp 
captura de pantalla 2016-06-09 a las 13 42 10

datapimp:
Delete ngrok and reinstall it looks like an error in their script

PR #480, opened 2016-06-06
Fix the generator only works when indentation is 2.

ZhuGongpu:
Remove the magic numbers and the problem is solved.


mxstbr:
Why use `\s{0,}` vs. `\s*`? Don't they achieve the same thing?

ZhuGongpu:
`\s*` may contain other characters.

mxstbr:
I've restarted TravisCI, waiting for it to finish then I'll merge this. Thanks @ZhuGongpu!

mxstbr:
Thanks @ZhuGongpu, this is great! :tada:

PR #490, opened 2016-06-07
Add image compression to webpack process

Ndpnt:
https://github.com/mxstbr/react-boilerplate/issues/489

### DIFF
--- a/internals/webpack/webpack.base.babel.js
+++ b/internals/webpack/webpack.base.babel.js
@@ -36,7 +36,10 @@ module.exports = (options) => ({
       loader: 'file-loader',
     }, {
       test: /\.(jpg|png|gif)$/,
-      loader: 'file-loader',
+      loaders: [
+        'file-loader',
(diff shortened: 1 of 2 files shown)


Ndpnt:
I open this PR too quickly, documentation is missing. I close it until I add related documentation

mxstbr:
Awesome, thanks @Ndpnt!

Ndpnt:
The way I add this feature, it optimizes images both in development and production mode.
@mxstbr do you think it is a good idea as it allows us to see how images are compressed during our development process and potentially adjust settings or do you think it could slowdown things and it's better to compress images only for production?

mxstbr:
> do you think it is a good idea as it allows us to see how images are compressed during our development process and potentially adjust settings or do you think it could slowdown things and it's better to compress images only for production?

Hmm, good question actually, would you mind opening a new issue to discuss this?

Ndpnt:
Yes, I going to do this :)

PR #497, opened 2016-06-08
Support npm install --production. Fixes #491

### LINKED ISSUE #491: Server in production needs packages from devDependencies
After installing dependencies in our production server with `npm install --production`, when we run `npm run start:prod`, the command fails because some packages can't be found:
- `cross-env`,
- `ip`,
- `ngrok`,
- …

Do we miss something or there are some packages which should be moved from `devDependencies` to `dependencies`?

datapimp:
Needed to add cross-env and ip as production dependencies, and I needed to change the way the production server started up so that it did not depend on either webpack or ngrok


datapimp:
ugh; is there a way for me to add the linter as a pre-commit hook without making that a requirement for the project?

mxstbr:
> ugh; is there a way for me to add the linter as a pre-commit hook without making that a requirement for the project?

Hmm, that should happen automatically, we use `lint-staged` to make that happen—@okonet any ideas why that could not be happening?

datapimp:
Closing this; I have fixed these in the PR and don't feel like messing w/ merges;

PR #501, opened 2016-06-08
Add prompt to ask if the new reducer should be registered as static

harijoe:
[#331](https://github.com/mxstbr/react-boilerplate/issues/331)


mxstbr:
This won't work as easily, because we codesplit every container, meaning this reducer might be imported twice. (would have to double check)

Instead, lets get rid of static reducer importing in general!

PR #509, opened 2016-06-09
Remove static reducer from generator

harijoe:
[#331](https://github.com/mxstbr/react-boilerplate/issues/331) and [former pull request #501](https://github.com/mxstbr/react-boilerplate/pull/501#issuecomment-224637849)

### DIFF
--- a/internals/generators/container/index.js
+++ b/internals/generators/container/index.js
@@ -110,18 +110,6 @@ module.exports = {
         templateFile: './container/reducer.test.js.hbs',
         abortOnFail: true,
       });
-      actions.push({ // Add the reducer to the reducer.js file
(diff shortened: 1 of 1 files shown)


mxstbr:
This LGTM, I just want to get one last look at this from other collaborators before I merge this.

@oliverturner @somus @datapimp @tsm91 any thoughts against this change? We're code splitting containers, so adding the reducers to the static reducers list doesn't really make sense imo, so I think this is a good change :+1:

tsm91:
LGTM

oliverturner:
Sounds right, looks good 👍

mxstbr:
Awesome, thanks @harijoe!

PR #515, opened 2016-06-10
Cache node_modules for faster travis builds

yanglinz:


mxstbr:
I'm not entirely sure we really want this, we want to test against the newest versions if possible.

Is there any way to cache out-of-date node modules, but download only the updates everytime?

SilentCicero:
@yanglinz can you explain more how the caching system works, I agree with @mxstbr if this caches old modules, then I wouldn't want this in the build. Is there an override system?

mxstbr:
I think it updates them, since it'll still run `npm install` which overrides the out-of-date modules. This sounds like a great change, though I'm still not 100% sure on the semantics and implications…

mxstbr:
I think it updates them, since it'll still run `npm install` which overrides the out-of-date modules. This sounds like a great change, though I'm still not 100% sure on the semantics and implications…

yanglinz:
Hi @mxstbr @SilentCicero

Seems you guys are right, caching `node_modules` might have unintended behaviors with npm's dependency resolution(https://github.com/travis-ci/travis-ci/issues/3758). 

I'll close this for now, and possibly re-open after a bit of research :)

mxstbr:
Awesome, thanks @yanglinz!

PR #519, opened 2016-06-10
Fixed webm mimetype

cornedor:
webm and mp4 were both using video/mp4 as mimetype. I've added a seperate loader for webm.


mxstbr:
Good catch @cornedor, I missed that! Thanks!

Issue #524, opened 2016-06-11
Reading index.html from fileSysytem  on everyRequest , wont it hurt performance


sap9433:
In `middleware/frontendMiddleware` , 
`app.get('*', (req, res) => res.sendFile(path.join(options.output.path, 'index.html')));`

On every request we are reading index.html from fileSystem.

Even if  express `sendFile` will check it's internal cache (which I'm not sure) , it will still need to do a cacheCheck from some hashfile.

shouldn't we preload entire html in some variable `const indexHtml =  "<html>.....</html>`"`and send that variable instead..

mxstbr:
Yeah, that sounds reasonable!

FezVrasta:
What if you edit `index.html` while the server is running?

sap9433:
we will change only in `addProdMiddlewares` .

watson:
I'm pretty sure `res.sendFile` does not employ any caching. It's using the "send" module internally [which specifically states](https://github.com/pillarjs/send#caching) that it doesn't perform caching. The `res.sendFile` function it self [doesn't seem to have any caching either](https://github.com/expressjs/express/blob/f3d99a4fdbe1531ea609e92c4d4ae6c269d78c7a/lib/response.js#L385-L420).

It is using a stream though, which is the correct way to handle file I/O in Node. So it's non-blocking and will therefore not have any significant impact on the rest of the application as long as the event loop isn't saturated. But it does increase the time it takes for the client to receive the file of course.

I'm not aware of any function in Express similar to `res.sendFile` that have caching. So this is something you'd have to build your self. I would personally employ lacy loading using the [thunky](https://github.com/mafintosh/thunky) module if I needed to do this, but there are several possible solutions of course.

mxstbr:
Thanks for chiming in @watson, glad to hear it's not a big performance impact!

mxstbr:
Closing due to little interest and tiny perf impact.

PR #525, opened 2016-06-11
In dev mode All request matching is not required #524

sap9433:
In dev mode we are using webpackMiddleware , and webpackHotMiddleware , so they take care of serving us the App, and hot reloading . Generic request matcher is not required . This is effectively dead code and App should work as is without it.


sap9433:
sorry I missed all request matching part

PR #530, opened 2016-06-11
removed useless globals

FezVrasta:


FezVrasta:
Oh, didn't spotted that test... It's the only place where we use chai...

I'm rewriting it to use `expect`

FezVrasta:
Ok I guess we should keep Chai, it's useful for this kind of stuff...

mxstbr:
Ahhh, `chaiEnzyme` is the reason we have it in there! Then the question becomes, shouldn't we get rid of `mjackson/expect` instead?

FezVrasta:
I like the way spy functions are managed by mjackson/expect

mxstbr:
True. Thanks @FezVrasta!

PR #532, opened 2016-06-11
Rename `common.js` to `common` since CommonsChunkPlugin auto-appends `.js`

jperasmus:
A small fix that simply ensures webpack's CommonsChunkPlugin to generate a `common.js` bundle instead of `common.js.js`.


mxstbr:
Thanks @jperasmus!

PR #533, opened 2016-06-11
Allow use of Mocha grep option

FezVrasta:
This allows the use of `npm run test -- --grep ComponentName`


mxstbr:
Nice, I like it! :+1:

mxstbr:
Thanks! :tada:

elcojacobs:
When I try this the resulting command is:
`npm run test -- --auto-watch --no-single-run "--grep" "Button"`

`--grep` is incorrectly quoted. Where should I look for fixing this?

Issue #349, opened 2016-05-20
Server isn't compatible with Node v0.12


chappydboz:
hey guys, first time trying to run the boilerplate.. im on osx yosemite. 
i got thru the npm install.. but here is the console output after i run "npm start":

```
> react-boilerplate@3.0.0 start /Shared/www/tacotown/tacofinder/react-site
> npm run start:tunnel


> react-boilerplate@3.0.0 start:tunnel /Shared/www/tacotown/tacofinder/react-site
> cross-env NODE_ENV=development node server

/Shared/www/tacotown/tacofinder/react-site/server/index.js:25
app.listen(port, (err) => {
                       ^^
SyntaxError: Unexpected token =>
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:501:10)
    at startup (node.js:129:16)
    at node.js:814:3

npm ERR! Darwin 14.5.0
npm ERR! argv "node" "/usr/local/bin/npm" "run" "start:tunnel"
npm ERR! node v0.12.4
npm ERR! npm  v3.9.2
npm ERR! code ELIFECYCLE
npm ERR! react-boilerplate@3.0.0 start:tunnel: `cross-env NODE_ENV=development node server`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the react-boilerplate@3.0.0 start:tunnel script 'cross-env NODE_ENV=development node server'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the react-boilerplate package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     cross-env NODE_ENV=development node server
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs react-boilerplate
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls react-boilerplate
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /Shared/www/tacotown/tacofinder/react-site/npm-debug.log

npm ERR! Darwin 14.5.0
npm ERR! argv "node" "/usr/local/bin/npm" "start"
npm ERR! node v0.12.4
[...]

Issue #175, opened 2016-02-16
Switch to express.js


mxstbr:
Instead of using Webpack Dev Server, we should switch to express.js for the development and "production" server. (devserver breaks on complex nested routes, see #113)

I have never worked with express, so I'd rather somebody with a bit more experience fixes this, a PR would be very welcome. (on the `v3.0.0` branch)

amorino:
Hey, I just read your twitter message. I will try to work on this this week. :D

mxstbr:
In progress at #194!

grabbou:
@amorino do you need any help with that? I was thinking about setting up express.js as well.

@mxstbr not sure if that's been considered yet, but what about server-side rendering? Not a big deal nowadays, but still once we integrate `express`, it should be possible to write it as a middleware that is completely optional (and maybe even developed in a separate repo)

mxstbr:
@grabbou see #174, this is planned down the road, but I figured it'd require a complete architecture overhaul which is something that's definitely not happening for v3 – are you telling me that's not true? It'd be amazing to be able to switch in-n-out of server-side rendering very easily!

grabbou:
Well, I **might** be wrong, but back in the day when I was working on Este, we were having it as a middleware (to do all heavy processing) with just few tricks required. It probably still requires some changes, but maybe we can somehow workaround them :D

mxstbr:
I'd love if you could take a stab at that and provide a WIP PR!

oliverturner:
Coupling this project so completely to Express is worrisome to me. Completely get the motivation, but would make a plea to consider a more modular approach.

grabbou:
Could `express.(dev|prod).js` file be potentially a solution for that?

mxstbr:
The project isn't really coupled with express, I don't think any express-specific features need to be used here. Switching it out for a different node.js server framework shouldn't be too hard, but I'm not the expert – what do you think @grabbou?

grabbou:
Yeah, that's true. It might be just that some people would prefer sticking with plain webpack and maybe shipping HTML on its own? If that's the case, we can provide express as a submodule with its own dependencies that you can either plug-in or delete?

mxstbr:
We're going to switch to express anyway because of a problem with `WebpackDevServer`, see #194, so feel free to go ahead and implement SSR in express!

oliverturner:
[...]

PR #546, opened 2016-06-14
Add babel-watch and es6 support on the server

harijoe:
This PR proposes the addition of `[babel-watch](https://github.com/kmagiera/babel-watch)`, which brings two improvements:
- The tool reloads very quickly the server on every file change, (this is a necessary condition to allow people to extend the express server and make an API out of it as suggested in `index.js`, today only used to handle HMR)
- The tool allows the usage of ES6 in the server by transpiling the code on the fly

Notice: `babel-watch` should not be used in production. The server should be built with babel, shipped and then run with only just node. This has not been implemented and suggestions are very welcome to address it :)


mxstbr:
This looks like a solid start, I'm not yet 100% sure we really want to introduce another build step.

*Would fix #349*

harijoe:
I think this build step is necessary to grant auto restart of the server. May you try it and notice it is actually very fast. 

What would be the downside of adding such another build step ?

mxstbr:
No, I mean another build step when compiling for production, I don't mind the development. :grin:

harijoe:
Okay, I understand. What would be the next step to see this PR merged ? Other reviews ?

mxstbr:
Yeah, I'm not really a server person so I have no idea if this is reasonable or not to be honest.

*Let's ping some that chimed into the initial Express.js discussion in #175 and the related PRs: @grabbou @amorino @oliverturner*

oliverturner:
My practice has been to run nodemon watching for file changes in `/app` and `/server` and restarting the server accordingly.

Since the client state is unaffected this has no impact on the time taken to recompile via babel: it's effectively a background operation. It's simple, robust and doesn't add further complexity to our babel config.

oliverturner:
No slight intended on your approach btw @harijoe: this is simply my approach and rationale

oliverturner:
A big difference is that typically I'm not running webpack as server middleware though... `babel-watch` may be a better fit for this environment. Wd love to hear from @grabbou on his perspective...

grabbou:
Hey, great work here! I am not using Babel on Node.js side of things recently, mainly because I am running on 6.x that has a decent support for almost all features.

As @oliverturner proposed, I am also using `nodemon` as an opt-in to reload app on code changes. As far as I am aware of kmagiera module, it's a more performant way to reload server when you have to recompile it with babel. So it seems like a good solution if there's transpilation going on on the server side of things.

Having said so, I am not really sure if ES6/beyond are needed on the server side here as an extra step given that the scope of it is rather small. But I can see some use-cases for that, esp. when the app grows a bit.

harijoe:
@grabbou As this project is meant to be a strong base for new projects, I think it makes sense to enable ES6 on server side. I guess many users will want to enrich the provided server at some point. I find it cool to provide them the right tools to do so.

mnickkk:
This project is definitely missing this feature, however I don't think that this PR resolves this correctly. 
Backend changes cause whole frontend rebuild (and remember that initial build takes time). Should probably spawn two separable tasks (maybe use https://github.com/kimmobrunfeldt/concurrently I've seen this approach in one boilerplate for sure)?
These two packages might be useful:
https://github.com/jlongster/monkey-hot-loader
https://github.com/ericclemmons/webpack-hot-server-example

jolyonruss:
If it seems this PR is too prescriptive, maybe it should be closed and the docs updated to show usage of Nodemon, Superviser or Forever?

mxstbr:
Since there is quite a bit of discussion left to do before we can merge something like this and this PR is quite outdated I'll close this for now. Thank you @harijoe for providing this great reference implementation, that's definitely much appreciated! Sorry we didn't merge it.

ithinco:
This feature is fine to merge, though I'm looking for a method to add this feature as a configure file maybe "webpack.server.babel.js".

rclai:
Has there been any new developments on this?

harijoe:
I think it's better to wait for #1236 to be merged first. This branch can still be used as an implementation reference for anyone not willing to wait though.

rclai:
I don't know if anyone is running into this but why is it that code like this on the server:

```
router.get('/test', async (req, res) => {
  res.json({ hello: 'world' });
});
```

Runs fine in dev mode (`npm start`) but then it will give me an error in production mode (`npm run start:prod`) _even_ though I changed this line from `package.json` from this:

`"start:prod": "cross-env NODE_ENV=production node server`

to this

`"start:prod": "cross-env NODE_ENV=production babel-watch server --presets es2015 react stage-0 --plugins transform-runtime",`

Kr1an:
@rclai, or anyone else who has this problem when node code works in dev(with different features of es6/beyond - import/export/async/await/...) and does not work with the same stack in production:
You need to understand that in development its normal to use such tools as bable-watch bable-node, but in production these tools are't effective enough. So in production, code should be native. To make it native, you should do same process that you do with front-end code - transpile it to version thats supported by node6+(current).
If your server-side code locates in **server** directory, run script `babel server --out-dir dist --presets es2016` It will compile all unsupported features and create new version in **dist** folder. and then you just run `node dist/index.js`.

Actually, this approach has its cons: new code is less suitable for debugging because sometimes easy to read new feature transforms to horrible staff, but if you need it only for production, than its normal, i suppose.

In case of react-boilerplate, you just need to change start:production script with the steps from above. I think it does not brake any relative connections because transpiled folder locates on same level.
Good luck with it.

rclai:
I can't remember anymore why I was using babel-watch in production. It kind of baffles me now that I think about it, but I'm not doing that anymore. I decided to make the server code run on a separate project.

PR #553, opened 2016-06-14
Changing order of fs callback args to error then file

zanedev:
Changing order of fs callback args to error then file fixes https://github.com/mxstbr/react-boilerplate/issues/548


datapimp:
good catch.  will merge asap

datapimp:
@mxstbr not sure what is up with appveyor if there is any way to bypass it to push this simple fix

datapimp:
thanks alot @zanedev i merged this in another PR

PR #555, opened 2016-06-15
Refactor main saga to make side effects more explicit

justingreenberg:
so this breaks `getGithubSaga` into three smaller, simpler sagas and follows a more "idiomatic" redux-saga patterns like how to create and manage (cancel) async watchers. note this really does not implement any additional functionality since it's basically what the race effect is doing behind the scenes, however imo these patterns are more illustrative and can easily adapted for other use-cases ie. denormalizing and caching... 

this PR implements the new sagas and provides full test coverage

`getRepos` - Github repos request/response handler
`getReposWatcher` - Watches for LOAD_REPOS action and calls handler
`githubData` - Root saga manages watcher lifecycle


app/containers/HomePage/sagas.js
```diff
@@ -2,43 +2,54 @@
...
+ */
+export function* getReposWatcher() {
+  while (yield take(LOAD_REPOS)) {
+    yield call(getRepos);
   }
 }
 
+/**
+ * Root saga manages watcher lifecycle
+ */
```
justingreenberg: it may be worth noting that this is the point of contact between saga and route handler, since this seems to be the cause of a lot of confustion

app/containers/HomePage/sagas.js
```diff
@@ -2,43 +2,54 @@
...
 
-// Individual exports for testing
-export function* getGithubData() {
-  while (true) { // eslint-disable-line no-constant-condition
-    const watcher = yield race({
-      loadRepos: take(LOAD_REPOS),
-      stop: take(LOCATION_CHANGE), // stop watching if user leaves page
-    });
-
-    if (watcher.stop) break;
```
justingreenberg: also this never felt right to me

mxstbr:
This looks very solid, awesome @justingreenberg!

PR #558, opened 2016-06-15
Updated documentation for setting a custom port

tomazy:
The `-p` option doesn't work


mxstbr:
Ah yeah, we forgot to update that! Thanks!

PR #559, opened 2016-06-15
Update Redux dev-tools extension's store

MrOutis:
If you use Redux devTools extension, since v2.0.1, they added an `updateStore`, so any enhancer that changes the store object could be used with the devTools' store.

As this boilerplate uses Redux and Redux-Saga, the `updateStore` is needed if you want to `take` actions in your Sagas, dispatched from devTools.

For more information, refer to the [release v2.0.1][release].
[release]: https://github.com/zalmoxisus/redux-devtools-extension/releases/tag/v2.0.1

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -42,6 +42,15 @@ openSansObserver.check().then(() => {
 const initialState = {};
 const store = configureStore(initialState, browserHistory);
 
+// If you use Redux devTools extension, since v2.0.1, they added an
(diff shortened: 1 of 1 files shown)


harijoe:
Oh, I wondered why the sagas weren't triggered from the devTools. This would be a very nice improvement. 👍

mxstbr:
Nice on @MrOutis, thanks!

PR #560, opened 2016-06-15
let url loader infer mimetype from file extension

Moeriki:
See [url-loader README](https://github.com/webpack/url-loader).


mxstbr:
Thanks!

PR #572, opened 2016-06-18
Fix eslint errors

SimonSelg:
react/jsx-curly-spacing was not respected in some places.

```
... /react-boilerplate/app/components/A/index.js
  15:7   error  There should be no space after '{'   react/jsx-curly-spacing
  15:18  error  There should be no space before '}'  react/jsx-curly-spacing

... /react-boilerplate/app/components/H1/index.js
  7:37  error  There should be no space after '{'   react/jsx-curly-spacing
  7:48  error  There should be no space before '}'  react/jsx-curly-spacing

... /react-boilerplate/app/components/H2/index.js
  7:37  error  There should be no space after '{'   react/jsx-curly-spacing
  7:48  error  There should be no space before '}'  react/jsx-curly-spacing

... /react-boilerplate/app/components/H3/index.js
  5:9   error  There should be no space after '{'   react/jsx-curly-spacing
  5:20  error  There should be no space before '}'  react/jsx-curly-spacing
```

PR #573, opened 2016-06-18
Fix comment about how to use custom middlewares

### LINKED ISSUE #562: Adding your own custom backend-api is in the wrong place.
Started with this boilerplate to try it out. 
Just a note that I noticed that could confuse someone, when trying to add an api in the project. This is not the place to add an api because app is undefined. see: https://github.com/mxstbr/react-boilerplate/blob/master/server/index.js#L13

It needs to be in the frontendMiddleware or maybe middlewares directory needs a index.js file to structure different middlewares?

My solution for playing around at the moment has been to add an api directory in middlewares and require in frontendMiddleware.js and add `app.use('/api', api);` here: https://github.com/mxstbr/react-boilerplate/blob/master/server/middlewares/frontendMiddleware.js#L64

SimonSelg:
This fixes #562

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -9,11 +9,13 @@ const isDev = process.env.NODE_ENV !== 'production';
 const ngrok = (isDev && process.env.ENABLE_TUNNEL) || argv.tunnel ? require('ngrok') : false;
 const resolve = require('path').resolve;
 
+// In production we need to pass these values in instead of relying on webpack
+const app = express();
+
 // If you need a backend, e.g. an API, add your custom backend-specific middleware here
 // app.use('/api', myApi);
 
-// In production we need to pass these values in instead of relying on webpack
-const app = setup(express(), {
+setup(app, {
   outputPath: resolve(process.cwd(), 'build'),
   publicPath: '/',
 });


SimonSelg:
The CI Build should work w/o errors once #572 gets merged or the bug is fixed by other means.

mxstbr:
Nice one! Could you rebase onto master, then we'll get this merged!

alikhani:
hmm.. @SimonSelg will that work? The reason I put in here [L64](https://github.com/mxstbr/react-boilerplate/blob/master/server/middlewares/frontendMiddleware.js#L64) in In the frontendMiddleware is because of [L34](https://github.com/mxstbr/react-boilerplate/blob/master/server/middlewares/frontendMiddleware.js#L34)

That line is going to swallow all request which means that you'll receive a 404 when requesting /api, right? Or am I missing something?

SimonSelg:
@mxstbr: Done!

@alikhani: Good catch! Accidentally placed it in the wrong line. Fixed now. Thanks.

mxstbr:
Thank you!

PR #577, opened 2016-06-19
Add link to speech from dev conference to Readme

jeggett:
Video contains speech "Scaling React.js Applications"  delivered
by Mas Stoiber at oredev conference 2016. It provides insights
about scalable design of React applications which used in
react-boilerplate.


mxstbr:
This looks perfect, thanks so much!

PR #586, opened 2016-06-21
Manifest.json formatting fix

### LINKED ISSUE #580: manifest.json invalid
Hi,

First of all, thanks for all the hard work that's gone into this.

After doing a build (clean install), the `manifest.json` file appears to be wrapped as a module:

```
module.exports = {
	"name": "React Boilerplate",
	"icons": [
        ...
```

This in turn causes errors with the chrome devtools, and the app doesn't get recognised as a PWA:

Screenshot of devtools

PS: I'm not an webpack expert, so I might be missing something.

bnhovde:
Changed the loader for manifest.json from json-loader to file-loader
to prevent it being wrapped with module.exports

Used approach as described here:
https://github.com/webpack/webpack/issues/1747#issuecomment-164561113

Fixes #580

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -10,7 +10,7 @@ import 'babel-polyfill';
 
 // Load the favicon, the manifest.json file and the .htaccess file
 import 'file?name=[name].[ext]!./favicon.ico';
-import 'file?name=[name].[ext]!./manifest.json';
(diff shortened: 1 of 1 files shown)


mxstbr:
Nice oneeee, thanks!

PR #592, opened 2016-06-22
Do not run lint-staged on ignored files

okonet:
Updated lint-staged to the latest version that supports npm run tasks. Added `lint:eslint` task without paths that is used by `lint:js` and `lint-staged` to properly ignore files.


mxstbr:
Awesome, thanks @okonet!

Issue #598, opened 2016-06-24
Pin dependencies and enable Greenkeeper


mxstbr:
Currently we rely on SemVer of our dependencies, which might not always work. (one of the webpack beta updates in our range recently broke our build) We should pin the dependencies (a PR for this would be appreciated!) and enable @greenkeeperio (https://greenkeeper.io) instead!

gihrig:
@mxstbr `greenkeeper enable` used to pin deps automatically. [see here](https://www.youtube.com/watch?v=XbQvxCLkBbw) @0:55 but this seems not to be the case presently. Don't know why, sure seems like a good idea. For reference see a lodash GK PR [here](https://github.com/lodash/lodash/commit/fa29123c5e32e4bd07e1a3a80a11e5331f36d4a2)

I enabled GK on my react-boilerplate fork, without pinning deps, took about 5 minutes, totally automated.
GK [Getting Started](https://greenkeeper.io/#getting-started) tells everthing you need to do (three commands).

Here is [GK's PR](https://github.com/gihrig/react-boilerplate/pull/1/commits/52daa7d74cb70dd46bc46707027947fbdbf06598) on my fork.

Some things to note:

![image](https://cloud.githubusercontent.com/assets/1481063/16350353/196adcc4-3a13-11e6-8c50-67804be22a83.png)

GK sometimes ignores old deps that have been updated - Current stylelint is 6.7.1
GK sometimes updates to latest including breaking changes - stylelint-config-standard 8.0.0 -> 9.0.0 causes `Undefined rule "selector-pseudo-class-no-unknown"`. This is due to the outdated style-lint.

All in all, the changes required to accommodate GKs update were minor.

After making the required changes to get the project working and passing all tests, `npm install` reports many warnings, but I guess this is to be expected. The whole world can't be expected stay up to date. ;-)

![image](https://cloud.githubusercontent.com/assets/1481063/16351129/3dcd2f22-3a18-11e6-890d-c4b489284707.png)
![image](https://cloud.githubusercontent.com/assets/1481063/16351142/5afec006-3a18-11e6-8e6b-ce1b3b201dc3.png)

I'll open a PR momentarily.

gihrig:
@mxstbr I notice the GK PRs are getting a bit backlogged. I would be happy to take on checking and merging these on an on-going basis, if you would like.

mxstbr:
Yes, please do!! 

(Closing this since it's done)

gihrig:
[...]

PR #602, opened 2016-06-24
Gk prep

gihrig:
Prepare for Greenkeeper
- update package.json deps as determined by GK
- pin all deps
- fix minor breakage
ref #598 

Should be ready to [install GK](https://greenkeeper.io/#getting-started)


mxstbr:
Thanks so much @gihrig!

Issue #320, opened 2016-05-18
Problem linting CSS


thijsw:
I checked out the project, run the setup. Then, when I run the command `npm run test` (which calls `npm run lint:css`) it fails when compiling the css files.

```
app/components/A/styles.css
 2:10  ✖  Expected "ADDD" to be "addd"      unit-case
 2:10  ✖  Unexpected unknown unit "ADDD"    unit-no-unknown
 6:10  ✖  Expected "CC0E5" to be "cc0e5"    unit-case
 6:10  ✖  Unexpected unknown unit "CC0E5"   unit-no-unknown

app/components/Button/styles.css
 24:21  ✖  Expected "ADDD" to be "addd"     unit-case

npm ERR! Darwin 15.4.0
npm ERR! argv "/usr/local/Cellar/node/6.2.0/bin/node" "/usr/local/bin/npm" "run" "lint:css"
npm ERR! node v6.2.0
npm ERR! npm  v3.8.9
npm ERR! code ELIFECYCLE
npm ERR! react-boilerplate@3.0.0 lint:css: `stylelint ./app/**/*.css`
npm ERR! Exit status 2
npm ERR!
npm ERR! Failed at the react-boilerplate@3.0.0 lint:css script 'stylelint ./app/**/*.css'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the react-boilerplate package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     stylelint ./app/**/*.css
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs react-boilerplate
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls react-boilerplate
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/thijs/Projects/react-boilerplate/npm-debug.log

npm ERR! Darwin 15.4.0
npm ERR! argv "/usr/local/Cellar/node/6.2.0/bin/node" "/usr/local/bin/npm" "run" "lint"
npm ERR! node v6.2.0
npm ERR! npm  v3.8.9
npm ERR! code ELIFECYCLE
npm ERR! react-boilerplate@3.0.0 lint: `npm run lint:js && npm run lint:css`
npm ERR! Exit status 1
npm ERR!
[...]

PR #609, opened 2016-06-28
Add empty style file so stylelint doesn't fail the build after a clean.

marcdel:
I was also confused by the stylelint error discussed in #320. It seems to me an improved developer experience would be getting a green build after running a `git run clean` on a new clone.

It wasn't immediately obvious to me that the tests were failing because there were no css files in the app or even that this would be a failure case, so it would be nice to set newer developers up for success.

This is the least intrusive workaround I could come up with, but I realize it isn't super intuitive either (hence the comment). Let me know what you think, or if you have other suggestions!

### DIFF
--- a/app/style.css
+++ b/app/style.css
@@ -1,0 +1,4 @@
+/*
+  This stylesheet is not linked anywhere.
+  It is included here so that stylelint will not fail the build when no other css files are present.
+*/
--- a/app/style.css
+++ b/app/style.css
@@ -1,4 +1,0 @@
-/*
-  This stylesheet is not linked anywhere.
-  It is included here so that stylelint will not fail the build when no other css files are present.
-*/
--- a/internals/scripts/clean.js
+++ b/internals/scripts/clean.js
@@ -29,7 +29,7 @@ mkdir('app/containers/App');
 mkdir('app/containers/NotFoundPage');
 mkdir('app/containers/HomePage');
 cp('internals/templates/appContainer.js', 'app/containers/App/index.js');
-cp('internals/templates/appContainerStyles.css', 'app/containers/App/styles.css');
+cp('internals/templates/styles.css', 'app/containers/App/styles.css');
 cp('internals/templates/notFoundPage.js', 'app/containers/NotFoundPage/index.js');
 cp('internals/templates/homePage.js', 'app/containers/HomePage/index.js');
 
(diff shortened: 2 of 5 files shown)


mxstbr:
Hmm, what about adding that next to the default component?

marcdel:
Not sure what you mean by default component. The components folder is removed during clean, no? 
Unless you mean the app container, which I considered, but figured it might encourage global styles more than you'd like.

mxstbr:
Yeah that's true, hmm. Let me have a think about this!

gihrig:
> Not sure what you mean by default component. The components folder is removed during clean, no? 

The components folder is removed and repopulated on `npm run clean`, so a basic styles css could be added to` /internals/templates` folder and then copied to `app/containers/App` in `clean.js`. 

The stub css file could be as simple as a header comment:
```
/**
 * styles.css 
 *
 * App container styles
 */
```
This solves the issue of this PR and gives us a project 'stub' after `clean` that passes tests and displays a page in the browser after `npm start`. A win win IMO.

By describing it as 
> App container styles

I don't think that encourages global styles, thoughts?

mxstbr:
Yeah, I do prefer that solution too @gihrig. Would you mind updating the PR that way @marcdel?

marcdel:
Sure, I'll update the PR in a moment.

However, one thing I noticed will spiking that solution the other day is that the linter will fail if you import the file but don't reference it. So we can copy the file to `app/containers/App` and stop there, or we can include it in the appContainer template and give it an empty `.container {}` style or something to slap on the div so it doesn't complain.

gihrig:
I like the empty .container style. It creates a smoother experience for first time users. But when I create that file with:
```
/* App container styles */
.container {

}
```
and `npm run lint:css` I get an error:
```
app/containers/App/styles.css
 2:12  ✖  Unexpected empty block   block-no-empty
```
Placing a comment in the `.container` block removes the error
/* App container styles \*/
.container {
  /* Place styles for App here if needed */
}

marcdel:
@gihrig I couldn't leave it empty, unfortunately. I went with `display: block` since it's a div, but it could be anything really.

mxstbr:
Thanks @marcdel, this is awesome!

marcdel:
Woo! 🎉

PR #612, opened 2016-06-29
Rename utils/hooks to asyncInjectors

zavan:
I'm a naming freak and I don't think `hooks` say much about what are those functions and this file. So, sometime ago I've renamed it in my project, and now I'm porting it here so you guys can choose if you want to merge it.

-
BTW, thanks for the awesome project, it helped me a lot to grasp all those libraries and plugins.


mxstbr:
This sounds very reasonable, thanks so much @zavan! @justingreenberg what do you think?

justingreenberg:
this looks good to me 👍  nice work @zavan!

zavan:
Cool! Thank you guys!

PR #613, opened 2016-06-30
Missing "!" in app.js (template) manifest import

skosno:
Template of app.js is missing "!" when importing manifest.json - it makes manifest wrongly served and this not working after "npm start".


mxstbr:
Perfect, thanks @skosno! #586

PR #614, opened 2016-06-30
Do not remove app/components folder on clean

kenips:
Generator complains that there is no app/components after cleaning.


kenips:
Just noticed that it wasn't deleting the components folder to begin with. Not sure how it went wrong yet.

gihrig:
I just cloned, setup and cleaned a new copy of the repo, `components` existed and was populated as expected.

Mac OS 10.11.5
Node 5.8.0
Npm 3.8.2

PR #625, opened 2016-07-01
Server Side Rendering

datapimp:
This is still a WIP but I am submitting it because some people were interested in seeing an approach to server side rendering. 

Essentially the way it works is that the root container of the app is a `react-helmet` component that is responsible for rendering the entire HTML payload.  This component is used to render the page on the server side by using `react-router` match function to handle the express request object and render the given page.

Credit for this approach goes 100% to another boilerplate: https://github.com/DominicTobias/universal-react/

A few other changes have been made such as using webpack to compile and run the server code.  

I will clean this up and improve upon it over the weekend hopefully.
## TODO
- [ ] improve webpack server config and make server side hmr work 
- [ ] improve documentation 
- [ ] add tests
- [ ] usage example: dynamic request handler in production
- [ ] usage example: static site generator

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ stats.json
 .DS_Store
 npm-debug.log
 .idea
+lib/
--- a/app/app.js
+++ b/app/app.js
@@ -26,64 +26,94 @@ import configureStore from './store';
 // Observe loading of Open Sans (to remove open sans, remove the <link> tag in
 // the index.html file and this observer)
 import styles from 'containers/App/styles.css';
-const openSansObserver = new FontFaceObserver('Open Sans', {});
 
-// When Open Sans is loaded, add a font-family using Open Sans to the body
-openSansObserver.check().then(() => {
-  document.body.classList.add(styles.fontLoaded);
-}, () => {
-  document.body.classList.remove(styles.fontLoaded);
-});
+const isClient = typeof document !== 'undefined';
+
+if (isClient) {
+  const openSansObserver = new FontFaceObserver('Open Sans', {});
+
+  // When Open Sans is loaded, add a font-family using Open Sans to the body
+  openSansObserver.check().then(() => {
+    document.body.classList.add(styles.fontLoaded);
+  }, () => {
+    document.body.classList.remove(styles.fontLoaded);
+  });
+}
 
 // Create redux store with history
 // this uses the singleton browserHistory provided by react-router
 // Optionally, this could be changed to leverage a created history
 // e.g. `const browserHistory = useRouterHistory(createBrowserHistory)();`
 const initialState = {};
-const store = configureStore(initialState, browserHistory);
 
-// If you use Redux devTools extension, since v2.0.1, they added an
-// `updateStore`, so any enhancers that change the store object
-// could be used with the devTools' store.
-// As this boilerplate uses Redux & Redux-Saga, the `updateStore` is needed
-// if you want to `take` actions in your Sagas, dispatched from devTools.
-if (window.devToolsExtension) {
-  window.devToolsExtension.updateStore(store);
-}
+const createStore = (state = initialState) => {
+  const store = configureStore(
+     state, browserHistory
+  );
+
+  // If you use Redux devTools extension, since v2.0.1, they added an
+  // `updateStore`, so any enhancers that change the store object
+  // could be used with the devTools' store.
+  // As this boilerplate uses Redux & Redux-Saga, the `updateStore` is needed
+  // if you want to `take` actions in your Sagas, dispatched from devTools.
+  if (window && window.devToolsExtension) {
+    window.devToolsExtension.updateStore(store);
+  }
+
+  return store;
+};
 
 // Sync history and store, as the react-router-redux reducer
 // is under the non-default key ("routing"), selectLocationState
 // must be provided for resolving how to retrieve the "route" in the state
 import { selectLocationState } from 'containers/App/selectors';
-const history = syncHistoryWithStore(browserHistory, store, {
-  selectLocationState: selectLocationState(),
-});
+
+const syncedHistory = (store) => (
+  syncHistoryWithStore(browserHistory, store, {
+    selectLocationState: selectLocationState(),
+  })
+);
 
 // Set up the router, wrapping all Routes in the App component
 import App from 'containers/App';
 import createRoutes from './routes';
-const rootRoute = {
-  component: App,
-  childRoutes: createRoutes(store),
+
+export const buildRootRoute = (component, store) => ({
+  component, childRoutes: createRoutes(store),
+});
+
+export const clientRenderer = (components) => {
+  const { history, store, rootRoute } = components();
+
+  ReactDOM.render(
+    <Provider store={store}>
+      <Router
+        history={history}
+        routes={rootRoute}
+        render={
+          // Scroll to top when going to a new page, imitating default browser
+          // behaviour
+          applyRouterMiddleware(useScroll())
+        }
+      />
+    </Provider>,
+    document.getElementById('app')
+  );
+
+  // Install ServiceWorker and AppCache in the end since
+  // it's not most important operation and if main code fails,
+  // we do not want it installed
+  require('offline-plugin/runtime').install(); // eslint-disable-line global-require
 };
 
-ReactDOM.render(
-  <Provider store={store}>
-    <Router
-      history={history}
-      routes={rootRoute}
-      render={
-        // Scroll to top when going to a new page, imitating default browser
-        // behaviour
-        applyRouterMiddleware(useScroll())
-      }
-    />
-  </Provider>,
-  document.getElementById('app')
-);
+if (isClient) {
+  clientRenderer(() => {
+    const store = createStore();
 
-// Install ServiceWorker and AppCache in the end since
-// it's not most important operation and if main code fails,
-// we do not want it installed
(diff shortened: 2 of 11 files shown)


tzarger:
@datapimp Hi Jonathan, this is an excellent start. I was looking in to start working on this too and the first thing I was looking into was how to handle redux-saga on server side. Are you planning on working that into the ssr? Thanks for the great start on this, I look forward to see your refinements.

datapimp:
My goal is definitely to have as little of difference as possible in the way you think about redux-saga, just not sure how to go about enforcing some of the constraints that you have to put on yourself to be able to do that.  

I added isomorphic-fetch for example.  Would love some suggestions or tips from people with battle scars from doing this already

tzarger:
@datapimp Jonathan, since redux-saga is a bit of a bear to use server side, almost hack like at best, have you used or looked into https://github.com/makeomatic/redux-connect for loading up state server side? redux-saga is definitely nice, but not sure it is the best choice when doing a universal application.

mxstbr:
`redux-connect` is great, but doesn't quite solve the same problem as `redux-saga` does, specifically tying together decoupled components. (see [this talk](https://vimeo.com/168648012) at 14:40 for the saga specific part, but I encourage you to watch it in its entirety)

I'm strongly convinced that `redux-saga` will shape the way we build frontend applications for a long time, and I would very much despise getting rid of it!

justingreenberg:
hey guys, @datapimp solid work as usual. i haven't had a time to checkout locally and review but wanted to provide a resource and one quick thought before i head out for the weekend

redux saga ssr reference implementation:
https://gist.github.com/sompylasar/5e7157e451f4b7268def9ae1ce01edd4#file-serverrenderloop-js

this is a solid pattern for handling render loop. you will immediately notice that it is not trivial lol

this was also helpful in understanding, but keep in mind our routes are async and injected:
https://github.com/yelouafi/redux-saga/blob/master/examples/real-world/server.js

in our case, invoking sagas from router should mitigate some of the constraints associated with ssr but requires some consideration. for example, our current implementation will work, but it's not actually doing anything... eg, we would want to set a default username and dispatch a load repos action, either from componentWillMount or using put effect in root saga... we will also need to explicitly close the open take channel on success.. just thinking out loud here

one other quick thought on SSR—

getting client/server parity for asset name resolution is a serious challenge... ie

```js
import Banner from './banner-metal.jpg';
```

obviously this will not work in native node environment, and if it did the asset name wouldn't match the one generated by webpack, breaking checksum when mounting on client... canReuseMarkup throws

in my experience, **the simplest possible solution is to monkey `require`** and use webpack's `loader-utils` to manually interpolate the output name.... here's a gist: 
https://gist.github.com/justingreenberg/f1735abffbae3778014f8d14b599a2ef#file-require-patch-js

```js
```

this is a hack, but i have found it very reliable... alternatively, since you are propsing we use webpack to compile server code as well, we could leverage webpack itself to handle these assets during compilation... i've worked with various webpack plugins/hacks with mixed experiences:
https://github.com/halt-hammerzeit/webpack-isomorphic-tools
https://github.com/exponentjs/asset-module-webpack-plugin

**TL;DR: if we are using webpack to compile server, i recommend using https://github.com/jchip/isomorphic-loader**

totally did not mean to write so much, now i really gotta run.. let me know if you guys have any questions :) have a great weekend

datapimp:
@justingreenberg thanks a lot, thanks for the advice, I will definitely read everything.

@mxstbr don't worry i feel just as strongly about redux-saga

datapimp:
@justingreenberg  any thoughts on the following approach?

Essentially container components / routes would declare a static method 

```javascript
export class Container extends Component {
  static readyOnActions (dispatch) {
    return Promise.all([ ])
  }
 
  componentDidMount () {
    Container.readyOnActions(this.props.dispatch)
  }
}
```

The server renderer would have access to the route's getComponent, getChildRoutes, getComponents, etc and any of them which have this method would then resolve the various promises before returning the final markup.

I think this might eliminate the need for a render loop but I am still digging into that code you sent to get a better grasp.

Stalso:
@datapimp If I understand approach with `static readyOnActions` correctly, then it is implemented in [MERN](https://github.com/Hashnode/mern-starter). You can look at it. But if you look at  [this](https://github.com/yelouafi/redux-saga/blob/master/examples/real-world/server.js) code, you will see, how to use Saga `END` effect for ssr. Maybe it is reasonable to use this approach? And also we can use `redial ` with Saga, like it is described here -  https://github.com/yelouafi/redux-saga/issues/293 .

I am also trying to implement ssr for this boilerplate, and the main challenge for me is require of `sanitize.css` in `\app\containers\App\index.js`. It always fails on server side. Maybe it is because this css in externals during server-side build, and server-side webpack loaders for css (I have tried to use many approaches, [fake-style-loader](https://github.com/dferber90/fake-style-loader), for example) do not work? I do not know. I ended up with ignoring this `css` on server side like this:

    if(__CLIENT__)
      require('sanitize.css/sanitize.css');

And I was happy, until I have installed [react-flexbox-grid](https://github.com/roylee0704/react-flexbox-grid) which internally requires `flexboxgrid.css` . Of course I can not ignore it. And everything was broken. And no loaders helped me to correctly load this css.

 For application styles my loaders look like this and work:

    'css-loader/locals?localIdentName=[local]__[path][name]__[hash:base64:5]&modules&importLoaders=1&sourceMap!postcss-loader'

I do not know the reason, but images are loaded correctly using the same loaders as on client side. 

Also I have looked at your code and do not understand, how will you debug server side with your approach, and why do not you use [webpack-externals-plugin](https://github.com/Morhaus/webpack-externals-plugin)?

oyeanuj:
@datapimp Came across another server-side rendering with redux-saga attempt that might have some useful ideas: https://github.com/xkawi/react-universal-saga

tzarger:
@oyeanuj Yes, I just saw that too. Looks like it might do the trick here. Some things it is lacking is code splitting, etc. Not sure if that would be an issue since sagas are injected... I really would like to see an option here for SSR, does not need to be baked in by default, but a configuration would be sweet. It seems that SSR movement here has slowed down, and hope it is not dead. @datapimp

tzarger:
@datapimp @justingreenberg @Stalso Do you know if anyone is working on this?

Stalso:
@tzarger I am currently working on this,  but I am doing something wrong.  Let more experienced people solve this problem.  As I said in previous post,  css it the main problem for me,  when I try to make webpack build of the app for server rendering

oyeanuj:
@justingreenberg @datapimp Any thoughts on @Stalso's previous comment with the CSS as that seems a big blocker at the moment?

kachkaev:
What is the current state of this PR? Feels like after this feature is merged into the boilerplate it'll leave all the competitors so far behind!

What are you guys using to achive SSR in your React/CSS-modules projects meanwhile?

kachkaev:
This repository might be a useful source for some ideas:
https://github.com/reactGo/reactGo

Just found it via http://andrewhfarmer.com/starter-project/, a pretty nice little app for choosing a boilerplate. The list seems to be up to date (20 Aug 2016), but is not 100% complete (react-create-app is missing, for instance).

scf4:
@Stalso Did you manage to solve your issue?

davidwilde:
I found this article about isomorphic react isomorphic - https://reactjsnews.com/isomorphic-react-in-real-life

An application that I'm working on was inspired by this boilerplate but I wanted it to be isomorphic because SEO is particularly important. My own application basically ripped out all of the CSS module stuff because I couldn't solve it at the time - and time constraints yadda yadda yadda.

However possibly following this guide, could lead to an isomorphic version of this application.

gihrig:
@davidwilde That looks like a great article, thanks for sharing!

Another step in the 'universal' app direction, [react-engine](https://github.com/paypal/react-engine) recently came to my attention. Looks like it may solve some of the 'pain' points discussed in the above article.

someoneinatree:
I've been looking at approaches for this, and managed to pull together a half-working (i.e. JS for reduxified routes, no working CSS yet) SSR implementation on my react-boilerplate project (forked ~4 months ago). This pull request being the most helpful for what I ended up doing! (Besides realizing I had no clue what webpack was actually doing)

Can check it out here: https://github.com/opensessions/opensessions

The important files being `internals/webpack/webpack.server.babel.js`, `package.json` for the `npm run build:server` command, `server/src/server.js`, and any of the working pages `app/containers/{HomePage,SessionView,ListSessions}/index.js`.

asood123:
Wanted to mention this post here: https://medium.com/open-collective/asking-for-a-new-feature-on-an-open-source-project-3305deba1356#.6ehgqgzhc.

mxstbr:
Since this is very out of date and will highly likely not be merged, I'll close it for now. Thanks for the start, this is a great reference point @datapimp!

scf4:
@asood that sounds great. @mxstbr can you confirm this is legit?

mxstbr:
This is totally legit.

tzarger:
@mxstbr @asood123 Can we drop a message in the Gitter channel to everyone to see if we can raise the $850 mentioned in the article? I think it would be a great opportunity to get that done after all this time and would be interested in donating as well.

HTeker:
Will this be merged in the master branch any time soon?

Jayiitb:
@mxstbr Are we working on this?

gretzky:
This is currently benched for the time being.

PR #638, opened 2016-07-04
Typo fixes

dashawk:


mxstbr:
"instant" is actually correct in this case. :grin: Thanks for taking the time to submit a PR though!

PR #648, opened 2016-07-05
Include `childRoutes` example

JulienMelissas:
As promised, here is an example of how to use childRoutes in `routes.js`. See https://github.com/mxstbr/react-boilerplate/issues/499#issuecomment-230145069


mxstbr:
Awesome, this is great!! Thanks @JulienMelissas!

PR #657, opened 2016-07-06
Should not commit code for users.

jeregrine:
I did this on accident and lost a load of staged changes. Feels really bad.

### DIFF
--- a/internals/scripts/clean.js
+++ b/internals/scripts/clean.js
@@ -62,10 +62,4 @@ rm('-rf', 'internals/templates');
 
 process.stdout.write(' ✓');
 
-// Commit the changes
-if (exec('git add . --all && git commit -qm "Remove default example"').code !== 0) {
-  echo('\nError: Git commit failed');
-  exit(1);
-}
-
-echo('\nCleanup done. Happy Coding!!!');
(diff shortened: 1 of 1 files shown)


gihrig:
Please explain more about the steps that caused you to lose staged files. 

I made changes, staged them then executed `git add . --all` and `git commit -qm "Remove default example"` My changes were committed to my local repo as expected.

OS X 10.11.5
git version 2.7.4 (Apple Git-66)

mxstbr:
> I did this on accident and lost a load of staged changes. Feels really bad.

I'm very very sorry, but I'm not 100% clear on how this made you loose any changes? Why would you run `npm run clean` after you've made changes to your app?

jeregrine:
@mxstbr because I assumed it cleaned the build folder :P not remove all files. I will admit it was dumb to not read the code of the thing first.

@gihrig this command deletes all files that are non-generated, then commits. 

Just really bad practice to ever commit without user consent.

jeregrine:
To be specific if you look at this https://github.com/mxstbr/react-boilerplate/pull/657/files#diff-0643478c0f20dbb3ca2bf263399d5963L27 it removes everything, then re-creates it.

PR #661, opened 2016-07-07
issue 656 build:clean should clean everything

bcowgill:
also handles issue 628 clean coverage before testing
stats.json build/.htaccess and coverage now cleaned up on a build
analyze:clean target for stats.json
test:clean target for coverage
build:clean invokes them all.


mxstbr:
Awesome!

One small nitpick, we don't run `npm run analyze` when we build so we shouldn't clean it when we build either, and should probably be in `preanalyze` so it runs before the analyze script.

Thanks so much!

bcowgill:
ok made those changes. added a clean:all target to run all the cleans if you just want to delete everything built.

mxstbr:
Awesome, thanks @bcowgill!

PR #662, opened 2016-07-07
add trailing slash to publicPath in faq

olapersson:
the publicPath needs to end with a /, else the main.js file returns the content of index.html


mxstbr:
Awesome, good catch! Thanks @olapersson!

PR #670, opened 2016-07-08
Improve offline plugin configure and add favicon to html webpack plugin

suhaotian:
Add favicon option to HtmlWebpackPlugin.

Add offline plugin to webpack.dev.js without cache anything.

Beacuse after `npm run start:production`, then run `npm start` will make a error:

net::ERR_INSECURE_RESPONSE error.

Add absolute path to OfflinePlugin, issue see here: [https://github.com/NekR/offline-plugin/issues/77](https://github.com/NekR/offline-plugin/issues/77)


mxstbr:
Wait, why do we enable the `offline-plugin` in dev mode? We don't really want that?

suhaotian:
We don't cache anything.
I use it in dev mode, because when run `npm run start:production` and stop it. Then start dev server.Reload, the browser console will throw a error for `offline-plugin`, you can try it. 
And add offline plugin to dev, solve the problem.

gihrig:
Is this the cause the console warning in a normal dev mode page load?

> offline-plugin: runtime.install() was executed without OfflinePlugin being added to webpack.config.js. The installation of offline-plugin has not been performed. You can ignore this message if you are skipping the OfflinePlugin during development. Otherwise, you may have a configuration issue. runtime.js?11f5:7

suhaotian:
Close it.

NekR:
Just a quick note: it's fine to use `offline-plugin` in dev mode. I use it in dev mode in my projects. The problem is that `offline-plugin` doesn't support Hot Reloading and cannot reasonably support it. This is why I think it's better to not include it in dev configs, at least in boilerplate projects.

suhaotian:
Yeah, !

williamgranli:
I'm getting the following errors in the browser console:
```
The script has an unsupported MIME type ('text/html'). (sw.js:1)
Failed to load resource: net::ERR_INSECURE_RESPONSE (http://localhost/sw.js)
```

I started getting this error in dev mode after trying to run the project in production mode. Could anyone confirm if it's related to this issue? If it is, could you also clarify how to solve the error? Not clear to me from above comments.

Thanks!

GGAlanSmithee:
@williamgranli could be related to https://github.com/NekR/offline-plugin/issues/62#issuecomment-255528668. I am still having this problem (not sure if working as expected or not). For now, after testing `sw` locally in production mode, I delete it from `chrome://serviceworker-internals`

PR #679, opened 2016-07-08
Express proxy

theWaR13:
Added plugin express-http-proxy for proxy requests. It helps to avoid CORS problems when fetching data from remote server. You can specify host and port this way:
npm start --pxhost=192.168.0.1 --pxport=3001

By default pxhost == 127.0.0.1, pxport == 3000

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,6 +125,7 @@
     "chalk": "1.1.3",
     "compression": "1.6.1",
     "express": "4.13.4",
+    "express-http-proxy": "^0.7.3",
     "fontfaceobserver": "1.7.1",
     "history": "3.0.0",
     "immutable": "3.8.1",
--- a/server/index.js
+++ b/server/index.js
@@ -14,10 +14,6 @@ const app = express();
 const pxhost = process.env.npm_config_pxhost || '127.0.0.1';
 const pxport = process.env.npm_config_pxport || '3000';
 
-
- console.log({ pxhost: pxhost });
- console.log({ pxport: pxport });
-
 // If you need a backend, e.g. an API, add your custom backend-specific middleware here
 // app.use('/api', myApi);
 
@@ -32,7 +28,7 @@ const port = argv.port || process.env.PORT || 3000;
 
 // Proxy requests
 app.use('/api', proxy(`${pxhost}:${pxport}/`, {
-  forwardPath: function(req, res) {
+  forwardPath: function (req, res) {
     return require('url').parse(req.url).path;
   }
 }));
--- a/server/index.js
+++ b/server/index.js
@@ -27,11 +27,7 @@ setup(app, {
 const port = argv.port || process.env.PORT || 3000;
 
 // Proxy requests
-app.use('/api', proxy(`${pxhost}:${pxport}/`, {
-  forwardPath: function (req, res) {
-    return require('url').parse(req.url).path;
-  }
-}));
+app.use('/api', proxy(`${pxhost}:${pxport}/`));
 
 // Start your app.
 app.listen(port, (err) => {
--- a/server/index.js
+++ b/server/index.js
@@ -1,6 +1,7 @@
 /* eslint consistent-return:0 */
 
 const express = require('express');
+const proxy = require('express-http-proxy');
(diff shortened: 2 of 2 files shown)


mxstbr:
I like the way you did this, but I don't think we'll be merging this. Including this by default isn't necessary for our example app, and I'd rather have a documentation page explaining how to add this.

Thanks so much for taking the time to submit this PR though, much appreciated! :tada:

gihrig:
👍  for a documentation page showing how to add.

chaintan17:
Thank you!

ddltechy:
@theWaR13 I've tried your suggestion to add that plugin but I cannot make it work. I've installed express-http-proxy and added the changes in server/index.js. My backend is running in localhost:8000, it's a json-server, and the React app is running in the default 3000 port. When I enter localhost:3000/api, it just returns Page Not Found. 

I'd appreciate any help on this. 
Thanks in advance!!!

chaintan17:
@ddltechy it works for me. I thinks you should give us your code change so we could help look into it.

ddltechy:
@chaintan17 I'm using default react-boilerplate version 3.0.0 and only changed server\index.js as suggested by @theWaR13  (I'm running on Win 7)

**server\index.js**

const express = require('express');
const proxy = require('express-http-proxy');
const logger = require('./logger');

const argv = require('minimist')(process.argv.slice(2));
const setup = require('./middlewares/frontendMiddleware');
const isDev = process.env.NODE_ENV !== 'production';
const ngrok = (isDev && process.env.ENABLE_TUNNEL) || argv.tunnel ? require('ngrok') : false;
const resolve = require('path').resolve;

const app = express();

const pxhost = process.env.npm_config_pxhost || '127.0.0.1';
const pxport = process.env.npm_config_pxport || '8000';

// If you need a backend, e.g. an API, add your custom backend-specific middleware here
// app.use('/api', myApi);

// In production we need to pass these values in instead of relying on webpack
setup(app, {
  outputPath: resolve(process.cwd(), 'build'),
  publicPath: '/',
});

// get the intended port number, use port 3000 if not provided
const port = argv.port || process.env.PORT || 3000;

// Proxy requests
app.use('/api', proxy(`${pxhost}:${pxport}/`));

// Start your app.
app.listen(port, (err) => {
  if (err) {
    return logger.error(err.message);
  }

  // Connect to ngrok in dev mode
  if (ngrok) {
    ngrok.connect(port, (innerErr, url) => {
      if (innerErr) {
        return logger.error(innerErr);
      }

      logger.appStarted(port, url);
    });
  } else {
    logger.appStarted(port);
  }
});


Thanks!!!

theWaR13:
@ddltechy Hello. I've faced the same problem. In my case POST requests worked good, but when i tried GET request (for example open in browser localhost:3001/api) i received not found as well. To fix this problem you should do the next thing:
1. Remove 
```
const proxy = require('express-http-proxy');

const pxhost = process.env.npm_config_pxhost || '127.0.0.1';
const pxport = process.env.npm_config_pxport || '3000';

// Proxy requests
app.use('/api', proxy(`${pxhost}:${pxport}/`));
```

2. Paste this strings in server/middlewares/frontendMiddleware.js

ddltechy:
@theWaR13 I forgot to mention that I was doing a GET. That change made it work!  Thank you very much!!!

theWaR13:
@ddltechy you see welcome ;)

xtrinch:
Just gonna leave this here, since for those who are also searching for websocket proxy support, this will be much more appropriate:

```
const proxy = require('http-proxy-middleware');

app.use('/', proxy({
  target: `http://${pxhost}:${pxport}`,
  changeOrigin: true,
  logLevel: 'debug',
  ws: true,
}));
```

tahifaMed:
i tried all the suggestion in this post and nothing seems to work for me, i tried first to add the code in server/index.js and next to add it on server/middleware/frontendMiddleware.js
just wanna know, on the saga call i want to call locahost:8080/api/test
what format should i mention on requestURL ?
requestURL = http://localhost:8080/api/test or just api/test
please help me im stuck in this subject for a while now

Issue #684, opened 2016-07-08
ES6 class generator eslint disable


BraunreutherA:
The generator for creating components with es6 classes could directly set the `// eslint-disable-line react/prefer-stateless-function` flag so there wouldn't be the need to do it it manually every time.

gihrig:
Great idea :sparkles:

Would you consider submitting a PR?

BraunreutherA:
Jap made PR #685 - had to change only for the components in the container generator it's already present.

mxstbr:
Done in #685, thanks so much @BraunreutherA!

PR #685, opened 2016-07-08
ES6 class eslint disabling

BraunreutherA:
Adding the flag for disabling eslint on ES6 classes


mxstbr:
Thank you @BraunreutherA!

Closes #684

PR #687, opened 2016-07-08
refactor(match-action): create Action class and refactor reducer using pattern matching

Jiansen:
I created an **Action** class so that we could

1.  avoid referencing constants across multiple files.  (there is no need for constants.js files, always import from actions.js files when you were thinking about an action )

1. use the *.action* method as standard action creator, rather than defining an action creator for each action by hand

1. in reducer, instead of using switch-case to check action type, we use pattern matching on the name of action object (see the example below). 

For example, `HomePage\action.js` now only contains two lines of code.
```
import Action from 'utils/action';

export const ChangeUsername = new Action('ChangeUsername');
```
and the *homeReducer* function now looks like as follows
```
function homeReducer(state = initialState, action) {
  return match(action)(
    CASE(ChangeUsername, ({ payload }) =>
      // Delete prefixed '@' from the github username
      state.set('username', payload.replace(/@/gi, ''))
    ),
    CASE(otherwise, state)
  );
}
```

Using the **Action** class is completely optional.  It is here to help reducing repeated work.  Likewise, using the *match* function is also optional.   

This pull request suggests a way to refactor the code.  A user may even mix use the current code style or the proposed feature, whichever is appropriate.

Please take a look at the changes and let me know the part that you like and don't like.   

TODO Tasks if PR is accepted in principle
- [ ] refactor the generator script likewise, or provide a new generator as an alternative to the current one.
- [ ] when generate an action, also generate a unique action name/type.
- [ ] other received suggestions

In this PR, I updated both the HomePage container and the App container to demonstrate my ideas.  If you like the idea.  The final PR may only contain changes to one of the 2 containers.  As I said earlier, this PR is to provide an alternative.

Thanks :-)


mxstbr:
While I definitely like the way you handled this, I don't think we'll merge this generally. It adds a lot of complexity that might confuse people since this is a non-standard way of doing actions, which I'd like to avoid.

I still think this is a great way of doing actions if the entire team is in on it since it reduces a lot of boilerplate. Have you thought about releasing this as a standalone npm module?

Jiansen:
Thanks for feedback @mxstbr .

The changes are caused by using the **Action** class and the *match* method from the *[match-js](https://www.npmjs.com/package/match-js)* library.

The Action class, as you see, only contains 20 lines of code.  We save lines of code because action creator is provided by the Action class.

I guess the confusion at first sight, if any, is the match method in reducer, and wondering how it works under the hood.  The good news is, we don't have to write the reducer function using the match method.  We can use exactly the same code as the current boilerplate!

I will take your advise and think about release the Actor class as a standalone npm module.  I am very likely to use your react-boilerplate repo to demonstrate steps of refactoring.  My question is, apart from that 20 lines of code and a reference to an example (this PR?), what else would you suggest me to provide in that npm module?

mxstbr:
Some tests maybe? Not sure how much testing can be done, but at least that the action and the constants work.

PR #688, opened 2016-07-09
Update deps

gihrig:
Dependency roll-up, bypassing GreenKeeper

This PR is BROKEN!!
================

GK PR "Update fontfaceobserver to version 2.0.1" breaks the project but all tests and build pass in GK, Travis, AppVeyor and Local (Mac OS).

Execution in browser produces console error:

`Uncaught TypeError: openSansObserver.check is not a function localhost/:75`

This PR is submitted to demonstrate the hazard of merging GK PRs that pass without further human inspection. Is it possible/practical to create automated integration testing?


gihrig:
PR is FIXED
----------------

- Brings all packages up to latest (July 8 2016 @16:46 PDT)
- Fixes breakage caused by `fontfaceobserver 2.0.1`
- Tested in dev and production, including tunnel, setup, clean and generate (does not fix existing generate issues #337)
- Triggers warning on browser console in dev mode only: 

> offline-plugin: runtime.install() was executed without OfflinePlugin being added to webpack.config.js. The installation of offline-plugin has not been performed. You can ignore this message if you are skipping the OfflinePlugin during development. Otherwise, you may have a configuration issue. runtime.js?11f5:7

mxstbr:
Awesome, thanks so much @gihrig!

>  Is it possible/practical to create automated integration testing?

We probably should, but the problem is that it shouldn't live in this repo since that's what users clone. Any ideas?

gihrig:
@mxstbr [Percy](https://github.com/integrations/percy) was the only thing that looked close on GitHub's Integrations page. It does cloud based UI testing, not exactly what I was looking for, but if the example app works as expected, I guess that's a pretty good integration test.

Percy integrates with Travis, [setup](https://percy.io/docs) looks to be straightforward 

If this looks like something you would consider, I'll be glad to look into it in more detail.

mxstbr:
Please do @gihrig!

gihrig:
As of today, percy still does not [support](https://percy.io/docs) React.

PR #713, opened 2016-07-13
Remove a css block that is not being used and is causing syntax warnings

kachkaev:


mxstbr:
Thanks, not sure how I missed this for so long, my bad!

PR #733, opened 2016-07-17
ISSUE-732 - Fix default selector test template to pass

chaintan17:
https://github.com/mxstbr/react-boilerplate/issues/732


gihrig:
What is the status of this? Work in progress? Working fix?

chaintan17:
@gihrig it's my working solution.

I get your idea that error would remind programmer to include the test.
Anyway, it look the error i found is a mistake in coding.
I think we should fix this first, because newbie like me would take a lot of time try to figure it out what's went wrong during the test.

chaintan17:
Update default generated unit test to throw error to user that he/she didn't specify test case yet.

image

mxstbr:
This looks very reasonable, thanks so much @chaintan17! 🎉

PR #743, opened 2016-07-18
Change build:clean for build instead

qzapaia:
build:clean script removes the "build" folder and makes server crash in heroku.


sethmcleod:
Since `postinstall` is already set to `npm run build:dll` *(which does nothing on prod)*, we could instead use Heroku's specific postbuild step like this: 

>*Step 3:* Add this line to your Package.json file in the scripts area: `"heroku-postbuild": "npm run build"`

Also *(and this may be unrelated)* I was having trouble deploying to Heroku because `npm run build` was triggering `postbuild` which was triggering `npm run test`. Karma was failing and crashing the build because there is no Chrome binary on the platform.

image

If you *need* to run Karma tests as part of the build step, you may be able to fix this with a custom Heroku buildpack that includes Chrome *(similar to this: https://github.com/dwayhs/heroku-buildpack-chrome)* but if not, you can fix this by just removing the test script so it looks like this: `"prebuild": "npm run build:clean",`.

After that I was finally able to deploy to Heroku successfully 🎉

qzapaia:
??

jgrancher:
@sethmcleod is right, here is chunk of my package.json I have that made it working:
```js
    ...
    "preinstall": "npm run npmcheckversion",
    "postinstall": "npm run build:dll",
    "prebuild": "npm run build:clean",
    "heroku-postbuild": "npm run build",
    ....
```

mxstbr:
I like @sethmcleod's solution, would you mind submitting a new PR with that change? Thanks!

PR #744, opened 2016-07-18
Altered injectAsyncSagas to prevent injecting duplicate sagas

hansoksendahl:
Duplicate sagas were being injected on Location_Change events.

Background:
This [issue](https://github.com/mxstbr/react-boilerplate/issues/384) was closed but no bug fix was submitted.  This PR implements the fix mentioned in issue 384 by @kraffslol.

### DIFF
--- a/app/utils/asyncInjectors.js
+++ b/app/utils/asyncInjectors.js
@@ -14,7 +14,14 @@ export function injectAsyncReducer(store) {
  * Inject an asynchronously loaded saga
  */
 export function injectAsyncSagas(store) {
-  return (sagas) => sagas.map(store.runSaga);
+  const prevSagas = [];
+
+  return sagas => sagas.forEach(saga => {
+    if (!prevSagas.includes(saga)) {
+      store.runSaga(saga);
+      prevSagas.push(saga);
+    }
+  });
 }
 
 /**


mxstbr:
Wait, didn't @justingreenberg fix this already?

kraffslol:
@justingreenberg solution was decided upon where I think, the duplication was "intented" and the way to avoid it was by running a Race in the saga. @justingreenberg seems to have changed to another approach now which seems to be these lines in [sagas.js](https://github.com/mxstbr/react-boilerplate/blob/master/app/containers/HomePage/sagas.js#L40-L50). I'm currently using his approach in my own project and it seems to work out fine. So I don't think this pull is needed.

mxstbr:
I don't think so either, this should be fixed on the sagas side instead! Take a look at the watcher pattern we use in the example saga and do the same thing in yours. Thanks for taking the time to submit this PR though, we appreciate it!

zanedev:
If it's harmless I like this solution as a site wide fix rather than having to put in every saga but I guess it's a matter of preference.

gihrig:
If the _right way_ is less than obvious, or or if there's reasons for several approaches, sounds like an excellent topic for a FAQ entry.

Further, the _Root saga manages watcher lifecycle_ code block is not included in generated container's saga.

PR, anyone?

mxstbr:
True, that'd be a good addition @gihrig!

justingreenberg:
the current code reflect idiomatic saga patterns and can easily be scaled and tailored to any purpose. that said—given the scope, the same could (should probably) be accomplished without redux-saga at all... i think many people misconstrue the saga code as a scaffold or recommendation on general saga usage—which is definitely not the case. imo far too much emphasis being placed on the saga implementations... i feel like it's like a distraction or something. in reality, the app is just a foundation for building scalable apps; with a tiny demo included for reference (as @kraffslol as done)

some ideas...

- explicitly document **that the saga code is not a copy paste solution**. this is a new style for many people and the type of issues we're seeing reflect a general lack of understanding, so maybe we could urge caution in using at all without 

- maybe consider providing some "built in" sagas, ie xhr handler with incremental backoff, that can be both used by the demo app and provided as a utility to end users of boilerplate

- we should also probably create an issue template that eg. directs redux-saga related questions to stack overflow, encourages people to search closed issues, etc  😄

hansoksendahl:
@justingreenberg I strongly support the implementation of the run, watch, once API you mentioned.  It provides flexibility and helps users new to sagas to reason about them.

It can be difficult to debug issues related to sagas being injected multiple times in the page.  I think this API would help new users avoid these bugs.

gihrig:
As @justingreenberg commented

> given the small scope of demo app, the same thing could be accomplished without redux-saga at all. i think many people misconstrue our saga implementation as being some sort of durable data-loading solution—which it is most definitely not

I did not realize this (still new to react/redux/saga). Given that, it's unfortunate that while containers can be generated without sagas, routes can not. This, and the errors that ensue when a saga is removed from a route, give the impression that sagas are a required feature of the bp.

mxstbr:
> explicitly document that the demo saga is not a general solution. this is a new control style for many people and the type of issues we're seeing reflect a general lack of understanding, so maybe we could urge caution in using at all without an understanding of redux-saga first

:+1: We should definitely add that to the documentation!

>  Given that, it's unfortunate that while containers can be generated without sagas, routes can not.

They can't? They should adapt to the container IIRC! That's a bug then, would you mind submitting an issue about that? Thanks!

learntoswim:
I found my way here due to realizing there's bug where sagas are being injected multiple times in a page when using routes (thus being triggered multiple times on an action).

I used @hansoksendahl's changes in `asyncInjectors.js`, and the bug is no longer apparent.

mxstbr:
@learntoswim this is the intended behaviour, this should be solved with the watcher pattern in a saga, see the example saga!

learntoswim:
Hmm. I can see that example working in that case, but in others it's a little more complicated. Right now, @hansoksendahl's fix is working perfectly. I had trouble trying to use the example pattern.

PR #747, opened 2016-07-18
add yaml loader and node exceptions to fix errors in dll building

jeffscottward:
The node.js runtime expections and file-type loaders from the webpack.base.config are not inherited to the webpack.dll.config so they need to be redeclared. (Or possibly seperated out?) Not sure what the solve is.

### DIFF
--- a/internals/webpack/webpack.dll.babel.js
+++ b/internals/webpack/webpack.dll.babel.js
@@ -28,6 +28,24 @@ module.exports = {
     path: outputPath,
     library: '[name]',
   },
+  module: {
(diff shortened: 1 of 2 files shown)


mxstbr:
Why do we need them for the DLL?

mxstbr:
This is no longer relevant I'm afraid, we're extending `webpack.dll.babel.js` from `webpack.base.babel.js`, so those are inherited now.

Sorry for letting this sit for so long, and thanks for taking the time to submit this PR!

PR #759, opened 2016-07-20
(docs) Remove dollar sign ($) in front of terminal commands

### LINKED ISSUE #757: Remove dollar sign ($) in front of terminal commands in docs
When writing the docs I prefixed all terminal commands with `$` like so:

```
$ npm run something
```

Due to a recent discussion on twitter I've come to realize that this is confusing to some people, so we should remove all of the `$`. The above example would thus be:

```Sh
npm run something
```

----

*If you're keen to submit a Pull Request for this, please comment underneath stating that you're looking into it so we don't duplicate work! All of our documentation is either in `README.md` or the `docs/` folder in the root.*

dmnkgrc:
I think this fixes #757 :)

Hope I did this right

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -58,11 +58,11 @@
 
 ### Development
 
-Run `$ npm start` to see your app at `localhost:3000`
+Run `npm start` to see your app at `localhost:3000`
 
 ### Building & Deploying
 
-1. Run `$ npm run build`, which will compile all the necessary files to the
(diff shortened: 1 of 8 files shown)


mxstbr:
Awesome work! Good catch on the Windows ones too, I would've missed those :grin:

Very well done, congratulations on landing your first PR! :tada: :tada: :tada:

PR #762, opened 2016-07-21
Add saga prompt & detection on route generator.

### LINKED ISSUE #761: Bug - generate route registers non-existent saga
ref [#744 (comment)](https://github.com/mxstbr/react-boilerplate/pull/744#issuecomment-233851411)

Generating a route to a container with a reducer, but without a saga registers a saga in the route. This results in a broken route/container that is not trivial to fix.

Repro steps:
- Create a new container `npm run generate container` Accept defaults _except_ saga.
image

- Create a route `npm run generate route`
image

- Start server `npm start`
image

- Edit ./app/routes.js to remove reference to saga
image
image

- Restart server (no error), open `http://localhost:3000/test` Get blank screen and console error
image

kraffslol:
This adds a prompt to the route generator asking you if you want to use sagas for the route you're generating. It automatically detects if the component has a sagas file and sets the default accordingly.

Fixes #761.

### DIFF
--- a/internals/generators/route/index.js
+++ b/internals/generators/route/index.js
@@ -47,12 +47,6 @@ module.exports = {
 
       return 'path is required';
     },
-  }, {
-    type: 'confirm',
-    name: 'useSagas',
-    default: true,
-    when: answers => sagasExists(answers.component),
-    message: 'Do you want sagas for asynchronous flows? (e.g. fetching data)',
   }],
 
   // Add the route to the routes.js file above the error route
@@ -60,6 +54,7 @@ module.exports = {
   actions: data => {
     const actions = [];
     if (reducerExists(data.component)) {
+      data.useSagas = sagasExists(data.component);
       actions.push({
         type: 'modify',
         path: '../../app/routes.js',
--- a/internals/generators/route/index.js
+++ b/internals/generators/route/index.js
@@ -1,10 +1,9 @@
 /**
  * Route Generator
  */
-
 const fs = require('fs');
 const componentExists = require('../utils/componentExists');
-const componentName = '';
+var componentName; // eslint-disable-line no-var
 
 function reducerExists(comp) {
   try {
@@ -53,8 +52,8 @@ module.exports = {
   }, {
     type: 'confirm',
     name: 'useSagas',
-    default: sagasExists(componentName) ? true : false,
-    message: 'Do you want to use sagas?',
+    default: sagasExists(componentName),
+    message: 'Do you want sagas for asynchronous flows? (e.g. fetching data)',
   }],
 
   // Add the route to the routes.js file above the error route
--- a/internals/generators/route/index.js
+++ b/internals/generators/route/index.js
@@ -4,10 +4,20 @@
 
 const fs = require('fs');
 const componentExists = require('../utils/componentExists');
+const componentName = '';
 
 function reducerExists(comp) {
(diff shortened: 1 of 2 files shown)


internals/generators/route/index.js
```diff
@@ -39,6 +50,11 @@ module.exports = {
 
       return 'path is required';
     },
+  }, {
+    type: 'confirm',
+    name: 'useSagas',
+    default: sagasExists(componentName) ? true : false,
+    message: 'Do you want to use sagas?',
```
gihrig: How about changing message to:
> Do you want sagas for asynchronous flows? (e.g. fetching data)

This is the sagas prompt from the container generator.

kraffslol: Sounds good. I'll change the message.

kraffslol:
The generator will now only prompt you if it can find a related sagas file to your component. If it doesn't it assumes that no saga is being used and adds a route without the saga lines.

mxstbr:
a) This looks great, thanks!

b) Not a fan of the prompt, if a saga exists add it to the route automatically. Why would we ask the user again?

gihrig:
Tested:
Container: no reducer, no saga -> no saga prompt -> app works 
Container: no reducer, has saga -> answer No to saga prompt -> app works 
Container: no reducer, has saga -> answer Yes to saga prompt -> app works 
Container: has reducer, no saga -> no saga prompt -> app works 
Container: has reducer, has saga -> answer No to saga prompt -> app works
Container: has reducer, has saga -> answer Yes to saga prompt -> app works

Works for me 😃 

Nice work @kraffslol, @mxstbr  recommend merge 💯

kraffslol:
@mxstbr I haven't found a way to pass a variable to the handlebars template without adding a prompt entry. There doesn't seem to be a way to pass a variable through the action directly.

The alternative I see to bypass the prompt is creating another template i.e. `routeWithReducerNoSagas.hbs` and doing the `sagasExists` call in the actions method, like reducerExists currently does. I having trouble deciding which solution is the best.

mxstbr:
I would strongly prefer not to have a prompt, since I cannot think of a case where the user wants to code split the route but not the saga. The saga will not load and the user will be heavily confused…

kraffslol:
Are you okay with the approach of adding another template then?

mxstbr:
Yeah, I guess that's fine if there's no other way… /cc @amwmedia who is the creator of plop, maybe he has some insights here!

amwmedia:
@kraffslol @mxstbr Can one of you boil down the question for me there? Something about passing variables to handlebars?

amwmedia:
I THINK the answer is that yes you can pass data into handlebars by just adding another property to the `data` object within your actions function. The `data` object that is passed into the actions function is the same, mutable, object that handlebars receives. So you can change the user's answers, or add data to it as needed.

kraffslol:
@amwmedia Thank you! It's just what we needed.

amwmedia:
@kraffslol glad I could help

mxstbr:
Awesome, thanks so much for doing this @kraffslol!! :tada: (and also @amwmedia for chiming in obviously!)

gihrig:
@kraffslol Tested all 4 combos of reducers and sagas(none, R, S, RS) works perfectly.

Awesome stuff, very helpful! ❤️

PR #771, opened 2016-07-23
Delete a duplicate dependency

lvarayut:
`minimist` has been defined twice, in both dependencies and devDependencies. The one in the dependencies got removed.

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,7 +133,6 @@
     "history": "3.0.0",
     "immutable": "3.8.1",
     "ip": "1.1.3",
-    "minimist": "1.2.0",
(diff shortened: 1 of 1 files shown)


mxstbr:
We actually use it on the `server`, so I think we should delete the one in the `devDeps`? Very good catch!

lvarayut:
Is it already correct? We're using it on the server. That's why I deleted from the dependency and kept the one in the `devDependencies`.

mxstbr:
Ah yeah, I guess that makes sense, my bad! Thanks so much @lvarayut!

PR #579, opened 2016-06-20
Istanbul

FezVrasta:
This PR replaces the isparta coverage reported (not maintained anymore) with Istanbul instrumenter.

Please carefully check that everything still works as expected, I've not a lot of experience with coverage reports.

Closes #274

PR #782, opened 2016-07-25
Switch from isparta to babel-plugin-istanbul for coverage

kane-c:
This follows up #579 by replacing isparta with babel-plugin-istanbul. The advantage of this is that it reports coverage on the original pre-transpiled ES6 code *without* invisible `ignore` statements. The reports are fully accurate now which is awesome.

### DIFF
--- a/internals/testing/test-bundler.js
+++ b/internals/testing/test-bundler.js
@@ -8,6 +8,6 @@ import chaiEnzyme from 'chai-enzyme';
 chai.use(chaiEnzyme());
 
 // Include all .js files under `app`, except app.js, reducers.js, routes.js and
-// store.js. This is for isparta code coverage
+// store.js. This is for istanbul code coverage
 const context = require.context('../../app', true, /^^((?!(app|reducers|routes|store)).)*\.js$/);
 context.keys().forEach(context);
--- a/internals/webpack/webpack.test.babel.js
+++ b/internals/webpack/webpack.test.babel.js
@@ -2,7 +2,6 @@
  * TEST WEBPACK CONFIGURATION
  */
 
-const path = require('path');
 const webpack = require('webpack');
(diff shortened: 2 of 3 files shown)


mxstbr:
CI fails because:

```
/home/travis/build/mxstbr/react-boilerplate/internals/webpack/webpack.test.babel.js
  5:7  error  'path' is defined but never used  no-unused-vars
```

kane-c:
D'oh, should've spotted that. Thanks.

mxstbr:
> Coverage decreased (-6.5%) to 91.781%

Any idea why that could be? What files are added to coverage that weren't there before/are we maybe not ignoring certain things we didn't before?

kane-c:
It looks like isparta counts more things as valid lines that the istanbul plugin doesn't, such as imports at the start of a file.

isparta doesn't seem to look quite as accurate for saga coverage either.

**Comparison**
Home page
[isparta](https://coveralls.io/builds/7140808/source?filename=app%2Fcontainers%2FHomePage%2Findex.js) vs [istanbul plugin](https://coveralls.io/builds/7147518/source?filename=app%2Fcontainers%2FHomePage%2Findex.js)
Notice how imports count in the former.

Home page sagas
[isparta](https://coveralls.io/builds/7140808/source?filename=app%2Fcontainers%2FHomePage%2Fsagas.js) vs [istanbul plugin](https://coveralls.io/builds/7147518/source?filename=app%2Fcontainers%2FHomePage%2Fsagas.js)
Notice isparta doesn't seem to pick up all the `yield`s

mxstbr:
Absolutely, this is awesome @kane-c, thanks so much! I'll take this and merge it with some test updates :+1:

mxstbr:
Continued in #795!

PR #788, opened 2016-07-26
Constrain eslint import/no-unresolved rule in app.js

lvarayut:

- [x]  Internal code generators and templates are updated (if necessary)

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -8,10 +8,12 @@
 // Needed for redux-saga es6 generator support
 import 'babel-polyfill';
 
+/* eslint-disable import/no-unresolved */
 // Load the favicon, the manifest.json file and the .htaccess file
 import 'file?name=[name].[ext]!./favicon.ico';
 import '!file?name=[name].[ext]!./manifest.json';
 import 'file?name=[name].[ext]!./.htaccess';
+/* eslint-enable import/no-unresolved */
 
 // Import all the third party stuff
 import React from 'react';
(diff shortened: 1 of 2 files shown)


app/app.js
```diff
@@ -25,7 +27,7 @@ import configureStore from './store';
 
 // Observe loading of Open Sans (to remove open sans, remove the <link> tag in
 // the index.html file and this observer)
-import styles from 'containers/App/styles.css';
+import styles from './containers/App/styles.css';
```
mxstbr: These `'./'` aren't necessary, webpack takes care of correcting them I think?

lvarayut: I'm don't think that it's because of webpack. The reason it's working would be `app.js` placed in the some folder as `containers`. That's why the `./` could be omitted. However, Eslint will warn that the path can't be resolved. [On line 26th](https://github.com/lvarayut/react-boilerplate/blob/3b19ef54dc5d2db42ae2449aa4bc536b233abc39/app/app.js#L26), we could do `import configureStore from 'store';` as well, we just need to be consistent.

mxstbr: > I don't think that it's because of webpack.

That's definitely because of webpack, we use the `resolve` option to specify which directories weback should search in. That's why in your components you can do

```JS
// components/Something/index.js

import SomethingElse from 'containers/SomethingElse';
```

and that'll work. (notice the non-relative path) I think the problem here is in our eslint-import-plugin setup, though I think we had that working before…

lvarayut: You're absolutely right. I just carefully read through the webpack and see the `resolve` option in the `webpack.base.babel.js`. Will modify the commit soon :)

mxstbr:
Thanks!

Issue #779, opened 2016-07-25
AppCache didn't cache other pages beside homepage (/)


chaintan17:
We do cache all the webpack file using AppCache
However, look like it currently only cache for `/` page but not the other.
I'm guessing it may be because of the maifest file is only set for `/` page
(http://www.html5rocks.com/en/tutorials/appcache/beginner/#toc-manifest-file-creating)

I still not sure how to fix it...

**Step to reproduce**
1. Checkout the project 
2. `npm run setup`
2. `npm run start:production`
3. open `http://localhost:3000`
4. click feature page
5. set throttling to offline
6. refresh the page
7. Chrome browser cannot reach the /feature page

(EDIT 1 :Update reproduce steps)

mxstbr:
I think this might be due to your webserver configuration, but I can't say for sure. @NekR is there anything the offline-plugin can do to fix this?

gihrig:
> Step to reproduce
> 1. Checkout the project 
> 2. npm run start:prod
> 3. click feature page
> 4. set throttling to offline
> 5. refresh the page
> 6. Chrome browser cannot reach the /feature page

Hmm... Some steps missing (perhaps you just forgot to list?)
1.5 npm run setup
2. should be `npm run start:production` (`start:prod` needs a previous build, it only starts the server).

I have found that off-line mode only works when the page is called through the router, not the address line.

Do the same as ss you did before, except:

At step 5, open localhost:3000
At step 6, click `Features` button at bottom of home page.

I'm not saying this is correct behavior, but that seems to be the way it works.

NekR:
@mxstbr were you able to reproduce the bug? Is all routing (e.g. `/feature`) generated by webpack as its output (e.g. by `HtmlWebpackPlugin`)?

`offline-plugin` doesn't add manifest to your `index.html` or any other page, instead it load specially generated iframe which initializes AppCache. All pages which should be cached are explicitly defined in AppCache Manifest. `offline-plugin` takes the information about which files to cache from webpack's `compilation.assets`,  basically it's a list of your webpack build output files. 

So if `/feature` isn't loading, then it's simply not listed in AppCache Manifest (and I'm sure it won't work for SW too then).

Can anyone provide output of `appcache/appcache.manifest` content and `sw.js` JSON header part here?

chaintan17:
Hi @NekR , thanks for your help.

Below is the content of `manifest.appcache`

```
CACHE MANIFEST
[...]

Issue #450, opened 2016-06-02
heroku causes infinite loop with appcache file


mlennie:
A `appcache/appcach` path is being added in an infinite loop in heroku server logs
Please see issue here: http://stackoverflow.com/questions/37581134/heroku-react-appcache-file-causes-infinite-loop

mlennie:
Ok I found out it only cause the infinite loop with `http` and not with `https`. Weird?

mlennie:
Ok it happens on safari if it's http or https

ThomasMarnet:
Having the same issue on Safari (http or https), everything is fine on others browsers.

mxstbr:
Hmm, I've never had that issue before and I'm running on a 5$ DO Droplet. Any ideas what could be causing this?

ThomasMarnet:
After some digging, I found out that it only occurs when the location's path has at least 2 levels (such as `/post/:id` or `/post/new` for example).

The (weird ?) thing is that it doesn't occurs when I navigate to such a route, but only when I reload the page being already on this route.

I hope you understand what i'm trying to say, if not I'll try to explain it with more specifics.

mxstbr:
Hm, weird. What does you route configuration look like?

ThomasMarnet:
Something like that mostly:

```javascript
{
      path: '/',
      
      getComponent(nextState, cb) {...},

      indexRoute: {
        getComponent(nextState, cb) {...},
      },
      
      childRoutes: [
        {
          path: 'work',
          getComponent(nextState, cb) {...},
        }, {
          path: 'work/:slug',
          getComponent(nextState, cb) {...},
        },
      ],
    }
```

ThomasMarnet:
Something like that mostly:

```javascript
{
	path: '/',
     
	getComponent(nextState, cb) {...},

	indexRoute: {
		getComponent(nextState, cb) {...},
	},
  
	childRoutes: [{
		path: 'work',
		getComponent(nextState, cb) {...},
    }, {
		path: 'work/:slug',
		getComponent(nextState, cb) {...},
	}]
}
```

mxstbr:
I honestly have no idea how to fix this, and I also don't much experience with Heroku, so I won't really be of help here. Has anybody fixed this issue yet?

jupitercow:
I have this same issue on my local when I try to get the boilerplate working in a subdirectory. I changed the `publicPath` in webpack base and that got most stuff working. Except for webfonts and appcache.

[...]

PR #791, opened 2016-07-27
Disable AppCache generation

sethmcleod:
Disabling based on #450 and the fact that it’s depreciated: _https://developer.mozilla.org/en-US/docs/Web/HTML/Using_the_application_cache_

Also this would technically solve #779

### DIFF
--- a/internals/webpack/webpack.prod.babel.js
+++ b/internals/webpack/webpack.prod.babel.js
@@ -104,11 +104,7 @@ module.exports = require('./webpack.base.babel')({
       // Removes warning for about `additional` section usage
       safeToUseOptionalCaches: true,
 
-      AppCache: {
-        // Starting from offline-plugin:v3, AppCache by default caches only
-        // `main` section. This lets it use `additional` section too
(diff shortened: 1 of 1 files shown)


mxstbr:
Thanks @sethmcleod, much appreciated!

gihrig:
> Disabling [Appcache] completely is also completely fine.

[#779 (comment)](https://github.com/mxstbr/react-boilerplate/issues/779#issuecomment-235541295)

Would disabling AppCache cause problems in IE and Edge?

If I understand correctly `Service Workers` replace `AppCache`, but service worker support does not exist in IE or Edge.

See: [Service Workers](http://caniuse.com/#search=servicework) and [AppCache](http://caniuse.com/#search=appcache)

Given the above, it seems reasonable to use AppCache as a fallback for the foreseeable future as described in [#779 (comment)](https://github.com/mxstbr/react-boilerplate/issues/779#issuecomment-235541295) Rather than disablining it all together.

mxstbr:
You might be right...

PR #793, opened 2016-07-27
Fix issue #791

chaintan17:
In Safari browser,
when there is component using `<FormattedNumber />`
It will not get formatted and have some error in console.log

fixed it follow to this document in intl
https://github.com/andyearnshaw/Intl.js#intljs-and-browserifywebpack

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -106,6 +106,7 @@ if (!window.Intl) {
   Promise.all([
     System.import('intl'),
     System.import('intl/locale-data/jsonp/en.js'),
+    System.import('intl/locale-data/jsonp/de.js'),
   ]).then(() => render(translationMessages));
 } else {
   render(translationMessages);
--- a/app/app.js
+++ b/app/app.js
@@ -103,7 +103,10 @@ if (module.hot) {
 
 // Chunked polyfill for browsers without Intl support
 if (!window.Intl) {
-  System.import('intl').then(() => render(translationMessages));
+  Promise.all([
+    System.import('intl'),
+    System.import('intl/locale-data/jsonp/en.js'),
+  ]).then(() => render(translationMessages));
 } else {
   render(translationMessages);
 }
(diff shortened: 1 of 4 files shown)


app/app.js
```diff
@@ -103,7 +103,10 @@ if (module.hot) {
 
 // Chunked polyfill for browsers without Intl support
 if (!window.Intl) {
-  System.import('intl').then(() => render(translationMessages));
+  Promise.all([
+    System.import('intl'),
+    System.import('intl/locale-data/jsonp/en.js'),
```
mxstbr: Why do we import the `en` local-data here? Do we need to do the same thing for all languages?

chaintan17: It should but only when the website support other language.
So i guess i should add new line when user run `npm run generate` new langauge then.

mxstbr: Exactly, and add it to the documentation in `i18n.md`!

chaintan17: @mxstbr please review.
I don't see what should be added in the docs, i think it's okay to leave it like that.

mxstbr:
You're right, this LGTM code wise! @gihrig would you mind trying this out?

gihrig:
Sure!

Super busy with work ATM in about 12 hrs maybe...

mxstbr:
This works perfectly, thanks so much @chaintan17!

gihrig:
> It will not get formatted and have some error in console.log

@chaintan17 I am unable to reproduce any problem or console error with current master (loading GitHub repo list).

Have you observed problems with Safari, or are you just following recommendation on linked page?

If you see problems, steps to reproduce please.

internals/templates/app.js
```diff
@@ -83,7 +83,10 @@ if (module.hot) {
 
 // Chunked polyfill for browsers without Intl support
 if (!window.Intl) {
-  System.import('intl').then(() => render(translationMessages));
+  Promise.all([
+    System.import('intl'),
+    System.import('intl/locale-data/jsonp/en.js'),
```
chaintan17: I'm not sure but I think there is race condition in here.
in case, `System.import('intl/locale-data/jsonp/en.js'),` is run before `System.import('intl'),` it will cause this bug in Safari

image

PR #796, opened 2016-07-28
fix route usage example

jk2K:
update doc


mxstbr:
Awesome, good catch! Thanks @jk2K!

PR #806, opened 2016-07-31
fix(webpack): webkit prefix get disappear because of uglifyjs

chaintan17:
When i set `cssnext` to support browser `Safari >= 8`
it expect to prefix `webkit` for some styles.

It's already work properly in dev environment
but not in production environment.

After reading this issue
https://github.com/postcss/autoprefixer/issues/660

it suppose to add `&-autoprefixer` in `css-loader` parameter

So, i add a pull request here.


mxstbr:
Ah yes, good catch, thanks!

PR #811, opened 2016-08-01
feat(core): React.PureComponent demo/generators

justingreenberg:
Since data structures are immutable, use `Purecomponent` released in [15.3.0](https://github.com/facebook/react/releases/tag/v15.3.0). Under the hood, it just runs a [shallow comparison on props](https://github.com/facebook/react/pull/7195/files#diff-6363300a479c9057729248ca9dd3e7b6R874) before tagging the component for update. This also updates generators. although we use memoized selectors for view data, i would still consider this as a good foundational feature to leverage in pure class components and generators...

> Therefore, we're changing functional components to behave as pure when they're used inside a React.PureComponent but to rerender unconditionally when contained in a React.Component
> 
> We anticipate that this behavior will work well in real-world apps: if you use immutable data, your class-based components can extend React.PureComponent and your functional components will be pure too; if you use mutable data, your class-based components will extend React.Component and your functional components will update accordingly.

https://github.com/facebook/react/pull/6914


justingreenberg:
@mxstbr thoughts on this?

gihrig:
@justingreenberg I read over [React.PureComponent](https://github.com/facebook/react/pull/6914) Looks like logical progresson for RBP.

Since it's (mostly) a drop-in replacement, I'm inclined to say _let's do it_. Since `PureComponent` is (still) officially undocumented, this PR should probably reference something in code comments like React [change log](https://github.com/facebook/react/blob/092f5ae867721bdc386120a14f789d0a64d31916/CHANGELOG.md#react)

I'm short on time just now, so won't have time to test for a few days...

Seems like an excellent candidate for @mxstbr suggested `development` branch 👍

justingreenberg:
@gihrig yes, and of course it's really just a demo (in the default app)... the real meat of this PR is the option to create a `PureComponent` when using internal generators

mxstbr:
How does this work together with ImmutableJS? @justingreenberg I'm not sure we really want people to do this since we do use immutable data structures for cheap `nextProps.equals(this.props)`…

mxstbr:
> @mxstbr i don't see any immutable comparisons in the current codebase?

I had no idea what I was doing back when I wrote the current example. 🙈 We really should be doing immutable `shouldComponentUpdate` checks everywhere, but I haven't really updated the example much because I've been thinking about writing a new, slightly more complex one from scratch so people can really get a grasp on all the stuff we use.

Stay tuned for updates on that!

> maybe i'm misunderstanding, but i think this means that using legacy Components will cause our stateless functional child components (Button, H1, etc) to rerender on every diff...

Yeah yeah alright alright I'm convinced.

> also, can you trigger new appveyor build whenever you have a sec 🙇 😃

On it!

mxstbr:
Let's not merge this for now because `React.PureComponent` isn't documented yet so it might confuse some people. Ref https://github.com/facebook/react/issues/7379

justingreenberg:
@mxstbr sounds good... i've had some extra time this week (in case you didn't notice 😉) so i'm thinking maybe i'll take a stab at writing the docs later

justingreenberg:
> I had no idea what I was doing back when I wrote the current example. 🙈 We really should be doing immutable shouldComponentUpdate checks everywhere

extending from `PureComponent` will automatically take care of this :)

> I haven't really updated the example much because I've been thinking about writing a new, slightly more complex one from scratch so people can really get a grasp on all the stuff we use.

let me know what you are thinking here... i def had some thoughts on this as well

mxstbr:
I really want to get some authentication similar to what I did with [`login-flow`](https://github.com/mxstbr/login-flow). Showcasing that and the general workflow with sagas could really help people out!

What I'm not 100% on yet is what happens once you're logged in, what were your ideas?

justingreenberg:
nice! yes authentication comes up a lot 👍

i was thinking about the saga utilities, i wrote a quick POC on my `feature/saga-utilities` branch:
https://github.com/mxstbr/react-boilerplate/compare/master...justingreenberg:feature/saga-utilities

as you know many people have been confused by the current implementation, so i've been thinking about how we can simplify the API a bit... this is one idea, provides a more familiar api for binding sagas directly to component (instead of route definition) ala react-redux.. provides two  methods `run` and `watch`... run just maps over sagas, watch maintains references and cancels when component unmounts

mxstbr:
That looks awesome. 
Why do you use `store.runSaga` in `app.js`? Couldn't you

```JS
import sagaMiddleware from 'redux-saga';
/* … */
<SagaProvider run={sagaMiddleware.run} />
```
No need to attach something to the store, right? If that's the case as I think it is, you could even assume that as a default and only let it be overriden by passing in a custom `run`?

smashercosmo:
react's PureComponent has issues with hot reloading https://github.com/facebook/react/issues/7388

justingreenberg:
The issue seems related to `react-transform` (deprecated for a while) and react native (not a render target for this project)... Have you tried checking out this pr locally?

Either way, we need to remove https://github.com/danmartinez101/babel-preset-react-hmre anyway, and replace it with https://github.com/gaearon/react-hot-loader v3 or vanilla webpack hmr @mxstbr lmk your thoughts 

As a side note, "break hmr" usually just means components hot reload but do not maintain local state, which is generally OK for developing redux applications imo

mxstbr:
> As a side note, "break hmr" usually just means components hot reload but do not maintain local state, which is generally OK for developing redux applications imo

That's absolutely fine.

> replace it with https://github.com/gaearon/react-hot-loader v3 or vanilla webpack hmr

Is that stable yet? AFAIK it isn't… (not that we've ever cared about that much, but you know…)

justingreenberg:
> (not that we've ever cared about that much, but you know…)

agreed lol—we have been using webpack v2 alpha (many feature not documented) without any known issues for months right?? living on the edge dude :metal: 

> > As a side note, "break hmr" usually just means components hot reload but do not maintain local state, which is generally OK for developing redux applications imo

> That's absolutely fine.

if you're cool with that, my recommendation would be to pull out `babel-hmre` and deprecated `react-transform` and just use webpack's vanilla hmr system

gihrig:
Huston, we have a problem...

This build (and others) are failing due to an eslint (and friends) problem described in #864. This problem is resolved by #862. But #862 is stalled due to Coveralls not completing...

justingreenberg:
@gihrig thank you for the explanation—this was driving me nuts :)

justingreenberg:
@mxstbr is netlify building off `dev`?

mxstbr:
Yeah, check the last commit, it should have a deploy!

justingreenberg:
monitoring https://github.com/facebook/react/pull/7730

justingreenberg:
https://facebook.github.io/react/docs/react-api.html#react.purecomponent 🎉

sedubois:
Is there a way to add linting, to enforce the use of PureComponent?

justingreenberg:
@mxstbr this should be ready as well needs to be rebased but in meantime if you have a few mins to take the generators for a spin and let me know if there are any general changes will try to handle today :)

mxstbr:
> Is there a way to add linting, to enforce the use of PureComponent?

I don't think we want to do that, as PureComponents can have unexpected effects in terms of rerenders. If people use them, they should make a choice to do so.

justingreenberg:
@sedubois sorrry meant to respond!

> Is there a way to add linting, to enforce the use of PureComponent?

As max pointed out we probably don't want to force shallow comparisons

justingreenberg:
@mxstbr rebased with `dev` and CI is passing 👍 
@gihrig can you guys confirm generator support is working properly?

mxstbr:
Generator support is working _perfectly_. Thank you so much, glad we can finally land this!

PR #812, opened 2016-08-02
feat(utils): Start hardening internal utilities

justingreenberg:
Starts adding some basic type checking to supplement our tests for internal utilities. This also adds `invariant` dependency, but since it's already in our graph and optimized for production shouldn't affect build size. I'm starting to think internals should probably be written kinda like external libraries, since many users seem to really use these apis.

### DIFF
--- a/app/utils/asyncInjectors.js
+++ b/app/utils/asyncInjectors.js
@@ -1,20 +1,23 @@
+import { conformsTo, isEmpty, isFunction, isObject, isString } from 'lodash';
 import invariant from 'invariant';
+import warning from 'warning';
 import createReducer from '../reducers';
-
-export const isType = type => target => typeof target === type;
-export const isFunction = isType('function');
 
 /**
  * Validate the shape of redux store
  */
-const reduxKeys = ['dispatch', 'subscribe', 'getState', 'replaceReducer'];
-const internalKeys = ['runSaga']; // asyncReducers does not exist yet
-const expectedKeys = [...reduxKeys, ...internalKeys];
-
 export function checkStore(store) {
+  const shape = {
+    dispatch: isFunction,
+    subscribe: isFunction,
+    getState: isFunction,
+    replaceReducer: isFunction,
+    runSaga: isFunction,
+    asyncReducers: isObject,
+  };
   invariant(
-    expectedKeys.reduce((acc, curr) => (acc ? isFunction(store[`${curr}`]) : false), true),
-    '(app/utils...) asyncInjectors: Received invalid redux store.'
+    conformsTo(store, shape),
+    '(app/utils...) asyncInjectors: Expected a valid redux store'
   );
 }
 
@@ -26,8 +29,8 @@ export function injectAsyncReducer(store, isValid) {
     if (!isValid) checkStore(store);
 
     invariant(
-      isType('string')(name) && name.length && isFunction(asyncReducer),
-      '(app/utils...) injectAsyncReducer: `asyncReducer` must be a reducer function'
+      isString(name) && !isEmpty(name) && isFunction(asyncReducer),
+      '(app/utils...) injectAsyncReducer: Expected `asyncReducer` to be a reducer function'
     );
 
     store.asyncReducers[name] = asyncReducer; // eslint-disable-line no-param-reassign
@@ -43,8 +46,13 @@ export function injectAsyncSagas(store, isValid) {
     if (!isValid) checkStore(store);
 
     invariant(
-      Array.isArray(sagas) && sagas.length,
-      '(app/utils/injectAsyncSagas...) `sagas` must be an array of generator functions'
+      Array.isArray(sagas),
+      '(app/utils...) injectAsyncSagas: Expected `sagas` to be an array of generator functions'
+    );
+
+    warning(
+      !isEmpty(sagas),
+      '(app/utils...) injectAsyncSagas: Received an empty `sagas` array'
     );
 
     sagas.map(store.runSaga);
--- a/app/utils/asyncInjectors.js
+++ b/app/utils/asyncInjectors.js
@@ -1,10 +1,35 @@
+import invariant from 'invariant';
 import createReducer from '../reducers';
+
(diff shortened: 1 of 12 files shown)


app/utils/asyncInjectors.js
```diff
@@ -1,10 +1,35 @@
+import invariant from 'invariant';
 import createReducer from '../reducers';
 
+export const isType = type => target => typeof target === type;
+export const isFunction = isType('function');
```
mxstbr: Should we add tests for these?

mxstbr:
Love it @justingreenberg, great start!

I'm not sure we want to throw there, these would be runtime errors that happen when you visit a page. Blocking execution there seems kinda wrong, shouldn't we just warn or am I missing something?

justingreenberg:
@mxstbr so we may want to throw in specific critical paths, where passing an invariant is certain to cause runtime errors downstream... for example, passing an invalid store (ie. missing `dispatch` or our custom `runSaga` method) or passing a generator function directly to `injectSagas`

invariant throws so that developers have access to stack for determining callsite

if you want, we can remove checks in production eg using https://github.com/yahoo/strip-loader or by wrapping checks in `__DEV__` and delegating to webpack's dead code elimination

alternatively, if we simply wanted to continue execution in these cases, we could replace `invariant` with https://github.com/BerkeleyTrue/warning (also extracted from `fbjs`)

side note: it may not be a bad idea to include `warning` for [logging errors](https://github.com/mxstbr/react-boilerplate/blob/master/app/routes.js#L8) either way

app/utils/asyncInjectors.js
justingreenberg: i actually checked last night—lodash isn't explicitly included in our dependencies, but it is in our devDeps... I can move into dependencies and use destructured imports (for tree shaking) if you are ok with it?

mxstbr:
I'm convinced, I think the way you did it is fine, you're right. Should we merge this or do you want to get any more changes in?

justingreenberg:
awesome! 👍 it should be good to go, i just wanted to add a few `instanceof` checks... will push up in a the next few minutes—did you want me to move `lodash` to deps for type checking eg `isFunction`?

mxstbr:
Yeah I think that's reasonable since I'm about 99% sure one of our deps already uses lodash. (would be very surprised if that wasn't the case)

gihrig:
Re lodash, just checked, it's there 💯

package.json
```diff
@@ -207,7 +210,6 @@
     "html-webpack-plugin": "2.22.0",
     "image-webpack-loader": "2.0.0",
     "imports-loader": "0.6.5",
-    "ip": "1.1.3",
```
justingreenberg: duplicate, used in server logger

justingreenberg:
@mxstbr @gihrig please review, should be good to go 👍

gihrig:
Cloned to new dir and pulled branch
`npm run setup` ✓
`npm start` ✓
`npm run start:tunnel` ✓
`npm run start:production` ✓
 - Offline ✓
 - Change local to de ✓

`npm run clean` ✓
`npm run generate container` (Test - accept defaults) ✓
`npm run generate route` (Test -> /test) ✓
`npm start` -> localhost:3000/test (Get Blank Screen & console error) ✘
`routes.js?6db9:49 Uncaught (in promise) ReferenceError: injectReducer is not defined(…)`

After above steps, `app/routes.js:15`
``` javascript
export default function createRoutes() {
  // Create reusable async injectors using getAsyncInjectors factory
  // const { injectReducer, injectSagas } = getAsyncInjectors(store);
```

Removing comments fails because `store` is not defined. 

If not related to this PR, What is the reasoning behind this choice?

Mac OS 10.11.6
Chrome 52.0.2743.82

justingreenberg:
@gihrig I'm not sure about the generator behavior, but this issue seems to be related to internal template for routes:
https://github.com/mxstbr/react-boilerplate/blob/master/internals/templates/routes.js

I ran the same steps and got the same result, but we just have to add `store` param to `createRoutes()` and uncomment the relevant `getAsyncInjectors` import and function call...

```js
import { getAsyncInjectors } from 'utils/asyncInjectors';

  // Create reusable async injectors using getAsyncInjectors factory
  const { injectReducer, injectSagas } = getAsyncInjectors(store);
```

and everything works :) if you make adjustments outlined above to `templates/routes.js`, route generator script works as expected

gihrig:
@justingreenberg Sure enough, works as expected with those mods.

@mxstbr I see no reason not to merge.

@mxstbr Do you know why the store was commented in templates/routes ?

mxstbr:
Yeah, that's because if you don't have actions/reducers/… the store won't be used in the route and eslint will throw an "Unused variable" error… :confused:

justingreenberg:
@gihrig @mxstbr that makes sense, there's only so much that can be automated within reason. was there anything else you wanted me to add to this pull?

gihrig:
> Yeah, that's because if you don't have actions/reducers/… the store won't be used in the route and eslint will throw an "Unused variable" error… 😕

@mxstbr That's understandable, but having generator configured code that crashes feels like a bug to me. If automating 100% workable code generation is not practical, IMO, at the very least the code should work as generated and code comments should describe what's going on.

If you agree, I'll open an issue on it, can't promise that I have time to fix it any time soon though 😞 

As for this PR, I see no reason not to merge.

mxstbr:
Yeah please do open an issue! 

On another note, we should have a `development` branch that all the PRs are submitted against instead of merging everything into master, and then every two weeks or something merging that into master and doing an "official release" – what do you think?

justingreenberg:
@mxstbr @gihrig i think this is a matter of balancing opinions and setting constraints. since redux-saga is an integral part of this project, maybe it should not be optional when creating a new route. 

either way, imo `routes.js` template should support these out of the box by making the changes outlined above. if you guys agree i will open PR

mxstbr:
:+1:

gihrig:
@mxstbr development branch - Yes, absolutely 👍 

@justingreenberg PR 👍

PR #813, opened 2016-08-02
feat(intl): Use `appLocales` to populate i18n toggle

justingreenberg:
use `i18n.js#appLocales` as source of locales, remove hard coded array

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -105,11 +105,17 @@ if (module.hot) {
 
 // Chunked polyfill for browsers without Intl support
 if (!window.Intl) {
-  Promise.all([
-    System.import('intl'),
-    System.import('intl/locale-data/jsonp/en.js'),
-    System.import('intl/locale-data/jsonp/de.js'),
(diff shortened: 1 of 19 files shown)


mxstbr:
Have you tried adding a new language? I'm 99% sure this will throw because there is no text in `messages.js` for the new language.

mxstbr:
Just tried this locally, this throws because we need add a message with the key to the `LocalToggle/messages.js`!

justingreenberg:
@mxstbr that makes sense... i will take a few mins tomorrow (later today for you) to see if i can find a workaround, if not will close... hard coding the locales in numerous places just feels wrong

justingreenberg:
we could populate messages.js in generator

justingreenberg:
@mxstbr threw up a super simple solution that works out of the box as you generate more languages... not priority by any means, feel free to merge or close :)

mxstbr:
Hah I love it! Great work @justingreenberg! 🔥

PR #814, opened 2016-08-02
fix(containers): Remove unnecessary param from LanguageProvider

justingreenberg:
removes unnecessary `mapDispatchToProps` from `LanguageProvider` container


mxstbr:
:+1:

PR #815, opened 2016-08-02
chore(core): Use React.Children to manage opaque prop data

justingreenberg:
Use [React.Children](https://facebook.github.io/react/docs/top-level-api.html#react.children) to manage opaque data structures passed to `props.children`


mxstbr:
This seems useful, thanks @justingreenberg!

PR #551, opened 2016-06-14
Change resolve.packageMains ==> resolve.mainFields

mgmcdermott:
Noticed this in my own project and thought I'd submit a PR here. If you look in the source code of webpack, https://github.com/webpack/webpack/blob/6b0c20a53ad7b04d4282a8a5c334ea0982fd364c/lib/WebpackOptionsDefaulter.js#L82, you'll see that the actual name of packageMains is mainFields. Easiest way to see the implementation is by checking out the source code: https://github.com/webpack/webpack/search?utf8=%E2%9C%93&q=mainFields

PR #831, opened 2016-08-05
FIX  #551  - Enable webpack resolve to work with jsnext:main within npm modules th…

andykenward:
Should close issue #551.


mxstbr:
Thanks @andykenward!

PR #838, opened 2016-08-05
Feature/remove babel-preset-es2015-webpack

andykenward:
Since the latest release of `babel-core` and `babel-preset-2015` (6.13.2) can handle presets options.
No longer need to use `babel-preset-es2015-webpack`

Thanks goes to @zoontek on https://github.com/gajus/babel-preset-es2015-webpack/issues/14#issuecomment-237869435

### DIFF
--- a/internals/webpack/webpack.base.babel.js
+++ b/internals/webpack/webpack.base.babel.js
@@ -75,7 +75,7 @@ module.exports = (options) => ({
       '.jsx',
       '.react.js',
     ],
-    packageMains: [
+    mainFields: [
       'jsnext:main',
       'main',
     ],
--- a/package.json
+++ b/package.json
@@ -55,20 +55,13 @@
   "pre-commit": "lint:staged",
   "babel": {
     "presets": [
-      [
-        "es2015",
-        {
-          "modules": false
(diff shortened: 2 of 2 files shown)


mxstbr:
This is great! 🎉 Can we undo the other whitespace changes?

andykenward:
@mxstbr whitespace done by npm automatically when the package were updated via command line. So will come back eventually even if they are removed

mxstbr:
Awesome, thanks!

garajo:
Thank you very much for these changes and this thread. I'm trying to implement these changes on an existing react-boilerplate project, but I get the following error after running `npm run postinstall`. 

```
ERROR in ./~/diffie-hellman/lib/primes.json
Module parse failed: /media/gajo/ssd/home/gajo/local/proyct/bean.cab/broadbean/node_modules/diffie-hellman/lib/primes.json Unexpected token (2:11)
You may need an appropriate loader to handle this file type.

```
I've looked through package.json and compared my webpack configurations with the latest version of master, but I don't see any difference that would cause this loader failure. Could anyone give a pointer on how I could fix this?

PR #839, opened 2016-08-05
chore(templates): include curried injectors after clean script

justingreenberg:
@gihrig @mxstbr updates default route template to include injectors curried with instantiated store

ref: https://github.com/mxstbr/react-boilerplate/pull/812#issuecomment-237315789


mxstbr:
Ahh, nice one @justingreenberg, thanks!

PR #856, opened 2016-08-11
Update lint-staged config to v2.x format.

raizyr:
The version of lint-staged was changed from v1.x to v2.x in
commit 052b15648f758fe1d7835342d66404ee92ae2426.  The config format
for v2.x has keys and values flipped.


mxstbr:
Oooh, thanks so much!

Issue #631, opened 2016-07-02
store.test.js is duplicated


akaRem:
subj test suite is located in 2 places with just import difference
```
$ wc -l app/store.test.js
      27 app/store.test.js
$ diff app/store.test.js app/tests/store.test.js
6c6
< import configureStore from './store';
---
> import configureStore from '../store';
```

mxstbr:
Is that after cleaning or in the example app?

(PR? :grin:)

akaRem:
@mxstbr Noticed after clean command.. And looks like it didn't exit before that

mxstbr:
Fixed by the amazing @uzarubin in #858!

Issue #458, opened 2016-06-03
tests/store.test.js is not executed


FezVrasta:
When I run `npm test` I get:

```Shell
START:
ERROR: 'Warning: Failed propType: Required prop `href` was not specified in `A`.'
  <A />
    ✔ should render its children
ERROR: 'Warning: Failed propType: Required prop `children` was not specified in `A`.'
    ✔ should adopt the className
    ✔ should adopt the href
    ✔ should adopt the target
  <Button />
    ✔ should render its children
ERROR: 'Warning: Failed propType: Required prop `children` was not specified in `Button`.'
    ✔ should adopt the className
    ✔ should render an <a> tag if no route is specified
    ✔ should render a button to change route if the handleRoute prop is specified
    ✔ should handle click events
  <Footer />
    ✔ should render the copyright notice
    ✔ should render the credits
  <H1 />
    ✔ should render its text
  <H2 />
    ✔ should render its text
  <H3 />
    ✔ should render its text
  <Img />
    ✔ should render an <img> tag
    ✔ should have an alt attribute
  <IssueIcon />
    ✔ should render a SVG
  <List />
    ✔ should render the component if no items are passed
    ✔ should render the items
  <ListItem />
    ✔ should adopt the className
    ✔ should render the content passed to it
  <LoadingIndicator />
    ✔ should render 14 divs
  App Actions
    loadRepos
      ✔ should return the correct type
    reposLoaded
      ✔ should return the correct type and the passed repos
    repoLoadingError
      ✔ should return the correct type and the error
ERROR: 'Warning: Failed propType: Invalid prop `src` of type `object` supplied to `Img`, expected `string`.'
  <App />
    ✔ should render the logo
    ✔ should render its children
    ✔ should render the footer
  appReducer
    ✔ should return the initial state
    ✔ should handle the loadRepos action correctly
    ✔ should handle the reposLoaded action correctly
    ✔ should handle the repoLoadingError action correctly
  selectGlobal
    ✔ should select the global state
  selectCurrentUser
    ✔ should select the current user
  selectLoading
    ✔ should select the loading
  selectError
    ✔ should select the error
  selectRepos
    ✔ should select the repos
  selectLocationState
    ✔ should select the route as a plain JS object
  <FeaturePage />
    ✔ should render its heading
    ✔ should link to "/"
  Home Actions
    changeUsername
      ✔ should return the correct type and the passed name
  <HomePage />
    ✔ should render the loading indicator when its loading
[...]

PR #858, opened 2016-08-12
fixing store.test.js file duplication and test execution

uzarubin:
### Solving issue #631 : 

When `npm run clean` was executed, it copied `store.test.js` to `/app/store.test.js` instead of the original location in the `tests` folder. A simple path change in `clean.js` took care of it. 

### Solving issue #458 

This one is a bit tricky, and  I would love to have some input. Currently, there is a regex inside 
`test-bundler.js` that avoids coverage for `store.js`. This regex also captures the `store.test.js` file inside the tests folder. I assumed that since we have a test for the store.js file, it would also need to be included in the coverage. So, the regex was modified to remove `store`.

### DIFF
--- a/internals/scripts/clean.js
+++ b/internals/scripts/clean.js
@@ -80,7 +80,7 @@ cp('internals/templates/index.html', 'app/index.html');
 cp('internals/templates/reducers.js', 'app/reducers.js');
 cp('internals/templates/routes.js', 'app/routes.js');
 cp('internals/templates/store.js', 'app/store.js');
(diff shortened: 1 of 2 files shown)


mxstbr:
This looks about right, awesome! Let me check out the Travis logs… 🚢

mxstbr:
This works perfectly, amazing stuff @uzarubin! Thanks so much!

uzarubin:
Awesome. Glad I could help!

PR #860, opened 2016-08-12
fix(internal): make `rm -rf .git` work in Windows console

Dattaya:
Windows' cmd and powershell do not support `rm` command, hopefully this part of the script now works in all of three major OS's. One question—`dir` argument is unused in `cleanRepo` and `initGit`, should I use it in `path.join` or delete it a altogether?

### DIFF
--- a/internals/scripts/setup.js
+++ b/internals/scripts/setup.js
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+var shell = require('shelljs');
 var exec = require('child_process').exec;
 var path = require('path');
 var fs   = require('fs');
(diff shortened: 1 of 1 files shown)


mxstbr:
Awesome, thanks so much!

> dir argument is unused in cleanRepo and initGit, should I use it in path.join or delete it a altogether?

You can delete it altogether, would appreciate a followup PR!

jwinn:
@mxstbr Should we just change the `presetup` task to install `shelljs` with `chalk` (e.g. `"presetup": "npm i chalk shelljs"`) as it'd (now) similarly--to chalk--be required for `internals/scripts/setup.js` to work?

gihrig:
@jwinn sounds like a good solution to me.

mxstbr:
:+1: agree with @gihrig, sounds reasonable!

jwinn:
@Dattaya would you mind making the change and resubmitting? (I want you to get credit for the submission) :-)

All that needs updated is to modify the `scripts` -> `presetup` value to `npm i chalk shelljs` in `package.json`

Dattaya:
Sorry about the regression, using `shelljs` before it's installed was a genius strategy, wasn't it :stuck_out_tongue: 
@jwinn, @mxstbr, @gihrig, follow up PR: #876

PR #862, opened 2016-08-14
chore(deps): update deps roll-up aug-13-2016

gihrig:
chore(deps): update deps roll-up aug-13-2016

npm run setup
Tested Dev, Dev tunnel, Features, lang = de, NotFound, Production, off-line
npm run clean
npm run generate container Test /w defaults
npm run generate route Test - /test
git commit
 
Node 6.3.1
 npm 3.10.5 
Mac OS 10.11.6
Chrome 52.0.2743.116 (64-bit)

 lodash                          4.14.1  →  4.15.0 
 react-router-scroll              0.2.1  →   0.3.1 <--- rolled back
 redux-immutable                  3.0.6  →   3.0.7 
 eslint                           3.2.2  →   3.3.0 
 eslint-config-airbnb            10.0.0  →  10.0.1 
 eslint-import-resolver-webpack   0.4.0  →   0.5.1 
 eslint-plugin-import            1.12.0  →  1.13.0 <--- UNRELEASED
 eslint-plugin-jsx-a11y           2.0.1  →   2.1.0 
 karma                            1.1.2  →   1.2.0 
 karma-webpack                    1.7.0  →   1.8.0 
 mocha                            3.0.1  →   3.0.2 

ERROR: app.js?2de9:126 Uncaught TypeError: __WEBPACK_IMPORTED_MODULE_10_react_router_scroll___default(...) is not a function

Caused by  react-router-scroll         0.2.1  →   0.3.0 

Rolled back to 0.2.1 apparently waiting for react-router 3 https://github.com/taion/react-router-scroll/issues/28.

eslint-plugin-import@1.13.0 listed on npm is not available (unreleased). 
see: [CHANGELOG](https://github.com/benmosher/eslint-plugin-import/blob/master/CHANGELOG.md#unreleased).

```
npm run clean
Cleanup started... ✓
/Users/glen/Documents/JavaScript/Research/react-bp/app/tests/store.test.js
  6:28  error  Unable to resolve path to module './store'  import/no-unresolved

pre-commit: We've failed to pass the specified git pre-commit hooks as the `lint:staged`
pre-commit: hook returned an exit code (1). If you're feeling adventurous you can
```

`./store` is an error -> `../store` Will open a separate PR see #863.

PR #863, opened 2016-08-14
fix(clean): minor errors in cleaned boilerplate

gihrig:
templates/store.test.js - ./store is an error (should be ../store)
Requires eslint-disable-line - store.test.js is moved to tests folder by clean.

*/styles.css.hbs - stylelint-ignore is an error (should be stylelint-disable)


gihrig:
Will need to be merged with #862 before it will pass CI. There is a problem with eslint and friends present in the current dependency tree (before #862).

internals/templates/store.test.js
```diff
@@ -3,7 +3,7 @@
  */
 
 import expect from 'expect';
-import configureStore from './store';
+import configureStore from '../store'; // eslint-disable-line
```
gihrig: eslint is disabled for this line as ../store is invalid when store.test.js is located in templates as a sibling of store.js. Hence ./store which satisfied eslint.

When copied into app/tests it is a level below and ./store fails to resolve.

Disabling eslint for the line seemed the most expedient solution. Really, IMO, the templates folder should mirror the app folder structure more completely.

mxstbr:
Awesome, :shipit: when the other PR is landed! :fire: 

> IMO, the templates folder should mirror the app folder structure more completely.

I totally agree, would you mind opening an issue to track this?

uzarubin:
I believe my PR caused this issue when I switched over to using `tests/`

mxstbr:
Can we rebate this into master and land it?

mxstbr:
Awesome, thanks so much for the :fire: work!

gihrig:
Really _really_ fixed by #1414 thanks to @outdooricon! ✨

PR #871, opened 2016-08-15
Add entry for vendor modules

griffinmichl:
I think it's a bit confusing to have the CommonChunksPlugin for 'vendor' without seeing anything related to it in the entry. I spent some time trying to figure out if I was missing something. I believe the code here is what the user is expected to add. Given that the vendor bundle gets created either way, I think this is a bit more explicit and helpful.

### DIFF
--- a/internals/webpack/webpack.prod.babel.js
+++ b/internals/webpack/webpack.prod.babel.js
@@ -12,9 +12,10 @@ const postcssReporter = require('postcss-reporter');
 
 module.exports = require('./webpack.base.babel')({
   // In production, we skip all hot-reloading stuff
-  entry: [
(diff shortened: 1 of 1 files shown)


mxstbr:
Sorry, I'm not 100% on what the benefit here is?

gm758:
I just thought it would add some clarity to the configuration of CommonChunksPlugin down belong for first time users.

mxstbr:
I don't think I want to merge this, I think this is just more confusing for users. ("Why is there an empty array? Should I add my vendor deps to that array?") I think it's fine as-is – thanks for taking the time to submit this PR, we appreciate it!

PR #880, opened 2016-08-17
Mention Create React App

gaearon:
Seems like a good idea to me

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -69,6 +69,11 @@ project's developers might not want to merge into the project.
 Please adhere to the coding conventions used throughout a project (indentation,
 accurate comments, etc.) and any other requirements (such as test coverage).
 
+Since the `master` branch is what people actually use in production, we have a
+`dev` branch that unstable changes get merged into first. Only when we
+consider that stable we merge it into the `master` branch and release the
+changes for real.
+
 Adhering to the following process is the best way to get your work
 included in the project:
 
@@ -86,11 +91,11 @@ included in the project:
 2. If you cloned a while ago, get the latest changes from upstream:
 
    ```bash
-   git checkout master
-   git pull upstream master
+   git checkout dev
+   git pull upstream dev
    ```
 
-3. Create a new topic branch (off the main project development branch) to contain your feature, change, or fix:
+3. Create a new topic branch (off the `dev` branch) to contain your feature, change, or fix:
 
    ```bash
    git checkout -b <topic-branch-name>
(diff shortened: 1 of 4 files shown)


kachkaev:
Why not to mention this boilerplate in react-create-app's README in return? :-)

gaearon:
One of the reasons for Create React App’s existence is that people used to complain “omg I have to learn redux, sagas and ton of other things to create a React hello world”. I don’t think pointing beginners to it is a good idea 😄 . Advanced users will find it anyway I think.

kachkaev:
I understand this motivation and I′m actually really glad that you guys created that repo. I wish it existed half a year ago when I was entering this world of lots scary worlds and my head was just exploding :–)  React-create-app is something I now recommend to colleagues to make their first steps in React.

But at the same time, react-create-app is something that the newcomers might want to outgrow at some point. Or at least it would be good for them to know that there exist quite an advanced ways of linking React with lots of other things. This understanding will help them build a better initial model of the ecosystem and increase the chances of preferring React to Angular, for example.

Mentioning the most starred boilerplates like this one in some ‘advanced alternatives’ section won't harm, but instead provide people with richer set of options, which they can switch between at different stages of their journey into React.

mxstbr:
I agree with @kachkaev here, that seems fair! _ping_ @gaearon

kachkaev:
This might be relevant here too: https://github.com/facebookincubator/create-react-app/issues/580

gaearon:
I'm hesitant partly because of naming. I feel that if we link to something called "react boilerplate" too many beginners will go there because it sounds familiar. It's a model they already understand. But react-boilerplate is not really about React. If anything, this project is react/redux boilerplate, with an opinionated twist (sagas). I recognize these are valuable but it being called react-boilerplate contributes to the confusion about what react is, and to the fear of missing out. I don't have a good solution to this but I hope you can see where I'm coming from.

gaearon:
The only users RB is going to lose by linkjng to CRA is those who didn't need RB anyway. Those who are looking for runtime setup will see that CRA doesn't provide it.

On the other hand, people completely new to CRA might not even understand the difference between runtime and build setup, and can't assess the difference. Those who need runtime setup will google for "react boilerplate" if they are unsatisfied with CRA, and this project will be their first hit anyway.

gaearon:
Closing as there seems to be no interest from the maintainers. I added some more thoughts on how I think RB approach could be improved here: https://github.com/facebookincubator/create-react-app/issues/580#issuecomment-249362743. Cheers!

PR #886, opened 2016-08-19
Adding webpack-dashboard

kenwheeler:
Adds webpack-dashboard. cc @mxstbr


mxstbr:
This is such a trivial change, I'm wondering if it's worth including by default. Let's ask the users!

**React with a 👍 if you think this should be merged, react with a 👎 if you think we should leave this optional!** (reasonings as comments appreciated too)

hummlas:
Does this play nice with cygwin/babun on Windows? I've had some issues with webpack-dashboard on Windows before, but I'm currently on a Mac so cannot test it myself.

gihrig:
I voted 👎  simply because I don't feel the extra 'clutter' is justified in all cases.

Having said that, webpack-dashboard is a _cool_ plugin. I would love to see an addition to the docs covering how to make the changes covered in this PR.

kachkaev:
@gihrig but if it is cool (like cssmodules, sagas, reselect, etc.), why not to go for _enabled-by-default_ strategy? This boilerplate does so for many features and then suggests to uninstall unwanted things in `docs/*/remove.md`. What's the principal difference between `webpack-dashboard` and, let's say, `offline-plugin` in terms of their right to be in the boilerplate by default?

samit4me:
👍 on OS X
👎 on Windows, but only because of [this bug](https://github.com/FormidableLabs/webpack-dashboard/issues/85)

amilajack:
@mxstbr is the poll over?

mxstbr:
This issue (https://github.com/FormidableLabs/webpack-dashboard/issues/85) is kind of a blocker, as we can't ship something by default that doesn't work for some people.

Let's close this for now and reevaluate once `webpack-dashboard` is a bit stabler, but I'll happily merge a guide on how to add it for the docs!

kenwheeler:
@mxstbr I'll let you know when I get a web view option released and/or get windows stable

PR #896, opened 2016-08-21
chore(selectors): unwrap selectors; fixes #882

### LINKED ISSUE #882: Nesting selectors inside empty function?
Is there any reason behind nesting a selector inside an empty function?
For example, this code:
```js
const selectGlobal = () => (state) => state.global;

const selectCurrentUser = () => createSelector(
  selectGlobal(),
  (globalState) => globalState.currentUser
);
```
could be rewritten like so:
```js
const selectGlobal = (state) => state.global;

const selectCurrentUser = () => createSelector(
  selectGlobal,
  (globalState) => globalState.currentUser
);
```
same for `selectCurrentUser`.
Another question that I have is `selectCurrentUser` doesn't perform any heavy computations. Was it memoized just as a showcase of how to use `reselect`?

Dattaya:

### DIFF
--- a/app/containers/App/selectors.js
+++ b/app/containers/App/selectors.js
@@ -4,25 +4,25 @@
 
 import { createSelector } from 'reselect';
 
-const selectGlobal = () => (state) => state.get('global');
+const selectGlobal = (state) => state.get('global');
 
 const selectCurrentUser = () => createSelector(
-  selectGlobal(),
+  selectGlobal,
   (globalState) => globalState.get('currentUser')
 );
 
 const selectLoading = () => createSelector(
-  selectGlobal(),
+  selectGlobal,
   (globalState) => globalState.get('loading')
 );
 
 const selectError = () => createSelector(
-  selectGlobal(),
+  selectGlobal,
   (globalState) => globalState.get('error')
 );
 
 const selectRepos = () => createSelector(
-  selectGlobal(),
+  selectGlobal,
   (globalState) => globalState.getIn(['userData', 'repositories'])
 );
 
--- a/app/containers/App/tests/selectors.test.js
+++ b/app/containers/App/tests/selectors.test.js
@@ -11,13 +11,12 @@ import {
 } from '../selectors';
 
 describe('selectGlobal', () => {
-  const globalSelector = selectGlobal();
   it('should select the global state', () => {
     const globalState = fromJS({});
     const mockedState = fromJS({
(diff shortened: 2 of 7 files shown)


mxstbr:
Thanks so so so much!

I just realized that I think we do need to do this. We're re-use them, and if we use the same instance for every container that has them the memoization gets lost? Does that make any sense?

Dattaya:
Got it. I've also just realized that you are passing a `props` object to selectors so it makes sense to create a new selector because when you transition from one route of the same type to another, the component is preserved (but otherwise there is no benefit because a component is destroyed?).
But regular selectors still could be simplified. They are in an official example in section [Sharing Selectors with Props Across Multiple Components](https://github.com/reactjs/reselect#sharing-selectors-with-props-across-multiple-components):

```js
import { createSelector } from 'reselect'

const getVisibilityFilter = (state, props) =>
  state.todoLists[props.listId].visibilityFilter

const getTodos = (state, props) =>
  state.todoLists[props.listId].todos

const makeGetVisibleTodos = () => {
  return createSelector(
    [ getVisibilityFilter, getTodos ],
    (visibilityFilter, todos) => {
      switch (visibilityFilter) {
        case 'SHOW_COMPLETED':
          return todos.filter(todo => todo.completed)
        case 'SHOW_ACTIVE':
          return todos.filter(todo => !todo.completed)
        default:
          return todos
      }
    }
  )
}

export default makeGetVisibleTodos
```

Dattaya:
I've changed my previous comment because it contained too much misinformation.

mxstbr:
You are absolutely right, that's probably the best explanation I've heard about this topic so far! :clap:

> But regular selectors still could be simplified.

This is true, but isn't it then really confusing for the user? Some selectors have to be called as functions, some can just be passed in, and you can't really know which one's which other than looking at the code.

I think it's preferable to have a few extra `() =>` than the above behaviour? Maybe?

Dattaya:
> But regular selectors still could be simplified

done; memoized selectors left untouched

Dattaya:
> This is true, but isn't it then really confusing for the user?

Please take a look at the changes, I don't think it's confusing, in official example and in Redux documentation they are all unwrapped. But I don't insist though, either way is fine with me.

Dattaya:
Closing this pr for now. I don't like that each time a regular selector is called, a new function is created, but it's not hard to refactor them to call selectors directly.

PR #901, opened 2016-08-22
chore(dev branch): merge master into dev

gihrig:
chore(dev branch): merge master into dev

Previous PR lost link from `master` to `dev`. This should fix that.


gihrig:
@mxstbr I've messed something up here, and my git fu is not up to figuring out what it is 😢 

With dev selected:
> This branch is 6 commits ahead, 4 commits behind master.

`dev` is 6 commits ahead of `master` but all of `master`s changes are visible in `dev`. So, `dev` should not be be behind `master`.

Further, when I attempt to merge `upstream/dev` into my `local/dev` a fast forward merge is not possible, a full merge is required.

This says to me that something is wrong with git's linkage between branches...

Thoughts, advice?

PS. feel free to revert either of these commits if that's appropriate.

PR #805, opened 2016-07-31
racing condition when polyfill intl in old browser

chaintan17:
Old code in `Promisel.all([])` create racing condition bug when `intl` get import AFTER the others.

image

Fix it by using `Promise()` and make them to get called in order.

(p.s. i've no idea why my pull request is mess up, many commits shown, please feel free to advise how can i make it better, thank you.)

PR #906, opened 2016-08-23
Racing condition when polyfill intl in old browser (in templates)

kachkaev:
Adds a forgotten bit to #805

### DIFF
--- a/internals/templates/app.js
+++ b/internals/templates/app.js
@@ -83,10 +83,16 @@ if (module.hot) {
 
 // Chunked polyfill for browsers without Intl support
 if (!window.Intl) {
-  Promise.all([
-    System.import('intl'),
-    System.import('intl/locale-data/jsonp/en.js'),
-  ]).then(() => render(translationMessages));
+  (new Promise((resolve) => {
+    resolve(System.import('intl'));
+  }))
+    .then(() => Promise.all([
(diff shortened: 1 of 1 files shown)


kachkaev:
A bit of explanation to help understand what this PR does.

#805 helped fix an exception in Webkit that occurred if the second polyfill promise out of two was faster. The fix worked, but only until running `npm run clean`, because the template missed the fix. This PR replaces a few lines of code in `nternals/templates/app.js`, so the bug no longer appears after `clean`.

I spotted the problem it in PhantomJS while generating some screenshots of my app, so it's not only Safari-specific.

mxstbr:
Ah yeah, good catch, thanks so much!

internals/templates/app.js
```diff
@@ -83,10 +83,16 @@ if (module.hot) {
...
 if (!window.Intl) {
-  Promise.all([
-    System.import('intl'),
-    System.import('intl/locale-data/jsonp/en.js'),
-  ]).then(() => render(translationMessages));
+  (new Promise((resolve) => {
+    resolve(System.import('intl'));
+  }))
+    .then(() => Promise.all([
+      System.import('intl/locale-data/jsonp/de.js'),
```
KevinMartin: This was changed from `en.js` to `de.js`. Is this on purpose?

Issue #909, opened 2016-08-23
Test warnings


samit4me:
When running the tests `npm test` the following warnings are mixed in with the test output:

```
ERROR: 'Warning: Failed prop type: Required prop `href` was not specified in `A`.
ERROR: 'Warning: Failed prop type: Required prop `children` was not specified in `A`.
ERROR: 'Warning: Failed prop type: Required prop `children` was not specified in `Button`.
ERROR: 'Warning: Failed prop type: Invalid prop `src` of type `object` supplied to `Img`, expected `string`.
ERROR: 'Warning: ReactComponentTreeDevtool: Missing React element for debugID 17 when building stack'
```

It does not cause any issues and does not fail the tests, it simply clutters the test output.

After some investigation, I've managed to resolve the warnings, but I'm new to open source and wasn't sure if I should submit a PR or create an Issue first?

gihrig:
@samit4me Yes, please open a PR. 👍 

See [CONTRIBUTING.md](https://github.com/mxstbr/react-boilerplate/blob/master/.github/CONTRIBUTING.md) for details.

PR #912, opened 2016-08-23
Reduce test warnings

samit4me:
PR for #909.

Reduce test warnings by passing required propTypes to the components.

To resolve the `<App />` test warning I added `babel-plugin-rewire` to mock the `Banner` so it is a string and not an object. This involved modifying the "internals" so this may need attention.

Resolving the `... ReactComponentTreeDevtool ...` warnings was a coincidence of extending the `` test coverage. This warning has been resolved in React `15.3.1`, see [React Issue 7240](https://github.com/facebook/react/issues/7240).

### DIFF
--- a/app/containers/App/tests/index.test.js
+++ b/app/containers/App/tests/index.test.js
@@ -1,11 +1,15 @@
+import App from '../index';
+import Footer from 'components/Footer';
+
 import expect from 'expect';
 import { shallow } from 'enzyme';
 import React from 'react';
 
-import App from '../index';
-import Footer from 'components/Footer';
-
 describe('<App />', () => {
+  /* eslint no-underscore-dangle: ["error", {"allow": ["__Rewire__", "__ResetDependency__"]}] */
+  before(() => App.__Rewire__('Banner', 'test.png'));
+  after(() => App.__ResetDependency__('Banner'));
+
   it('should render the logo', () => {
     const renderedComponent = shallow(
       <App />
--- a/app/containers/App/tests/index.test.js
+++ b/app/containers/App/tests/index.test.js
@@ -6,7 +6,6 @@ import { shallow } from 'enzyme';
 import React from 'react';
 
 describe('<App />', () => {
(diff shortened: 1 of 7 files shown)


app/containers/App/tests/index.test.js
```diff
@@ -1,11 +1,15 @@
...
+
 import expect from 'expect';
 import { shallow } from 'enzyme';
 import React from 'react';
 
-import App from '../index';
-import Footer from 'components/Footer';
-
 describe('<App />', () => {
+  /* eslint no-underscore-dangle: ["error", {"allow": ["__Rewire__", "__ResetDependency__"]}] */
```
mxstbr: Let's move this to the `package.json` and allow `__Rewire__` and `__ResetDependency__` globally – I have some other ideas where we could use rewire! (e.g. reducer initial state)

mxstbr:
Thanks so much for taking the time to fix this!

I love that you included rewire, good idea 👌 Also a fan of the `renderComponent` pattern, makes the tests a lot cleaner!

One small nitpick then this is good to merge.

samit4me:
Thank you for the positive feedback @mxstbr  😃 . 

Just re-read the [CONTRIBUTING](https://github.com/mxstbr/react-boilerplate/blob/master/.github/CONTRIBUTING.md) docs and noticed that PR's should be against the `dev` branch, would you like this PR changed? 

Unfortunately, this PR tests fail against the latest `dev` branch because **rewire** is not working. I'm not sure why but it's most likely got something to do with the dependency versions.

gihrig:
@samit4me 
> Just re-read the CONTRIBUTING docs and noticed that PR's should be against the dev branch, would you like this PR changed?

Yes, this PR must be submitted against `dev`. Since it breaks `dev`, it makes merging dev into `master` a delayed debugging nightmare. Merging this PR to `master` breaks the whole idea of having a `dev` branch.

Sorry about that 😢

samit4me:
No problems @gihrig , thank you for letting me know, what you explained makes allot of sense, thanks.

PR #923, opened 2016-08-28
Fix test warnings

samit4me:
PR for #909.

Fix test warnings by passing required propTypes to the components.

Added `babel-plugin-rewire` to allow dependency mocking.

_**NOTE:** the latest version of `rewire@1.0.0-rc-6` does not work with `istanbul`, but `rewire@1.0.0-rc-4` seems to work fine, see [rewire issue](https://github.com/speedskater/babel-plugin-rewire/issues/148) or [istanbul issue](https://github.com/istanbuljs/babel-plugin-istanbul/issues/22)._


bt:
I've been having the same issue lately and my stupid OCD is kicking in with my tests showing low coverage so I've been trying hard to get `rewire` to work in this boilerplate.

I've finally figured out that this issue has been fixed in the latest version of `istanbul-lib-coverage` (a dependency of `istanbul-lib-instrument`, which is a dependency of `babel-plugin-istanbul`) but because the old version is still being mentioned, we'll need to use a workaround.

I've done the hard work and forked both packages to use the newer version of `istanbul-lib-coverage`. For the time being, do the following ...

In your `package.json` file, use the following line for `babel-plugin-istanbul`:
```
"babel-plugin-istanbul": "git://github.com/bt/babel-plugin-istanbul.git#git-build",
```

or use:

```
$ npm install "git://github.com/bt/babel-plugin-istanbul.git#git-build" --save-dev
```

If we could please bring this to the attention of both packages, that'd be great. I've opened up PR/issues ([istanbul-lib-instrument#20](https://github.com/istanbuljs/istanbul-lib-instrument/issues/20) and [babel-plugin-istanbul#41](https://github.com/istanbuljs/babel-plugin-istanbul/issues/41)) respectively.

mxstbr:
Loving this, except the annoying `rewire` issue. So essentially we have to wait until they release a new version as per @bt's comment? (thanks for digging in!)

samit4me:
Thank you for the info and the work you have done @bt . I have updated to your `babel-plugin-istanbul` using the instructions 👍 and found the following:

- It works with `rewire@1.0.0-rc-4`.
- It does not work with `rewire@1.0.0-rc-6` as I get the same errors.

If there is something else I can try, I'd love to assist where possible, cheers.

@mxstbr This fix has certainly gone much deeper than I thought. As `rewire@1.0.0-rc-4` seems to work we could add `rewire` now and only update once the issue has been fixed, but in saying that I would totally understand if you wanted to wait until the issues have been resolved in the lastest versions.

bt:
Hey @samit4me, so what I've noticed is the fork that I made will fix an issue with `rewire` working with `istanbul`. It then gets into another issue which is to do with the `istanbul` HTML/LCOV reporting.

Can you do me a favour and try the following ...

- In file `internals/testing/karma.conf.js`, comment out the HTML and LCOV reporters, and only leave the `text-summary`.

Your `coverageReporting` block should look like this:

```
    coverageReporter: {
      dir: path.join(process.cwd(), 'coverage'),
      reporters: [
        { type: 'lcov', subdir: 'lcov' },
        { type: 'html', subdir: 'html' },
        { type: 'text-summary' },
      ],
    },
```

... and then run another test and see if you still get the error (with the latest `rewire`).

If this is the case, I will put in yet another PR (this time to `istanbul`) to add a guard.

samit4me:
HI @bt you are correct about the reporters, commenting them out stops the error `ERROR [coverage]: TypeError: Cannot read property 'start' of undefined`, it seems `lcov` and `html` throw 1 error each.

I tried a few different combinations and below are my findings:

#### 1) Using this PR "as is" - It Works!
`package.json`:
```JSON
"babel": {
  "env": {
    "test": {
      "plugins": [
        "rewire",
        "istanbul"
      ]
    }
  }
},
"devDependencies": {
  "babel-plugin-istanbul": "2.0.0",
  "babel-plugin-rewire": "1.0.0-rc-4",
}
```
`karma.conf.js`:
```Javascript
coverageReporter: {
  dir: path.join(process.cwd(), 'coverage'),
  reporters: [
    { type: 'lcov', subdir: 'lcov' },
    { type: 'html', subdir: 'html' },
    { type: 'text-summary' },
  ],
},
```

#### 2) After updating `babel-plugin-rewire` - It Fails! [error log](http://pastebin.com/raw/X31JPWfu)
`package.json`:
```JSON
...
"devDependencies": {
  "babel-plugin-istanbul": "2.0.0",
  "babel-plugin-rewire": "1.0.0-rc-6",
}
```
`karma.conf.js`:
```Javascript
...
```

#### 3) Switch babel [plugin order](https://babeljs.io/docs/plugins/#pluginpreset-ordering) - It Fails! [error log](http://pastebin.com/raw/yBMVyazx)
`package.json`
```JSON
"babel": {
  "env": {
    "test": {
      "plugins": [
        "istanbul",
        "rewire"
      ]
    }
  }
},
"devDependencies": {
  "babel-plugin-istanbul": "2.0.0",
  "babel-plugin-rewire": "1.0.0-rc-6",
}
```
`karma.conf.js`:
```Javascript
...
```

#### 4) Comment out `lcov` and `html` reporters - It Works!
`package.json`:
```JSON
"babel": {
  "env": {
    "test": {
      "plugins": [
        "rewire",
        "istanbul"
      ]
    }
  }
},
"devDependencies": {
  "babel-plugin-istanbul": "2.0.0",
  "babel-plugin-rewire": "1.0.0-rc-6",
}
```
`karma.conf.js`:
```Javascript
coverageReporter: {
  dir: path.join(process.cwd(), 'coverage'),
  reporters: [
    // { type: 'lcov', subdir: 'lcov' },
    // { type: 'html', subdir: 'html' },
    { type: 'text-summary' },
  ],
},
```

#### 5) Tried all of the above using `"babel-plugin-istanbul": "git://github.com/bt/babel-plugin-istanbul.git#git-build"` - Same results

#### 6) Swapped out `istanbul` for `babel-plugin-__coverage__` to test the plugin ordering and `__coverage__` also failed if `rewire` was first.

mxstbr:
I'd much rather wait until the `rewire` issue is properly solved, but I'm not that far into it – what do you think?

samit4me:
Agree, waiting feels like a better option. It looks like `rewire` is also waiting on `istanbul` to be fixed before updating the docs regarding  [plugin ordering](https://babeljs.io/docs/plugins/#pluginpreset-ordering).

mxstbr:
Any updates here?

bt:
Hey @mxstbr,

So a number of PRs had been merged into istanbul that fixed an issue that was mentioned before, but now it is failing on another line, specifically the `annotateBranches` function. I'm not sure what the stance is for the istanbul community (ie. fixing it by putting a guard in istanbul or to fix it in rewire). I used a few days to do some debugging to understand why this was happening but I didn't get any real findings.

At this point, I've given up on using Karma as my test runner and have replaced the entire testing flow in my own private fork of this repository with `ava`, including the test generators. It works with `rewire` pretty much straight out of the box. The only major problem is that it doesn't have support for test grouping and I don't think the community there is interested at all in test grouping.

Maybe that's something you'd like to consider moving forward?

samit4me:
No updates @mxstbr, I tried using the latest versions, including `babel-plugin-rewire@1.0.0-rc-7` and got the same [issues](http://pastebin.com/raw/yBMVyazx) 😞.

I was secretly hoping that Jest would resolve all of these issues, shame about the generators.

Maybe we could remove `rewire`? This would reduce the test errors, leaving only the following error.
`ERROR: 'Warning: Failed prop type: Invalid prop `src` of type `object` supplied to `Img`, expected `string`.
    in Img'`

gihrig:
@samit4me Could you handle that with something similar to:

```javascript
  Img: React.PropTypes.oneOfType([
    React.PropTypes.object,
    React.PropTypes.string,
```

samit4me:
Your totally correct @gihrig, that certainly works, in fact, that is exactly what I did before using rewire. It was removed as it didn't feel right, the src attribute is expecting a URI.

Happy to use this PropType solution if others also agree?

mxstbr:
> Happy to use this PropType solution if others also agree?

Sure!

samit4me:
Rewire has been removed @mxstbr and the PropTypes solution recommended by @gihrig added, no more test warnings 🎉.

gihrig:
I can confirm: All tests pass no errors! 💯 

Node 6.4.0
npm 3.10.6
Mac OS 10.11.6
Chrome 53.0.2785.116 (64-bit)

mxstbr:
Awesome, let's 🚢

PR #931, opened 2016-08-30
Shrink translation messages files

chaintng:
https://github.com/mxstbr/react-boilerplate/pull/391#issuecomment-236089492

There is some work need to be done.
This pull request assume `en` as a default message.
And set the constant in `App` container.
I'm not sure if it best practice, please review

https://github.com/mxstbr/react-boilerplate/pull/931/files#diff-0bdc92e05d8f3ab638aa855679db059eR25

### DIFF
--- a/app/containers/App/constants.js
+++ b/app/containers/App/constants.js
@@ -12,3 +12,4 @@
 export const LOAD_REPOS = 'boilerplate/App/LOAD_REPOS';
 export const LOAD_REPOS_SUCCESS = 'boilerplate/App/LOAD_REPOS_SUCCESS';
 export const LOAD_REPOS_ERROR = 'boilerplate/App/LOAD_REPOS_ERROR';
+export const DEFAULT_LOCALE = 'en';
--- a/internals/templates/i18n.js
+++ b/internals/templates/i18n.js
@@ -5,6 +5,7 @@
  *
  */
 import { addLocaleData } from 'react-intl';
+import { DEFAULT_LOCALE } from 'containers/App/constants';
 
 import enLocaleData from 'react-intl/locale-data/en';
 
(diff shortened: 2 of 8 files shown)


mxstbr:
Generally not opposed to this change, I like it! Would the `extractIntl` script need to be updated?

chaintan17:
@mxstbr yes, i already made a change in this pull request.

app/containers/App/constants.js
```diff
@@ -12,3 +12,4 @@
 export const LOAD_REPOS = 'boilerplate/App/LOAD_REPOS';
 export const LOAD_REPOS_SUCCESS = 'boilerplate/App/LOAD_REPOS_SUCCESS';
 export const LOAD_REPOS_ERROR = 'boilerplate/App/LOAD_REPOS_ERROR';
+export const DEFAULT_LOCALE = 'en';
```
chaintan17: Help is needed on this line, is it okay to have this as constant value ?

gihrig:
> +export const DEFAULT_LOCALE = 'en';

I don't see why not. Seems like it might be very handy for sites/app that don't want 'en' as the default to a have single location to make the change.

internals/templates/i18n.js
```diff
@@ -16,15 +17,21 @@ import enTranslationMessages from './translations/en.json';
 
 addLocaleData(enLocaleData);
 
-const formatTranslationMessages = (messages) => {
+export const formatTranslationMessages = (locale, messages) => {
+  const defaultFormattedMessages = locale !== DEFAULT_LOCALE ? formatTranslationMessages(DEFAULT_LOCALE, eval(`${DEFAULT_LOCALE}TranslationMessages`)) : {}; // eslint-disable-line no-eval
```
chaintan17: And i have to use `eval` here, is it okay ?

gihrig:
> And i have to use eval here, is it okay ?

Airbnb a has pretty strong [opinion](https://github.com/airbnb/javascript#strings--eval) against `eval()`
> 6.5 Never use eval() on a string, it opens too many vulnerabilities.

Because you are using `eval()` to process "code" that indirectly might be entered by someone other than the developer (translation messages) and given `eval()`'s reputation, hence Airbnb's strong opinion, I would suggest finding an alternative approach.

I'm really not the best one to give advice on this area, but to me, the current setup, having to edit code beyond `app/translations/xx.json` and sending all translations down to the client feels incompletely thought out. I would prefer to see only the default language delivered to the client and alternate languages pulled from the server as needed. But then, I stayed away from the original discussion surrounding this feature, so I really can't complain 😉 

Besides all that that code is not working. Travis and Appveyor are complaining about ` 'enTranslationMessages' is undefined`.

chaintan17:
My last commit is removing `eval` and use fix variable name instead (because it doesn't work with `eval` anyway)

Please let me know if i could improve anything for the current approach of managing translation files.

mxstbr:
This looks great, thanks so much @chaintng, sorry for the long delay!

PR #942, opened 2016-09-01
Added configuration to copy static files to build result - fixes #941

### LINKED ISSUE #941: adding a directory with static files that get copied after building
It would be nice if we had the possibility to have a folder (call it `static` or whatever) where we could place files that we need copied after building the project. Examples for these files include

* favicons
* robots.txt
* humans.txt
* google webpage tools html
* php files

In my current example I would love to use this boilerplate as a "child" of my Laravel PHP backend, so I would like to move everything from the build into the /public/ folder of laravel. However I can't use that if the rimraf removes every trace. And I don't want to use a subfolder for better URLs.

So the way react-redux-starter-kit approaches this is by a [static folder](https://github.com/davezuko/react-redux-starter-kit/tree/master/src/static). The contents of it simply get copied to the build directory after finishing the deploy process. Plain and simple.

I guess this is just a small matter of [webpack configuration](https://github.com/davezuko/react-redux-starter-kit/blob/0bde219574933922169e0290a6e5d2fdc6b47d1e/bin/compile.js#L19), but to be honest I always feel a bit lost in webpack configs.

I would love that functionality in this boilerplate, as it seems to be way more friendly for usage in general.

lucidlemon:
added copy-webpack-plugin to address the issue mentioned in #941

### DIFF
--- a/app/static/robots.txt
+++ b/app/static/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow:
--- a/internals/webpack/webpack.prod.babel.js
+++ b/internals/webpack/webpack.prod.babel.js
@@ -4,6 +4,7 @@ const webpack = require('webpack');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 const OfflinePlugin = require('offline-plugin');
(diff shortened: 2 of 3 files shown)


mxstbr:
https://github.com/mxstbr/react-boilerplate/issues/941#issuecomment-244310696

PR #953, opened 2016-09-04
fix(deps) Move cross-env and minimist to "dependencies" section

Dattaya:
IMO this should work (if someone decides to prebuild a project themselves and install only production dependencies on a production server):
```bash
> npm run build
> rimraf node_modules
> npm i --production
> npm run start:prod
```
But it won't without cross-env and minimist


mxstbr:
Sounds reasonable, thanks!

Issue #949, opened 2016-09-02
Reducers execution accumulation problem!


mnemanja:
Reducers get accumulated when using them.

I'm experiencing this in my project, but also tested it on the clean master branch.
Log the action and then switch pages between the Features and Home. You should be seeing the accumulation of the execution. 

For example the `action.type`'s that will be logged is `@@INIT` each time once but the `@@router/LOCATION_CHANGE` every time more and more.

Node v6.2.0
npm v3.10.6
Windows 10 - anniversary update

goldo:
I think i'm having the same issue, I though it was coming from my route declarations, but I didn't take the time to try the master branch

mxstbr:
That's definitely a bug – cc @justingreenberg who should know what's up here!

justingreenberg:
i can confirm this is a bug... very bizarre. i will investigate

mnemanja:
@justingreenberg any luck with this?

Dattaya:
Does this have anything to do with `react-router-redux`? Because i see multiple `LOCATION_CHANGE` action, but only one `CHANGE_USERNAME` when I try to change a name.
![image](https://cloud.githubusercontent.com/assets/387256/18351422/ceed60f2-75e2-11e6-994b-e5cc80c64058.png)

justingreenberg:
i'm sorry for the delay in getting around to this, i'm swamped this week. i believe reducer stacking can be simply fixed by checking if the key exists before assignment... ie

```js
// asyncInjectors.js ~ line 36...
if (!Object.hasOwnProperty(store.asyncReducers, name)) {
  store.asyncReducers[name] = asyncReducer
  store.replaceReducer(createReducer(store.asyncReducers))
}
```

i think there is a better solution tho, so i would like more time to research. i will put aside a few hours over the weekend to open a PR to address this either way

as an aside: while this definitely needs to be fixed, reducers are pure functions, so duplicate actions *shouldn't* affect state. in this case, impure console.log exposed the stacking issue

goldo:
I can confirm it doesn't. I've been working on my project for 1 week without noticing that bug :)
Thanks for your help/time

—
Franck Achkouyan

mnemanja:
@justingreenberg Thanks, I appreciate it.

mnemanja:
Good job @justingreenberg

maslianok:
[...]

PR #948, opened 2016-09-02
v3.2.0

mxstbr:
*It's been two weeks since the last release, so I figured we might as well prepare a new one!*

# Changelog

- Use `babel-plugin-istanbul` for more accurate code coverage (thanks to @kane-c!)
- Move `cross-env` and `minimist` to `dependencies` so `npm install --production` works (thanks @Dattaya!)
- Update dependencies (thanks @gihrig!)
- Fix indentation of `Helmet` when generating a container (thanks @Dattaya!)
- Fix intl polyfill race condition (thanks @kachkaev!)
- Fix container template (thanks @tbash!)
- Fix spelling errors around the codebase (thanks @stevenspiel!)
- Fix reducers stacking up (thanks @justingreenberg!)
- Fix new linting errors due to eslint upgrade and enable stricter linting (thanks @peter-mouland!)
- Fix test warnings in example (thanks @samit4me!)

PR #976, opened 2016-09-11
hotfix(internals): reducer accumulation

justingreenberg:
@mnemanja @goldo @Dattaya can you guys confirm this fix for reducer accumulation bug? it's a one-liner so you don't really even need to pull down the code

**app/utils/asyncInjectors.js**
```js
 // Line 36
 if (Reflect.has(store.asyncReducers, name)) return;
```

@gihrig @mxstbr may want to include this hotfix in v3.2 (#948) if it resolves the bug (#949)... this includes internal template updates and shouldn't affect coverage.


goldo:
Sorry I am currently working on a July version, and I don't have the file update of August 3rd:
https://github.com/justingreenberg/react-boilerplate/commit/96e9441838ba2ff372e8f3ef490e7fa38d29974c

Testing that in a fresh new react-boilerplate would be easy but I have currently no time :(

Dattaya:
:+1: Now I don't experience this bug anymore (Win10), tested in Chrome, FF, Edge stable, IE11. Couldn't test in IE10 and 9, this boilerplate does't work in them for now.

mxstbr:
LGTM, @gihrig?

app/utils/asyncInjectors.js
```diff
@@ -33,6 +33,8 @@ export function injectAsyncReducer(store, isValid) {
       '(app/utils...) injectAsyncReducer: Expected `asyncReducer` to be a reducer function'
     );
 
+    if (Reflect.has(store.asyncReducers, name)) return;
```
smashercosmo: What's the benefit of using Reflect.has instead of hasOwnProperty?

mxstbr: Good question, I'm not sure either! @justingreenberg?

mnemanja:
I can confirm that the fix works. Good job @justingreenberg

mxstbr:
Merged, will be included in v3.2! Thanks for testing @mnemanja @Dattaya!

smashercosmo:
So, no answer about Reflect.has (((

app/utils/asyncInjectors.js
justingreenberg: @mxstbr @smashercosmo sorry guys just getting on my computer... I have been using Reflect mainly as a first-class replacement for `in` and `delete` operators...

> In ES5, the way to detect whether an object obj defines or inherits a certain property name is to write (name in obj). Similarly, to delete a property, one uses delete obj[name]. While dedicated syntax is nice and short, it also means you must explicitly wrap these operations in functions when you want to pass the operation around as a first-class value.

> With Reflect, these operations are readily defined as first-class functions: Reflect.has(obj, name) is the functional equivalent of (name in obj) and Reflect.deleteProperty(obj, name) is a function that does the same as delete obj[name].

there are other good reasons to use Reflect api, check out the wiki: [Harmony `Reflect`](https://github.com/tvcutsem/harmony-reflect/wiki#harmony-reflect)

Issue #955, opened 2016-09-04
Extracting component to its own npm package


pavlin-policar:
I had been working on a project for a while, and decided that one of the components was generalised enough to go into its own package, that I would then include in my main project. The component wasn't done, but it made more sense that it be separate from the main project. I tried to include the package into my main project using `npm link`. The package is written with react and es6, but has no need for HRM because its just a library. Because the HRM babel preset was not present in the package, I got this error while trying to run `npm start` in my main project:

```
Module build failed: Error: Couldn't find preset "react-hmre" relative to directory ".../local_package/lib"
```

This is not an issue with this boilerplate, but with the babel loader and webpack, there are many issues open regarding this. They also mention a workaround, which worked for me (kind of).

By changing line 61 in `internals/webpack/webpack.dev.babel.js`, from `presets: ['react-hmre'],` to `presets: [require.resolve('babel-preset-react-hmre')],` I am now able to work on both my local npm package and my main project (which uses this boilerplate) with the main project running with hot reloading for both the main project and package. I also had to add my local project to the `dllPlugin::exclude` packages. Obviously this only works for the HRM preset, and the same would have to be done to other presets as well, but this worked for me.

The key thing here is that now I can write my local package in es6 without having to build the package whenever I make a change, so in the `package.json` of my local package I can have `"main": "src/index.js"` instead of `"main": "lib/index.js"`.

This fails when trying to build the package due to previously mentioned issues in webpack. Also, the main issue for me was ease of development.

I think this should be documented somewhere (if I missed this in the docs on this, I apologise), since one of the most compelling points of the self contained containers/components to me was that you could easily extract them from the project into their own package.

My solution isn't perfect, but I would love to get input on how this could be improved, or what a better way of working on local packages in parallel would be.

mxstbr:
Absolutely agree, this warrants a discussion! I don't have time to dive in at the moment, but I would love for you to submit a PR adding docs and fixing bugs to make this easier to do!

pavlin-policar:
Sure thing, I'll do what I can!

pavlin-policar:
I've created a PR #979 that adds some the documentation and fix for what I've described above.

There are still a couple of issues and I don't have much experience with webpack to know what causes some of them. The development process using my fix is very nice, webpack updates the bundle each time the linked package is changed.

[...]

PR #979, opened 2016-09-11
feat(docs) Extracting packages into their own modules

pavlin-policar:
This addresses the issues I've pointed out #955  and attempts to fix them. It also adds some documentation of how to go about extracting components into their own packages (this is one of the most compelling arguments for self contained component architecture).

This is more of a "best effort" attempt, but it has worked well enough for me so far. I am not very familiar with webpack so there's probably a better way, but this solution does make development a lot simpler.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -10,6 +10,7 @@
   - [FAQ](general/faq.md)
   - [Gotchas](general/gotchas.md)
   - [Remove](general/remove.md)
+  - [Extracting components](general/components.md)
 - [Testing](testing)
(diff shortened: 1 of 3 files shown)


mxstbr:
Sorry for being so slow to respond here, this looks very very good to me! Would you mind rebasing on top of the newest `dev` so we can get this merged?

pavlin-policar:
No problem, tbh, I've kind of forgotten about it too. I've rebased it, as you asked.

mxstbr:
Amazing stuff, thank you so much!

PR #980, opened 2016-09-11
Add Jest

dmitriiabramov:
The two main things that need to be done first:
- `jest/jasmine` does not support generator functions in test descriptions. We can either add suporrt for it to jest and remove those tests from react boilerplate for now.
- `chai` and `jest` both use `expect` for assertions and that can get a little confusing. I can either migrate everything from `chai` to `jest-matchers`, or we can import `const jestExpect = require('jest-matchers')` separately to use jest snapshot testing feature (will require some work on jest side as well)

PR #985, opened 2016-09-12
filter test file from coverage report

### LINKED ISSUE #195: Exclude test files from coverage report
IMO it doesn't make much sense to include test files in the coverage, since they aren't being tested. I'm not sure we can do anything against them being used though – @jbinto any insights?

ktravers:
closes #195

- uses BASH commands to filter `lcov` input being piped to coveralls
- filters out all files with `test` in their path

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "test:firefox": "npm run test -- --browsers Firefox",
     "test:safari": "npm run test -- --browsers Safari",
     "test:ie": "npm run test -- --browsers IE",
-    "coveralls": "cat ./coverage/lcov/lcov.info | coveralls"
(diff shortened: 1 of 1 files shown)


mxstbr:
Yes, I like this! Can it possibly be achieved without adding another dependency?

ktravers:
@mxstbr absolutely, good call. since we're piping output to coveralls, we can use a nice simple `grep -v` to filter out any files with `test` in their path. will update description above accordingly

mxstbr:
While I really appreciate you taking the time to fix this, #980 will supersede our current test setup in the next release, so this isn't necessary anymore.

Thanks for submitting this PR, I really appreciate it! I'm sure people that use the current version will find this very helpful 👍

ktravers:
@mxstbr no problem, excited for the move to Jest/Enzyme!

PR #998, opened 2016-09-17
fix(webpack): Skip image optimisation in development

MozMorris:


mxstbr:
I'm not 100% convinced this is a positive change. Image optimization is a tricky thing, and can, under certain circumstances, mess up your images.
Having different behaviour for that in prod and dev could lead to confusion when things are messed up in prod that aren't in dev.

I see where you're coming from, but I'm not convinced this is something we want to merge/do by default…

MozMorris:
Max, thanks for your response. I shall make my case...

A reduction of ~60s in the initial boot time was enough to convince me. This is based on a project currently using only 9 pngs & 2 jpgs. We of course use a lot of SVGs and the pngs form part of our fallback.

As our project started to grow I found the slow down when booting the dev environment was mainly through image-loader optimising images. In my experience, image optimisation has always been part of the build process. Seems weird to optimise the images on every boot, even more so that these optimisations aren't persisted. I also can't recall when I've ever had a problem with imagemin messing up the optimisation, perhaps I've just been lucky.

You mentioned confusion. Isn't it just as confusing to see a difference (during development) between the browser and the image file on disk? At least in production I can quickly point to the image optimisation at build time and find the offending file on disk.

If you're still not convinced, then I'm happy to keep this timesaver for myself! 😉

Issue #794, opened 2016-07-27
Feature to remove Devtools from Production


ansumanshah:
We should strip the redux-devtools extension code in the production.

mxstbr:
Sounds like a good idea, do you want to submit a PR for that? :+1:

ansumanshah:
@mxstbr Sure, will submit a PR shortly

kachkaev:
Should this be done in all cases or be optional (i.e. backwards-compatible)?

mxstbr:
> i.e. backwards-compatible

There's no real "backwards-compatible" for a boilerplate since updating isn't really a thing. I do think this should be done by default!

maticrivo:
@ansumanshah any progress on this? I'm interested in having this feature too

ansumanshah:
Submitted a PR with the quick fix that i did for my project.

P.S. First PR so please help if something is abruptly wrong or if have to add something else too :)

zalmoxisus:
Just to notice that allowing Redux DevTools Extension in production isn't wrong and [rather helpful](https://github.com/zalmoxisus/redux-devtools-extension/issues/185), though agree that better to allow it explecitely.

mxstbr:
Yes, very true @zalmoxisus, it's just that by default it seems better to not have it! Closing since the PR was merged, this is released in v3.3! 🎉

PR #1005, opened 2016-09-19
Remove Devtools from Production 

ansumanshah:
Addresses #794 

@mxstbr should we implement this strategy as documented [here](https://github.com/gaearon/redux-devtools/blob/master/docs/Walkthrough.md#exclude-devtools-from-production-builds)


app/app.js
```diff
@@ -55,7 +55,7 @@ const store = configureStore(initialState, browserHistory);
 // could be used with the devTools' store.
 // As this boilerplate uses Redux & Redux-Saga, the `updateStore` is needed
 // if you want to `take` actions in your Sagas, dispatched from devTools.
-if (window.devToolsExtension) {
+if (process.env.NODE_ENV !== 'production' && window.devToolsExtension) {
```
ansumanshah: @mxstbr why is this of piece of code not in the templates/app.js file? I feel this is missing should I add it as well?

amilajack:
You should rebase and fix conflicts

mxstbr:
Can you update this PR with the new DevTools version syntax merged in #1129? Thanks! (just rebase on top of the latest changes and fix the conflicts)

Issue #933, opened 2016-08-30
[Enhancement] Routing transitions using react-router-transition


KarandikarMihir:
First of all, thank you for creating this state-of-the-art boilerplate! We're using this boilerplate in our project. 

One thing that I observed is that route transitions are not smooth (because of chunking) and cause a terrible user experience especially for the ones having slow internet. So we decided to use [react-router-transition](https://github.com/maisano/react-router-transition). 

Since, most of the users of this boilerplate will face this issue at some point of time, can we include react-router-transition in the boilerplate itself?

mxstbr:
`react-router-transition` is a bit too flashy imo, but I'd love to add a loading indicator while a new component is loaded. (something like the github top-loading-bar thingy) If you have any idea how to do that I'd love a PR!

KarandikarMihir:
Alright, I'll give it a try, once I'm on my time.

KarandikarMihir:
@mxstbr Just drawing your attention at [this MVP](https://github.com/KarandikarMihir/react-boilerplate/tree/feature-page-progress-bar). Could you take a look at it? Is that how we want to do it? It's very crude. I'd love to know what you think.

mxstbr:
Generally a fan, but I'm wondering if there's a way to integrate with `react-router` to get the current loading percentage instead of faking it…

KarandikarMihir:
@mxstbr Great. That gives a direction to me. I'll try to whip up something.

KarandikarMihir:
@mxstbr From what I have read, there's no way to calculate progress. Since `System.import` is built on top of `Promise`, there's no way to get real loading percentage. Simply because `Promise` does not know how much time it's going to take to `resolve`. That being said, I found [progress-promise](https://github.com/numtel/progress-promise). But `progress-promise` also fakes the progress.

 That leaves us with `react-router` hooks. Your thoughts?

mxstbr:
Thanks for digging in, that sounds way too hacky – let's just go with your solution in the example app, that's good enough! Could you submit a PR, and we'll iterate from there? Thanks so much!

PR #1008, opened 2016-09-21
Feature/route progress bar

KarandikarMihir:
This PR is pertaining to #933 

**Motivation for the new feature:**
Route transitions are not smooth over the network because chunks generated by webpack take some time to load. It causes a terrible user experience especially for the ones having slow internet. So it would make sense to show some kind of progress bar or loader to the users while browser downloads the chunk and renders it.

**Changes in the project**
1. Created a new `ProgressBar` component. Added tests for the same. (Might need some help with the test coverage)
2. Created a new Higher order component `withProgressBar`. Wrote tests.
2. ~~Converted React component in `App/index` from stateless function to es6 class. Added tests.~~
3. ~~Please note that a higher order component (HoC) `withRouter` (from `react-router`) is used in `App/index` to access newer API of `react-router`.~~

**How it works**
The `ProgressBar` percentage depends on `listen` hook of `react-router` and React's component life cycle methods.
1. Initially, the percentage is `-1`.
2. Progress is set to `0` when the `location` changes and it keeps on increasing until the new route is mounted or it reaches `99` (whichever is faster).
3. Percentage is set to `100` when the new route mounts.
4. Percentage is reset to `-1` and `ProgressBar` is ready for a new route change.


mxstbr:
This looks amazing! ([deploy preview](http://deploy-preview-1008.react-boilerplate.netlify.com/) to see this live without cloning)

One thing I'm not a fan of (this is without a code review, will do that in due time) is how the progress bar animates back to 0 when the route change is finished. Instead of retracting back to the left, can it fade out or slide out to the top?

I think that's a nicer experience, just from a visual perspective!

KarandikarMihir:
@mxstbr You're right! It is a little annoying. I'll fix it. Just waiting for you to review the code, so that I can push everything at once.

app/containers/App/index.js
```diff
@@ -8,38 +8,61 @@
...
-    </div>
-  );
+export class App extends React.Component {
+
+  state = {
+    progress: -1,
+  };
+
+  componentDidMount() {
+    this.props.router.listen(() => this.setState({ progress: 0 }));
```
justingreenberg: nice work 👍

one quick thing... we need to store a reference to listener...

```js
this.unsubscribeHistory = this.props.router.listen(() => this.setState({ progress: 0 }));
```

and use lifecycle hook for teardown

```js
//...
componentWillUnmount () {
  this.unsubscribeHistory
}
```

otherwise this will cause memory leak since listeners won't be garbage collected :)

mxstbr:
I won't have time today or tomorrow to review the code in depth, so feel free to push the change! Also, @justingreenberg's comments are usually on-point 👌

app/containers/App/index.js
KarandikarMihir: Thanks! Nice catch! Out of curiosity, is there any way to check if it is really garbage-collected?

justingreenberg: if you pass a named function, you should see it pop off the call stack when transition hook is removed from listener array! (check out [createTransitionManager.js#L53](https://github.com/mjackson/history/blob/master/modules/createTransitionManager.js#L53))

sourcesoft:
I patched my own app with this pull request and it seems like working. I also updated my beta/soon-to-be-production website using your awesome boilerplate. check it out since it's big enough to test this progress bar:
http://thirdeye-render.com/

KarandikarMihir:
@sourcesoft That's great! Thanks for giving a live preview! However, I did find some quirks on mobile. Sometimes, I saw progress bar running twice after page load :confused: Can you confirm the same, please? 

@mxstbr Did you get any chance to see the code? Just checking in! I'm concerned about that quirk. If you could also reproduce it, maybe you could find the reason?

sedubois:
Should/could this be provided through an external library rather than within RBP's codebase?

KarandikarMihir:
@sedubois Yes! I had thought of it before submitting this PR. Creating another library will create a dependency. That library will have to be maintained by someone. And like most cases, that repo will be abandoned. 

If we keep `ProgressBar

KarandikarMihir:
@sedubois I had thought of it before submitting this PR. Creating another library will create a dependency. That library will have to be maintained by someone. And like most cases, it will be abandoned.

If we keep `ProgressBar` inside RBP, we will have two advantages-
1. People will maintain and improve the component.
2. Users who do not wish to use it can easily get rid of it by simply deleting the folder from `components` directory.

sedubois:
That's understandable, but according to that logic, then we could end up with this repository engulfing the entire universe ;-) It's just a matter of scope I guess.

There could also be the opposite argument that, by making it a lib, you allow other projects to reference it, and therefore it gains visibility.

sourcesoft:
@KarandikarMihir After disabling cache and choosing a really slow throttling in network tab of chrome dev-tools, it seems that if it takes long, this happens once before transition starts and once again after transition ends and new chunks are loaded, but if the connection isn't like unreliable mobile you can't see it happen twice.

KarandikarMihir:
@sedubois Hahah! :smile:  We'll leave it up to @mxstbr to decide. Let's see what he has got to say about that. Of course, you can go ahead and make a lib out of this! I, for one, know that I won't maintain a repo of my own. I have a very short attention-span :sweat_smile: 

@sourcesoft Thanks for confirming :+1:  It's strange! I'll look into it.

mxstbr:
> Should/could this be provided through an external library rather than within RBP's codebase?

This is _just_ in the example app, and will be removed after `npm run clean`.

It shows how to add a progress bar when loading a new route, something that every app with large pages/lots of content will do – otherwise users navigate to a new route, and nothing happens at all until the new route is loaded, which is horrible UX. We shouldn't be promoting that 😊

app/components/ProgressBar/index.js
```diff
@@ -0,0 +1,102 @@
+/**
+*
+* ProgressBar
+*
+*/
```
mxstbr: Add spaces before the individual `*`, i.e.

```JS
/**
 *<- SPACE BEFORE THE ASTERISK
 *<- SPACE BEFORE THE ASTERISK
 */
```

app/components/ProgressBar/styles.css
```diff
@@ -0,0 +1,22 @@
+.reactProgressBar {
```
mxstbr: No need to prefix `react` before every class, just call this `.progressBar`

app/components/ProgressBar/styles.css
```diff
@@ -0,0 +1,22 @@
...
+  top: 0;
+  left: 0;
+  width: 100%;
+  visibility: visible;
+  opacity: 1;
+  transition: all 400ms;
+  z-index: 9999;
+}
+
+.reactProgressBar.reactProgressBarHide {
```
mxstbr: No need to prefix `react`

app/components/ProgressBar/styles.css
```diff
@@ -0,0 +1,22 @@
...
+  opacity: 0;
+  visibility: hidden;
+  z-index: -10;
+}
+
+.reactProgressBarPercent {
```
mxstbr: No need to prefix `react`

app/components/ProgressBar/tests/index.test.js
```diff
@@ -0,0 +1,165 @@
+import ProgressBar from '../index';
+import styles from '../styles.css';
+
+import expect from 'expect';
+import sinon from 'sinon';
+import { shallow, mount } from 'enzyme';
+import React from 'react';
+
+describe('<ProgressBar />', () => {
+  beforeEach(function() { // eslint-disable-line
```
mxstbr: Why are we disabling eslint here?

app/components/ProgressBar/tests/index.test.js
```diff
@@ -0,0 +1,165 @@
...
+    this.clock = sinon.useFakeTimers();
+  });
+
+  afterEach(function() { // eslint-disable-line
```
mxstbr: Why are we disabling eslint here?

package.json
```diff
@@ -105,7 +105,10 @@
       }
     },
     "rules": {
-      "arrow-parens": ["error", "always"],
+      "arrow-parens": [
+        "error",
+        "always"
+      ],
```
mxstbr: Undo whitespace changes

package.json
```diff
@@ -141,7 +144,10 @@
       "prefer-template": 2,
       "class-methods-use-this": 0,
       "react/forbid-prop-types": 0,
-      "react/jsx-first-prop-new-line": [2, "multiline"],
+      "react/jsx-first-prop-new-line": [
+        2,
+        "multiline"
+      ],
```
mxstbr: Undo whitespace changes

GGAlanSmithee:
@KarandikarMihir this is super neat! It's funny how something this subtle really makes a difference Like @sourcesoft , I also patched this PR to my WIP project.

package.json
sedubois: IMHO it's good to have this whitespace change (even if you'd prefer to have it in other PR) because npm does this automatically and it's annoying to revert every time.

app/components/ProgressBar/index.js
KarandikarMihir: Done

app/components/ProgressBar/styles.css
KarandikarMihir: Done

app/components/ProgressBar/styles.css
KarandikarMihir: Done

app/components/ProgressBar/styles.css
KarandikarMihir: Done

app/components/ProgressBar/tests/index.test.js
KarandikarMihir: It had something to do with the scope of `this`. Nevermind. Fixed.

app/components/ProgressBar/tests/index.test.js
KarandikarMihir: It had something to do with the scope of `this`. Nevermind. Fixed.

package.json
KarandikarMihir: How did it even change in the first place? I did not touch `package.json`.

sedubois: Maybe when you ran some `npm` command such as `npm i`.

KarandikarMihir:
@GGAlanSmithee Thanks so much! I'd love to see it in action :)

package.json
KarandikarMihir: @sedubois Okay. @mxstbr I didn't get what you wanted to say. Even [master branch file](https://github.com/mxstbr/react-boilerplate/blob/master/package.json#L107) looks identical to this one.

mxstbr:
Sorry, can you rebase this on top of `dev` with `styled-components` and fix all conflicts? Then we can finally land this! 🎉

mxstbr:
Sorry for letting this hang for so long 😊

KarandikarMihir:
@mxstbr Sure thing!

KarandikarMihir:
@mxstbr If you see the test results, there are 4 lines in `App/index.js` untested. This is because I couldn't test `<App />` separately, there seems to be some problem with `react-intl`. I need a little help with that... otherwise, I think we're good.

web2style:
To my mind it transitions too quickly. When I test it with slow connection, it always transitions much much faster than the page changes and we still have this effect: progress bar animation has finished and already disappeared, but still nothing happened with with the page itself. Why don't we make if work like youtube's one? If transitions to about 60%, then stops and finishes after the new page has rendered. This is much more intuitive. What do you think?

KarandikarMihir:
@web2style You're right. I'm working on it. Please let me know if you have any ideas to improve the experience.

KarandikarMihir:
@web2style Yes. I'm working on it.

KarandikarMihir:
@web2style Hi. Thanks for pointing that out. I've made some subtle changes that handles the progress bar better. If I'm not mistaken, 'componentDidUpdate()` and `componentWillReceiveProps()` are the only two methods where we can call `setState()`. 

You won't see the exact same thing like YouTube/Github. That is because their progress bar has some kind of support from their back end which predicts the loading time. We had a [discussion](https://github.com/mxstbr/react-boilerplate/issues/933#issuecomment-248053516) on this. I didn't find any way to calculate loading time of javascript chunk. 

The whole thing relies on `listen()` method of `react-router` to know that the route has changed. And when the new child mounts in `<App />`, it should complete the progress. If you can think of some better hook, I'll be more than happy to see it working! After all, it's a small part of example app. Actual users of this boilerplate can use it better with a little help from their back end.

mxstbr:
Can we finally land this? Please review one last time @Dattaya @gihrig I'd love to finally finally merge this!

KarandikarMihir:
@Dattaya I'm glad that you look at it as a candidate. +1 for the last point. We shouldn't show progress for previously loaded routes. Thanks for taking time out for this. With all the flaws fixed, this could be a nice feature of rbp.

mxstbr:
Where does 1.3% decrease in coverage come from? Now that I see it, that would be good to get up to 99% again!

KarandikarMihir:
@mxstbr It comes from [here](https://github.com/mxstbr/react-boilerplate/pull/1008#issuecomment-260887611). I was just wondering how did you approved changes without saying anything about coverage :sweat_smile: I'll need a little help with this...

gihrig:
Once the app is loaded, transitions between routes look really nice :sparkles:

But... I'm seeing issues in local environment, production build:

OS X: 10.11.6
Chrome: 55.0.2883.87 (64-bit)
Safari: 10.0.1 (11602.2.14.0.7)
Firefox: 49.0

No connection throttling.

When the app is loaded (first load or CMD-R):
- The progress bar appears and completes in ~.25 sec.
- Fades out
- Reappears and progress to about 95% in ~9 sec.
- Stalls there, never clears.

On a route change (Features) operation becomes normal.

gihrig:
> 	Merge branch 'dev' into feature/route-progress-bar

Brings import order in line with #1275.

gihrig:
As for help with coverage, do you know about the `lcov-report` that is produced on build? After 'npm run build' see: 
`/coverage/lcov-report/index.html`
from here you can drill down to points of interest, e.g.
`/coverage/lcov-report/app/components/ProgressBar/index.js.html`

I think the < 100% branch coverage is here:
image

If I'm interpreting that report correctly.

gihrig:
Oh Bother! I broke your build with that merge :sad:

image

This is also the result of #1275 (higher code quality standards :wink:)

Let me see if I can fix that...

app/containers/App/tests/index.test.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 
+import ProgressBar from 'components/ProgressBar';
 import Header from 'components/Header';
 import Footer from 'components/Footer';
 import App from '../index';
```
gihrig: Well, isn't his fun, the Github online editor returns 404 on this file...

The problem created by merging `dev` is:

`containers/App/tests/index.test.js`
```javascript
import App from '../index';
```
Needs to change to:
```javascript
import { App } from '../index';
```

I don't understand why the braces are necessary considering this from `../index`:
```javascript
export default withRouter(App);
```

Something strange about the `withRouter` wrapper? But that's how it was when it worked and it works (for me) when I add the braces.

KarandikarMihir:
@gihrig @mxstbr I'll get back to this once I get a chance to be on my laptop. 😓

mxstbr:
> I would like to see this fixed before merging, but if others disagree, I won't hold it up.

Yeah that should definitely be fixed!

app/containers/App/tests/index.test.js
KarandikarMihir: This is getting weird. Might take some time to figure this one out. I'm on it...

KarandikarMihir:
@mxstbr Roger that

KarandikarMihir:
This will take some time. I'm on time-out for personal reasons.

mxstbr:
I hope everything is alright, looking forward to seeing you back around!

KarandikarMihir:
@mxstbr I'll be back in a few days. Can't wait to fix this!

ItsMrAkhil:
I am eagerly waiting for this. When it will be released.

KarandikarMihir:
@ItsMrAkhil Working on it! Sorry for the delay!

KarandikarMihir:
@gihrig Thanks for the in-depth analysis. I don't really want this merged unless it is completely acceptable to all reviewers. 

I've refined the behavior at the cost of keeping more boilerplate code in `<App />`. But I think, it does show a creative way of using `withRouter` to RBP adopters. Can you reconfirm that it indeed behaves the way we want? 

Testing part is still impaired. I'll update tests when I find some time.

ItsMrAkhil:
@KarandikarMihir It's great to here from you. Our team actually working on a application that is actually going to hit production in 15 days. And it's a wonderful boiler-plate that one can use.

gihrig:
Well, I guess it's perfect 😄 

Works great at various throttling settings ✨ 

The only "issue" is that in an offline-first app, a loading indicator is largely irrelevant 😢 

- It can't show on the initial page load because there is no page to show it on...
- Shows well on first page load made through an internal route change e.g. button click.
- Does not show on subsequent page (re)load on route change. There's no need as nothing is loaded from the server. It's all internal from service worker/cache (offline-first) and pretty much instant.

But still, it's a nice addition. Get that test coverage up and I'll gladly approve :tada:

KarandikarMihir:
*whew*. The latest commit has some major changes. 

- I brought down `<App />` back to a simple stateless function.
- Updated tests of `<App />`.
- Removed `withRouter` that comes with `react-router`.
- Created a new HoC called `withProgressBar`.
- Improved logic to prevent re-rendering of progress.
- Wrote new tests for `withProgressBar`.

**Things `<ProgressBar />` cannot handle:**
- Offline mode
- Progress before boilerplate loads in the DOM.

**ONE MINOR (MAJOR?) WORKAROUND**
I could not test `router.listenBefore`. If someone could [help me with this](https://github.com/mxstbr/react-boilerplate/pull/1008/files#diff-7053742290338ec89c91ae0b953a35c2R16), I'd appreciate it very much. 
screenshot from 2016-12-21 17 23 15

This is the original test coverage-
screenshot from 2016-12-21 18 12 03

So I had cheat (`/* istanbul ignore next */`. ) to make the build pass.

smashercosmo:
Wouldn't it be more convenient if withProgressBar HOC renders ProgressBar component, so you don't need to put it in the markup?

ItsMrAkhil:
It seems great. How can start using this now?

KarandikarMihir:
You can patch this into your boilerplate or manually copy progress bar related files.

ItsMrAkhil:
Thank will you please let me know what are those files.

KarandikarMihir:
@smashercosmo I missed that. Thank you! 😊 @ItsMrAkhil  you can patch this branch or manually copy files in your code base.

KarandikarMihir:
@smashercosmo Done.

@web2style I think this fixes the issue you had raised.
> When I test it with slow connection, it always transitions much much faster than the page changes 

It was actually a problem of re-rendering. That's why it was transitioning way too faster than it should and it was transitioning twice.. I'd like to hear from you, see if the latest changes are up to the mark.

And I take my words back!
>You won't see the exact same thing like YouTube/Github. 

I think this progress bar pretty much emulates their progress bar.

web2style:
@KarandikarMihir It looks much better now. Thank you!

BUT. I tested the app with throttling and noticed a very unpleasant moment. When we open the app with slow connection, we just see white screen. For a very long time. I'm quite sure most users will just close the window thinking the app doesn't work.

Maybe we add a preloader? What do you think? I could make a PR.

KarandikarMihir:
@web2style Thanks for confirming!

>When we open the app with slow connection, we just see white screen.

Yes. That still remains a problem. 

>Things `<ProgressBar />` cannot handle:
Offline mode
Progress before boilerplate loads in the DOM.

It's up to @mxstbr and collaborators to decide if we want a preloader. :wink: I think it'll be a nice feature.

mxstbr:
> Maybe we add a preloader? What do you think?

Let's make that a separate PR so we can finally land and ship this amazing improvement?

This looks great to me! What do you think @gihrig?

KarandikarMihir:
@gihrig :smile: Yo! 

I'll make those changes right away. 
>have docs briefly describing use of ProgressBar component and Remove instructions.

I should write those instructions in [remove.md](https://github.com/mxstbr/react-boilerplate/blob/master/docs/js/remove.md)?

> integrated into components/ProgressBar.

I'm not sure what you mean by "integrated". Do you want me to just move `withRouter.js` and `withRouter.test.js` in `components/ProgressBar`? 

I'll just push new changes. If they're acceptable, :shipit:

gihrig:
> I'm not sure what you mean by "integrated". Do you want me to just move withRouter.js and withRouter.test.js in components/ProgressBar?

That's right (if practical). 

As it stands, `components/ProgressBar` requires considerable modification to the target component. `containers/App/withProgressBar.js` represents what feels to me like excessive coupling between components/ProgressBar and containers/App.

IMO, the ideal would be in `containers/App` to simply import `components/ProgressBar` and wrap the App jsx in a `<ProgressBar>...` element. or as you are doing now using the `withProgressBar` HOC.

gihrig:
> Move withProgressBar to components/ProgressBar

Really nice 🎉

gihrig:
> But, it will be removed automatically after npm run clean, right?

While that's true, the example app does provide many reusable components that (hopefully) represent best practices. I'm pretty sure that all your hard work on this will live on in many apps created with RBP. I know I'll use it in mine. 😃

KarandikarMihir:
Oh, I pushed new changes without reading your comments :| 

>I know I'll use it in mine. 

I'm so happy to know that!! :smile: 

The last thing to do is describe how `<ProgressBar />` works :) Should take a few minutes.

KarandikarMihir:
@gihrig Done! :godmode: :sparkles:

ItsMrAkhil:
Hi. If I add `withProgressBar(connect(mapStateToProps, mapDispatchToProps)(App))` inside my App -> index.js It is working like charm. But whenever I shifted the places of `connect` and `withProgressBar` the browser it getting crashed. I've tried in both **Firefox, and Chrome browswers in Ubuntu 14.04**

KarandikarMihir:
@ItsMrAkhil  Then don't shift the places :sweat_smile: I'd say wait for the final release, because the code can change at the last moment. By default `<App />` is not a `connect()`ed component. But if you really want to connect it to the store, you'll have to do whatever that works, since your concern is not directly related to working of `<ProgressBar />`.

ItsMrAkhil:
@KarandikarMihir 😂.. Of course. But I'm just informing. That's it.

gihrig:
@KarandikarMihir I have made minor edits to your doc, but I am unable to get them to you (in retrospect, this project probably should have occurred in mxstbr/branch not a PR).

- I am not able edit your PR (GH `edit` returns a 404)
- Attempted to fork your repo but GH does nothing (mine just another mxstbr branch)
- Attempted to open a PR across forks, but GH says "nothing to compare"

Don't know if it can be done on an open PR, but GH has an "Allow edits from maintainers" checkbox at the bottom of the "Open a Pull Request" page. If you check that, I think I can edit your doc in the GH editor.

If that doesn't work you can see my changes [here](https://github.com/gihrig/react-boilerplate/commit/6ba2a09907dff29520b5c2a37dec6b3e3f5b07f4) and edit from there.

KarandikarMihir:
@gihrig I had never committed changes directly on GH. No idea why you couldn't make changes. Before opening this PR, I think I had enabled editing by collaborators.

KarandikarMihir:
@giautm can you create an issue and subsequently send a PR? If it's breaking UI, we should definitely remove extra dom nodes. Just make sure it works with your way before you do.

yantakus:
@KarandikarMihir I've found a bug in the ProgressBar. If we have an error in the console, for example invariant violation from React, and transition to another route, the ProgressBar never disappears, it hangs in the following state:

sign up step 2 2017-01-19 21-03-38

ItsMrAkhil:
@yantakus As I know if something goes wrong with javascript, it will break (in the case of uncaught exception). From that point, it will never work until you refresh. I think this does not an issue with the progress bar. Correct me if I'm wrong.

yantakus:
@ItsMrAkhil It's weird, I can't reproduce the issue today with the same error in the console. Everything works as expected.

PR #753, opened 2016-07-19
Dockerize

chaintng:
- Add Dockerfile for easy setup
- also add the doc how to use it.

regards to https://github.com/mxstbr/react-boilerplate/issues/751#issuecomment-233661187

PR #1022, opened 2016-09-23
Added Docker to deployment steps docs

amilajack:

### DIFF
--- a/docs/general/deployment.md
+++ b/docs/general/deployment.md
@@ -1,8 +1,6 @@
-# Deployment
+# Deploying React-Boilerplate
 
 ## Heroku
-
-### Easy 5-Step Deployment Process
 
 **Step 1:** Create a `Procfile` with the following line: `web: npm run start:prod`. We are doing this because heroku runs `npm run start` by default, so we need this setting to override the default run command. 
 
--- a/docs/general/deployment.md
+++ b/docs/general/deployment.md
@@ -4,16 +4,94 @@
 
 ### Easy 5-Step Deployment Process
 
-*Step 1:* Create a Procfile with the following line: `web: npm run start:prod`. We are doing this because heroku runs `npm run start` by default, so we need this setting to override the default run command. 
+**Step 1:** Create a `Procfile` with the following line: `web: npm run start:prod`. We are doing this because heroku runs `npm run start` by default, so we need this setting to override the default run command. 
 
-*Step 2:* Install heroku's buildpack on your heroku app by running the following command: `heroku buildpacks:set https://github.com/heroku/heroku-buildpack-nodejs#v90 -a [your app name]`. Make sure to replace `#v90` with whatever the latest buildpack is which you can [find here](https://github.com/heroku/heroku-buildpack-nodejs/releases).
+**Step 2:** Install heroku's buildpack on your heroku app by running the following command: `heroku buildpacks:set https://github.com/heroku/heroku-buildpack-nodejs#v90 -a [your app name]`. Make sure to replace `#v90` with whatever the latest buildpack is which you can [find here](https://github.com/heroku/heroku-buildpack-nodejs/releases).
 
-*Step 3:* Add this line to your Package.json file in the scripts area: `"postinstall": "npm run build:clean",`. This is because Heroku runs this as part of their build process (more of which you can [read about here](https://devcenter.heroku.com/articles/nodejs-support#build-behavior)).
+**Step 3:** Add this line to your Package.json file in the scripts area: `"postinstall": "npm run build:clean",`. This is because Heroku runs this as part of their build process (more of which you can [read about here](https://devcenter.heroku.com/articles/nodejs-support#build-behavior)).
 
-*Step 4:* Run `heroku config:set NPM_CONFIG_PRODUCTION=false` so that Heroku can compile the NPM Modules included in your devDependencies (since many of these packages are required for the build process).
+**Step 4:** Run `heroku config:set NPM_CONFIG_PRODUCTION=false` so that Heroku can compile the NPM Modules included in your devDependencies (since many of these packages are required for the build process).
 
-*Step 5:* Follow the standard Heroku deploy process at this point:
+**Step 5:** Follow the standard Heroku deploy process at this point:
 
 1. `git add .`
 2. `git commit -m 'Made some epic changes as per usual'`
 3. `git push heroku master`
+
+## Docker
+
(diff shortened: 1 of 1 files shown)


amilajack:
Heavily borrowing from https://github.com/mxstbr/react-boilerplate/pull/753. Followed these exact same steps and successfully built.

gihrig:
Beyond comment above, it looks pretty good. It would be nice to get someone who was involved in #753 to bless it.

I read it all and see no obvious errors. Can't vouch for technical though, not familiar with Docker.

amilajack:
@chaintan17 @kane-c @gornostal Thoughts?

chaintan17:
please see https://github.com/mxstbr/react-boilerplate/issues/751#issuecomment-249247723

amilajack:
@mxstbr @gihrig What would be preferred:

gihrig:
@amilajack or @chaintan17 Can the various 'docker files' be placed out of the root dir e.g. `internals/docker`?

If so, I would be happy to see them there. I'm somewhat disinclined to add three more files to the project root. This project has made a point to keep the root dir as clean as practical.

amilajack:
I'm pretty sure they can't be. @chaintan17 can you confirm this?

mxstbr:
> I'm somewhat disinclined to add three more files to the project root.

Thirded. If that's the case, I'd rather add a doc explaining how to deploy with Docker. If that's not a case, I'd love to have a `internals/docker` or something and a simple `npm run deploy` command that does a **production ready** deploy!

(Note: I have never used or tried Docker. This might be the most epic and easy **production ready** deployment method ever. IF THAT'S THE CASE (can somebody unbiased please try?) we might just add Docker support by default, but I doubt it)

amilajack:
From a user's perspective, docker is very similar to what Vagrant does. It **guarantees that processes will run as expected** regardless of a user's environment/OS/other differing configuration. It does this by running Linux Containers inside of a linux-based VM (this part isnt super crucial to understanding why its important).

The biggest pro of it is its guarantee that it will work anywhere.

chaintan17:
@amilajack @gihrig of course we can add those files in subdirectory.
Because we can setup `Dockerfile` path in automated test at http://cloud.docker.com
also, when build the image locally, developer then specified the part like `docker build -t myapp internals/docker`

For deploying, cc @mxstbr , i'm not certainly sure what are you expected from `npm run deploy`, could you please elaborate more ?

In my experience, i setup git repo and integrate with http://hub.docker.com and monitor new master branch push in github, then DockerHub with automatically create a new image from the new code changes. And if we have CI (e.g. Jenkins) setup in our infrastructure, we can automatically deploy to production using that newly build.

amilajack:
He means that he wants you to add an npm script that will make make docker run. Check the package.json's 'scripts' section for examples.

amilajack:
> of course we can add those files in subdirectory.

@chaintan17 thanks for the correction. Can you change your PR so that it reflects this?

chaintan17:
Yes, that could also possible by setting `docker run -p 3000:3000 internal/docker` in `package.json` and we now host our application inside docker container.

but `npm start` still sound easier for me in the sense that it has lower overhead cost.

mxstbr:
> of course we can add those files in subdirectory.

Ooooh let's see this, now I'm intrigued. Can you update the PR, then I'll check it out!

> Yes, that could also possible by setting docker run -p 3000:3000 internal/docker in package.json and we now host our application inside docker container.

`npm run deploy` would be upload-to-server-and-restart or something, contrary to `npm run start` which would be for development.

I guess the CI setup also makes sense, but not sure we want to enforce people to do Continuous Deployment just because they use `react-boilerplate`. Many people run manual deployments, and that's totally fine. I'd rather not enforce another thing…

gihrig:
> Ooooh let's see this, now I'm intrigued. Can you update the PR, then I'll check it out!

I have Docker on my short list of new tech to learn. This might just bump it up a notch or two 😉

chaintan17:
Ummmmm.... this is more complicated that what i initially thought...

I feel like, there are 2 use cases of using Docker in our project.
1. **For develop:** it help to make sure that each developer are working on the same environment as close to production env as possible.
2. **For deploy to production :** It will build current source code and pack them into docker image, and allow server to pull that image and run it live.

So, i separate 2 use cases into 2 dockerfile.
Each use case have their own docs, 
* `commands.md` for `development`
* `deployment.md` for `deploy to production`

Please review this pull request https://github.com/mxstbr/react-boilerplate/pull/753 and let me know what do you think.

Thank you so much :))

amilajack:
Having two dockerfile's would be fine if they are each in their own folder inside of internals.

gihrig:
@chaintan17 OK, so I get the use cases, and quite agree. But, can you tell us more about how Docker support would have to vary in each case?

My (naive) understanding is that development Docker would just be a local installation (probably on dev's computer), that is otherwise similar to production Docker. In that case, it seems the only difference would be the deployment target, local vs remote.

Please enlighten me.

amilajack:
The only difference that I can see between the development and production docker setups is the npm commands that we run on the repo. In development, we'll run `npm start` and in production we'll run `npm run build` or its equivalent.

chaintan17:
@gihrig actually you are right... using docker in development help developer able to edit the file in the container. To enable that, user need to specify `-v` parameter to map the volume from host directory to container directory. And override default command to `bash`
So, yes, actually, we can use same `Dockerfile`. I just update the pull request, please check.

Anyway, look like it's not working yet.
1. For development issue, i got error
```
ERROR in ./app/containers/App/banner-metal.jpg
Module build failed: Error: /reactapp/node_modules/jpegtran-bin/vendor/jpegtran: 1: /reactapp/node_modules/jpegtran-bin/vendor/jpegtran: Syntax error: "(" unexpected
```
Removing `import banner-metal.jpg` make the application working

2. for produciton issue, look like we have some problem with `airbnb-eslint-config`, please check this issue https://github.com/airbnb/javascript/issues/1088#issuecomment-249403495

chaintan17:
@amilajack there is some mistake in that pull request, it should have `VOLUME` stated, and correct `CMD` command for production. i updated the Dockerfile.

amilajack:
@chaintan17 Actually if you're going to PR for the `Dockerfile`'s, I don't think that this PR is useful anymore

chaintan17:
yes, let move the conversation to https://github.com/mxstbr/react-boilerplate/pull/753

PR #1023, opened 2016-09-23
Match travis node versions on appveyor

amilajack:


amilajack:
We just added travis ci support for node versions 4, 5, and 6. Makes sense to add them to appveyor as well.

mxstbr:
Seems reasonable, let's 🚢

PR #894, opened 2016-08-21
removing psi, pagespeed, and doc references

fisch0920:
Addresses #893

PR #1024, opened 2016-09-23
Add BrowserSync

sedubois:
Just noting that besides adding the basic browserSync functionalities, this switches tunnelling from ngrok to browserSync's built-in feature (uses `localtunnel`), whose behaviour might be slightly different. For example localtunnel doesn't work with my VPN, but OTOH it allows the tunnel to be nicely integrated within browserSync's dashboard, and the tunnel's URL can be configured for free.

If agreed and if wanting to clean up, ngrok could be removed altogether, which would require updating `pagespeed.js` (need to launch server, wait for browserSync init, then pass tunnel URL to PageSpeed Insights).

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -59,7 +59,7 @@
 
 ### Development
 
-Run `npm start` to see your app at `localhost:3000`
+Run `npm start` to see your app at `localhost:3000` and on your local network via BrowserSync, with hot reloading. 
 
 ### Building & Deploying
 
@@ -119,7 +119,7 @@ get a performance check right in your terminal!
 
 #### Browser testing
 
-`npm run start:tunnel` makes your locally-running app globally available on the web
+`npm run start:tunnel` makes your app available as with `npm start`, as well as on the web
 via a temporary URL: great for testing on different devices, client demos, etc!
 
 #### Unit testing
--- a/docs/general/commands.md
+++ b/docs/general/commands.md
@@ -19,7 +19,13 @@ history irreversibly by accident.
 npm run start
 ```
 
-Starts the development server running on `http://localhost:3000`
+Starts the development server and makes your application accessible with 
+[BrowserSync](https://www.browsersync.io/) at `localhost:3001` and on a local IP address (check your startup logs).
+
+This means that your changes will keep in sync on all your local devices, 
+even when you update the code or you scroll the page on one of them! 
+
+Make sure to check the BrowserSync dashboard at `localhost:3002`.
 
 ## Cleaning
 
@@ -53,9 +59,7 @@ generate container`)
 npm start
 ```
 
-Starts the development server and makes your application accessible at
-`localhost:3000`. Tunnels that server with `ngrok`, which means the website
-accessible anywhere! Changes in the application code will be hot-reloaded.
+Starts the development server and makes your hot-reloadable application accessible everywhere with BrowserSync.
 
 ### Production
 
@@ -141,8 +145,8 @@ Watches changes to your application and reruns tests whenever a file changes.
 ```Shell
 npm run start:tunnel
 ```
-Starts the development server and tunnels it with `ngrok`, making the website
-available on the entire world. Useful for testing on different devices in different locations!
+Starts the development server as usual for development, with an extra tunnels to make the website
+available to the entire world. Useful for testing on different devices in different locations!
 
 ### Performance testing
 
--- a/docs/testing/remote-testing.md
+++ b/docs/testing/remote-testing.md
@@ -4,9 +4,8 @@
 npm run start:tunnel
 ```
 
-This command will start a server and tunnel it with `ngrok`. You'll get a URL
-that looks a bit like this: `http://abcdef.ngrok.com`
+Starts the development server similarly to `npm start`, with an extra address XXX.localtunnel.me where XXX is 
+your app name configured in `package.json`.
 
-This URL will show the version of your application that's in the `build` folder,
-and it's accessible from the entire world! This is great for testing on different
-devices and from different locations!
+This means your app is accessible from the entire world! 
+This is great for testing on different devices and from different locations.
--- a/internals/config.js
+++ b/internals/config.js
@@ -21,11 +21,11 @@ const ReactBoilerplate = {
        * by listing them here.
        */
       exclude: [
+        'browser-sync',
         'chalk',
         'compression',
         'cross-env',
         'express',
-        'ip',
         'minimist',
         'sanitize.css',
       ],
--- a/internals/scripts/browserSync.js
+++ b/internals/scripts/browserSync.js
@@ -1,0 +1,35 @@
+const browserSync = require('browser-sync');
+const path = require('path');
+const pkg = require(path.resolve(process.cwd(), 'package.json'));
+
+const logger = require('../../server/logger');
+
+module.exports = function startBrowserSync(port, middleware, cb) {
+  const bsConfig = {
+    proxy: {
+      target: `localhost:${port}`,
+      middleware,
+    },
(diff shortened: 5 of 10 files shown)


amilajack:
I havent used browserSync much. What is the advantage/difference between it and react-hot-reloader?

sedubois:
Provides a nice dashboard, single-click to reload page on all devices, indicator when your pages get disconnected/reconnected from network, keep page scrolling in sync across all the browsers, etc. I suggest to try out to get a feel of it.

sedubois:
Also syncs clicks across devices, so you click a button on one of them and it does the same on all others. Same thing for form input.

sedubois:
Also visual CSS outlining, network throttling, remote debugging. Throttling kinda shows that this boilerplate isn't usable below 4G. Didn't try the remote debugging.

andykenward:
You arent running production build code when you run the `npm start` script you changed.

It is only worth using the network throttling when running production built code. This boilerplate is totally usable below 4G.

sedubois:
OK, good to know. With more work then maybe browserSync could be launched for either dev or prod.

gihrig:
>  ... browserSync's built-in feature (uses localtunnel), ... For example localtunnel doesn't work with my VPN ...

The main value I get from ngrok is the ability easy share work in development with clients and colleagues across the internet. I would not want to  lose that ability.

sedubois:
@gihrig `localtunnel` doesn't work for you?

gihrig:
@sedubois I never tied it. Your remark (quoted above) led me to believe that it would not.

Can you explain more about the differences between ngrok and localtunnel? 

Why/how does localtunnel not work with your VPN?

sedubois:
I have a Freedome VPN. With it enabled, localtunnel doesn't work whereas ngrok does. But I suspect not many people have such a setup? With VPN disabled, localtunnel works fine, I can access my machine from the internet.

OTOH as mentioned, localtunnel is integrated within browserSync's dashboard, and allows custom subdomains, which can be handy when sharing URLs with clients. In ngrok, this is a paid feature.

Note that I'm not fully familiar with either of these tunnel providers. I casually used ngrok in the past and recently discovered localtunnel as it is provided within browserSync.

amilajack:
Note that PR's should be made to the `dev` branch instead of master.

mxstbr:
> which would require updating pagespeed.js

Note: #894 will probably be merged, so this is probably not a concern at this point.

I have never personally used BrowserSync, so I'll have to play with it and see what it feels like. (not sure when I have the time, but I'll try to soon! Currently on vacation in 🗽) Question: Does this replace hot reloading?

amilajack:
Hot reloading is a more modern solution for updating the UI's during development. They can't save state. Instead, they reload the entire page.

mxstbr:
I think closing this was a bit premature, BrowserSync does have value.

Contrary to hot-reloading, it syncs button clicks and interactions across pages, which makes cross-device testing very easy and quick.

The question I have if those two techniques could be used alongside each other, or if they're completely separate?

sedubois:
> Does this replace hot reloading?

@mxstbr browserSync can be configured (as in this PR) to proxy webpack, the latter then handling hot reloading.

https://github.com/Browsersync/recipes/tree/master/recipes/webpack.react-hot-loader

mxstbr:
So, essentially, the only difference here to the current version is that if you open the webpage on multiple devices at the same time, all the interactions sync, do I have that right?

If so, that's pretty cool I think 😊

sedubois:
@mxstbr yes 😊 plus some nice-to-have tools like CSS outlining, remote debug, network throttling. And I switched tunnelling from ngrok to browserSync's, but that part isn't required, could be rolled back.

mxstbr:
I like the code I've seen (though I haven't done an in-depth review yet), the thing that's holding me back here is that I haven't tested this yet so I'm not sure how awesome this is.

I'm pretty busy at the moment, so if somebody who's never used BrowserSync could try this out and report back how useful it is that'd be amazing!

GGAlanSmithee:
> if somebody who's never used BrowserSync could try this out and report back how useful it is that'd be amazing!

I've never used browsersync before, so I took this opportunity to test it out. I ran the app in a couple of browser sessions, as well as on my Nexus phone and the syncing works great. The combination of browsersync with HMR makes for a great developer experience. Especially if you target phones, as you can navigate it remotely from your desktop.

_note: I am using [cloud9 ide](https://c9.io/), so I had to add [these lines](https://github.com/GGAlanSmithee/react-boilerplate/blob/browsersync/internals/scripts/browserSync.js#L14-L17) since c9 only allows for ports 8080, 8081 and 8082. With this setup you get HMR at 8080, browsersync+HMR at 8081 and the browsersync dashboard at 8082._

sedubois:
@GGAlanSmithee thanks, I updated the code to set up ports correctly even when changing the default. Feel free to try again.

GGAlanSmithee:
@sedubois tested again with your changes and got error

`RangeError: "port" option should be >= 0 and < 65536: 80801`

this happends because `port` is a string, so `port+1` performs a string concatenation rather than an addition of integers. `parseInt(port, 10) + 1` and `parseInt(port, 10) + 2` fixes this.

Other than that, this is working as expected 👍

sedubois:
Thanks @GGAlanSmithee, I fixed again :)

bsr203:
About hot reloading, please read this comment https://github.com/gaearon/react-hot-boilerplate/issues/97#issuecomment-249862775

and the priorities are 

```
Here is my current plan about adding hot reloading to React:

 Make Create React always refresh the page on change so people learn to trust it and stop pressing “refresh” themselves.
```

I came here as well to switching back from hot reloading to browser-sync.

GGAlanSmithee:
@bsr203 thanks for linking to that issue, that is interesting indeed. Hopefully they will integrate this with react some time in the future and hide it under a development flag, much like propTypes checks are done.

Do you know if browser-sync can fully replace hmr?

bsr203:
@GGAlanSmithee to be honest, I am planning to move away from `react-router`, as [this library](https://github.com/grahammendick/navigation) works great with relay and migration was quite smooth. The coe is clean, and minimal.

A talk by the author: https://www.youtube.com/watch?v=Pj-RakjfHDI#t=33m15s

They also have a [hot reloading example](https://github.com/grahammendick/navigation/tree/master/NavigationReact/sample/hotloader). gonna try it today.

sedubois:
There's an issue I didn't notice before, when running `build:dll`:

```
ERROR in ./~/browser-sync/index.js
Module parse failed: ./~/browser-sync/index.js Unexpected character '#' (1:0)
You may need an appropriate loader to handle this file type.
| #! /usr/bin/env node
| "use strict";
| 
 @ dll reactBoilerplateDeps
```

I don't really know why it's parsing this file as it's in `node_modules`?

sedubois:
So AFAIK, this browserSync works quite great with the exception of keyboard events. E.g in the example app there's a [`preventDefault`](https://github.com/mxstbr/react-boilerplate/blob/master/app/containers/HomePage/index.js#L151) which might be why it doesn't work. There's an issue about that as I described [here](https://github.com/BrowserSync/browser-sync/issues/1232#issuecomment-254479925), but no progress since 1.5 years.

sedubois:
Just got a [reply](https://github.com/BrowserSync/browser-sync/issues/1232#issuecomment-254520132) from BrowserSync: they intend to improve keyboard syncing (but no time frame).

mxstbr:
To be honest, I don't think I really want to merge something by default that doesn't even work for the tiny example app. That being said, I definitely see the value of BrowserSync, so when things improve on this front I'd love to add it in the future!

Thanks for this great reference PR, sorry for the slow response here!

PR #1035, opened 2016-09-25
Updated readme with quick start requirements

### LINKED ISSUE #1034: Update README with quickstart prereqs
quick start will fail if the minimum pre requirements are not met. (node v4+ and npm v3+)
These are currently missing from quick start steps.

toddsby:
resolves #1034


mxstbr:
If #1034 is a bug that needs to be resolved, this is not the right way to do it. Encouraging people to just randomly update their npm might break all their other setups and projects, which is not something we want to be responsible for at all.

A better idea would be to list the required versions in cursive above the setup instructions, but let's wait for a response in #1034 first to see if this is a bug at all, I'm not 100% convinced.

PR #1037, opened 2016-09-26
Split server into frontend and backend

sedubois:
Proposal to make it easier to start building backend things on top. E.g graphQL, etc could be added into `server/backend` instead of cluttering `server/index.js` more.


mxstbr:
I like it! 👍 Makes things a lot clearer.

peter-mouland:
Can I put my 2cents in? 

I see you have this concept in the code already, and this is an improvement to that.  

For a newbie though, the idea of server frontend/backend is a bit weird as the server is only backend.  In this context though it looks like you are talking about routers.  So i'd just have a `server/api` and `server/routes` or something similar

sedubois:
Hmm I'm not entirely familiar with web parlance, but in my understanding, server and backend are different things. The server (express in our case) "serves" both the code which will run both remotely in the browser client (frontend) and the code which will run within the server itself (backend).

I wouldn't talk about routes either, because not all servers have routes (e.g GraphQL).

peter-mouland:
fair enough - sounds like we're thinking about the concepts from completely different viewpoints! 

Once I read through the code more thoroughly I do see the distinction (serving the backend api and serving the frontend assets), but it was just that extra bit of effort which can sometimes be avoided completely.  

no worries, as I say, just my 2cents :)

mxstbr:
What about calling it `app` and `api` instead of `frontend` and `backend`? I think that'd make the distinction very clear!

sedubois:
First, the term API is quite vague: whose API? "Backend API" would be more correct (in full words: "the application programming interface of the backend").

Second, personally I don't see why using a term as `api` in the code: the API you refer to is just the _entry point_ to access the backend code, it's not _all_ of the backend code (i.e it would be incorrect to say that persistence-related code, or business-logic code, is part of an API).

IMHO

mxstbr:
While I agree with you, I don't think having a folder called `server/api/` is confusing, it's pretty clear that that's the backend API, or not?

sedubois:
Well, the server also serves the front-end code. Server and back-end are not the same thing. 

To make a digression, it's true that we abuse the language anyway with the word _server_ itself. Technically server is short-hand for HTTP server, but actually what we put under that directory is the Javascript code called by the HTTP server.

To come back to the term _API_, `backend/api` would be more accurate, then devs could add siblings `backend/services` and `backend/persistence` as needed.

IMHO 😄 I'm probably just pedantic. Naming things properly is one of the hard things in software dev.

But feel free to just close the issue, I don't need this any more anyway. And then feel free to rename things as you want.

mxstbr:
> But feel free to just close the issue, I don't need this any more anyway. And then feel free to rename things as you want.

No, I actually like this change, I'm just bikeshedding over the naming! Maybe you're right, I'm not a backend dev. I'll have a think about it, and if I can't come up with anything better I'll merge it tomorrow.

sedubois:
> I'm just bikeshedding over the naming

kudos for teaching me new vocab 😉

gihrig:
I _do_ come from a backend (Java/Spring) environment. IMHO app/api is fine.

May not be the most descriptive, but not hard to figure out looking at the code, could (should) even be documented.

One of things that really annoyed me in the Java/Spring world was the extremely verbose path and naming conventions:

`/Project/src/main/java/com/project/project/web/ApplicationConversionServiceFactoryBean.java`

Yes, that's from a real, live on the web project (Project is not the real name).

tandrewnichols:
Keep in mind that tab completion will be slightly less useful if you choose `app` and `api`.

samit4me:
Hi @sedubois, this branch looks strikingly similar to #1556. Would you be able to take a quick look at that PR and provide any feedback? Im going to close this PR in favour of the other only because this branch getting stale.

xavifuefer:
Hi @sedubois @samit4me I don't understand why would `server/index.js` use the same process for both the backend and the frontend server? why not 2 different express apps?

samit4me:
@xavifuefer To my understanding, this server is purely an application server, so I am unsure on the need to have two servers running. They would need to have different ports.

gihrig:
As discussed in this [#194 Comment](https://github.com/react-boilerplate/react-boilerplate/pull/194#issuecomment-192675025), the currently provided express server: 

>solely exist to make npm run serve work and tunnels your local connection with ngrok.

PR #1049, opened 2016-09-28
Don't error out if the substate does not exist

roderik:
In a clean install, and a newly generated component, the following error would be thrown:

> Uncaught (in promise) TypeError: Cannot read property 'toJS' of undefined

This is because there is no substate since the redux store does not contain anything yet.
This fix returns an empty object if the substate does not exists.


julianorafael:
The error happens when you run `npm run generate component` but do not run `npm run generate route`.

roderik:
Indeed!

mxstbr:
I don't understand, how are `generate component`, `generate route` and the template for the selector for _the container_ connected?

julianorafael:
The error is happening because the selector can't find the state. The state need to be passed as argument when creating the selector on mapStateToProps.

mxstbr:
I'm even more confused now. Do we need this or not?

julianorafael:
The error happens because the generators create all the files for the container but don't add the container's reducer to the reducers file, so the container's selector can't get the piece of the state related to it because it doesn't exist.
I believe we need to make the reason clear so people can quickly understand it and solve it accordingly. A couple of comments should do the trick.

KarandikarMihir:
Is this PR still relevant? V3.4 has a lot of changes.

gihrig:
@KarandikarMihir It may be relevant, but it first must be rebased against `dev` I just did that, and updated against current `dev`, did that too.

If the code survives those actions, someone should test to confirm that the OP condition still exists and that the proposed changes correct the issues.

Assuming the above pass, the community will need to decide if the error should be suppressed as in the OP solution, or documented, or both.

I vote for both, documentation in the form of a code comment.

gihrig:
> The error happens because the generators create all the files for the container but don't add the container's reducer to the reducers file, so the container's selector can't get the piece of the state related to it because it doesn't exist.

I think #1523 is addressing the same problem, but in a more informative manner.

Unless I'm mistaken, I vote for #1523.

samit4me:
I agree with @gihrig #1523 looks to have this covered. It looks like no one objects, so closing this PR in favor of #1523 Thank you for raising the issue @roderik and getting the ball rolling ✨

PR #1065, opened 2016-10-01
Demo GraphQL Relay client

sedubois:
NB: doesn't use GitHub GraphQL API because was [down](https://platform.github.community/t/graphql-api-endpoint-experiencing-downtime/582) this weekend, and was nice opportunity to experiment with [graph.cool](https://graph.cool/).
- [ ] improve UX
- [ ] test coverage
- [ ] doc

screen shot 2016-10-03 at 18 26 04

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ stats.json
 .DS_Store
 npm-debug.log
 .idea
+
+# https://github.com/jimkyndemeyer/js-graphql-intellij-plugin
+graphql.config.json
--- a/app/app.js
+++ b/app/app.js
@@ -24,6 +24,8 @@ import { syncHistoryWithStore } from 'react-router-redux';
 import FontFaceObserver from 'fontfaceobserver';
 import { useScroll } from 'react-router-scroll';
 import configureStore from './store';
+import useRelay from 'react-router-relay';
+import Relay from './relay';
 
 // Import Language Provider
 import LanguageProvider from 'containers/LanguageProvider';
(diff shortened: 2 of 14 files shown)


sedubois:
Feedback would be great, to know if this should be readied for merge 😊 

My main concern left right now is whether to choose Relay or Apollo. But they have similar functionalities and they will evolve in next weeks/months anyway (especially when Relay 2 is out). So I'd say going with Relay is good enough in the scope of this PR.

alisalaah:
Apollo seems the better choice in relation to Redux concerns and features. Why would we want to use Relay instead of Apollo?

sedubois:
Unfortunately Relay 1 is quite large within the final bundle:

<div align="center">
screen shot 2016-10-24 at 20 12 59
</div>

GGAlanSmithee:
@sedubois How does that compare to Apollo?

I'm looking at both Apollo and relay atm and must say that Apollo is light years easier to grok coming from a react, redux, router background.

I really think they struck the right abstraction level and everything feel familiar, the way you use it as a redux middleware, can observe requests in redux devtools, their HOC etc. Relay OTH feels like magic. Just my two opinionated cents :)

sedubois:
@GGAlanSmithee I haven't put RBP + Apollo in a working state (work so far was [here](https://github.com/apollostack/react-apollo/issues/267#issuecomment-253453743)), feel free to check and report 😄

GGAlanSmithee:
@sedubois ah ok, I have apollo working client side in a small project, if I get the time I'll see if there's anything I can contribute to your example.

DenisNeustroev:
@sedubois I think you should pass the immutable prop to ApolloProvider 
```jsx
<ApolloProvider immutable store{...} client={...}>
```
 and change reduxRootSelector to 
```javascript
reduxRootSelector: (state) => state.get('apollo')
```

PR #1067, opened 2016-10-02
ESLint config airbnb v12 migration

amilajack:


amilajack:
@gihrig @mxstbr What do you guys think?

gihrig:
Put it through its paces, LGTM 💯

PR #1072, opened 2016-10-04
Extract header

sedubois:
By symmetry with the footer and to make it easier to add new features in it.


mxstbr:
I like this, let's shipit!

PR #1075, opened 2016-10-05
Issue 1073 - Warn when js is disabled

Vadorequest:
## Add <noscript> message in index.html and its template.

Add a <noscript> tag in the index.html file and its template, following @gihrig 's comment https://github.com/mxstbr/react-boilerplate/issues/1073#issuecomment-251562561

I don't know if this should be tested, nor how. I believe it's not useful.
I don't know when the template file is used, maybe when using `npm run generator`?
I also fixed indentation in the template file, let me know if it shouldn't be done.

Merge is done in **dev** branch, let me know if it should be merged directly into **master**.

**Status:**

- [ ] ~~Pull request has tests (we are going for 100% coverage!)~~
- [ ] ~~Documentation is updated (if necessary)~~

### Results

image
---
image


gihrig:
Overall it looks great 😄 

There is one issue. The warning text appears at the bottom of the page, below the bottom of the screen and out out of sight in a production build: `npm run start:production`. 

This was observed in Chrome, Safari and Firefox (all latest on OS X 10.11.6).

This is due to `min-height: 100%` on the containing `<div id="app">`. 

I seem to recall that `<noscript>` is expected at the end of `<body>`. You might look into that. Placing `<noscript>` before `<div id-"app">` Should fix this.

Still I don't understand why this is a problem only in a production build. It needs more research...

Vadorequest:
Thanks for the quick feedback!

I didn't try in production mode, but it seems weird indeed.
I'll move the `<noscript>` before the app then and see what happens.

Vadorequest:
I took a quick look at the DOM and here are some difference that may explain why it's not the same behavior between dev and prod envs.

**Prod**
2016-10-05 23_56_35-greenshot


Some styling is applied.

**Dev**
2016-10-05 23_56_57-greenshot

`min-height` isn't applied.

---
Anyway, I'm just gonna move the noscript tag for now.

Vadorequest:
With the noscript tag on top, it works fine in production mode. I guess we can squash & merge (or just merge if you don't care about merging 3 commits) 

let me know

gihrig:
Yes, it works as expected now - Thanks for contributing 🎉 

I say let's merge as is, but because it's his project, I would like to give @mxstbr a chance to comment before merging.

mxstbr:
This looks great to me, thank you very much!

PR #1080, opened 2016-10-06
Global CSS vars color palette

sedubois:
I didn't change colors (only one slightly to simplify things), just put them into global variables to be able to easily edit colors globally. Works with HMR too.

### DIFF
--- a/app/components/A/styles.css
+++ b/app/components/A/styles.css
@@ -1,7 +1,9 @@
+@import '../../containers/App/colors.css';
+
 .link {
-  color: #41ADDD;
+  color: var(--color-accent);
 }
 
 .link:hover {
-  color: #6CC0E5;
+  color: var(--color-accent-light);
 }
--- a/app/components/Button/styles.css
+++ b/app/components/Button/styles.css
@@ -1,3 +1,5 @@
+@import '../../containers/App/colors.css';
+
 .buttonWrapper {
   width: 100%;
   text-align: center;
@@ -19,11 +21,11 @@
   font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
   font-weight: bold;
   font-size: 16px;
-  border: 2px solid #41ADDD;
-  color: #41ADDD;
+  border: 2px solid var(--color-accent);
+  color: var(--color-accent);
 }
 
 .button:active {
-  background: #41ADDD;
-  color: #FFF;
+  background: var(--color-accent);
+  color: var(--color-secondary-light);
 }
--- a/app/components/Footer/styles.css
+++ b/app/components/Footer/styles.css
@@ -1,6 +1,8 @@
+@import '../../containers/App/colors.css';
+
 .footer {
   display: flex;
   justify-content: space-between;
   padding: 3em 0;
-  border-top: 1px solid #666;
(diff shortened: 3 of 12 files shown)


GGAlanSmithee:
@sedubois might I suggest naming the variables to something more descriptive than `color{n}`

sedubois:
@GGAlanSmithee What would you suggest? I don't know what the best practices are, but the point of these vars is to be able to change colors easily. So it hardly helps to name them `--color-blue`, etc. 

NB: I intentionally ranked them from darker to lighter, though.

GGAlanSmithee:
@sedubois I have no idea about CSS variable best practices either, and I agree that naming them after their respective color is somewhat pointless. My suggestion would be to name them after their intended use. Maybe something like `primary`, `secondary`, `accent`, or if you'd like to be more specific, `prmiary-indigo`, `accent-pink` etc. What do you think?

_inspired by https://material.google.com/style/color.html#color-color-schemes_

sedubois:
Thanks for the link. This is beyond my expertise, but how about this?

mydearxym:
is there a possibility to switch these color with javascript ?  that would make theme staff much easier ?

sedubois:
@mydearxym could you precise your question?

GGAlanSmithee:
Just like @sedubois my familiarity to CSS vars is limited, but what you want seems to be possible @mydearxym https://eager.io/blog/communicating-between-javascript-and-css-with-css-variables/

mydearxym:
what i mean is if we manipulate those color with javascript ?

for example before: 

```js
:root {
  --color-primary: black;
  --color-primary-light: #CCC;
  --color-secondary: #FAFAFA;
  --color-secondary-light: #FFF;
  --color-accent: #41ADDD;
  --color-accent-light: #6CC0E5;
}
```

after some js magic : 

```js
:root {
  --color-primary: blue;
  --color-primary-light: #CCC;
  --color-secondary: #FAFAFA;
  --color-secondary-light: #FFF;
  --color-accent: #41ADDD;
  --color-accent-light: #6CC0E5;
}
```

it will be very useful to add some theme staff

sedubois:
@mydearxym that should be possible somehow, e.g looking at https://github.com/este/este/blob/master/src/common/app/components/ThemeProvider.js. But that's out of scope of this PR so if needed, suggest to open separate issue.

mydearxym:
@GGAlanSmithee  that solution is great, thanks!  but IE seems [not support it](http://caniuse.com/#feat=css-variables)

@sedubois  thanks, i will open a separate issue to discuss it.

mxstbr:
Since we're going to switch to styled-components, this is no longer necessary – thanks for taking the time to do this though, it would've been much better!

kopax:
@GGAlanSmithee thanks for the link. So I have read and yes, I suppose the way of the react boilerplate creator they suggest to go is to create independant templated component.

Does that mean we give up with the global styling ?

mxstbr:
> Does that mean we give up with the global styling ? no more global variable etc... ? 

No, `styled-components` can still inject global styling with the `injectGlobal` helper. Also, global variables are now just JavaScript variables instead of relying on the wrong pre-processing of a future CSS feature!

Issue #1074, opened 2016-10-04
Issue with local caching - Add a warning about it?


Vadorequest:
# Issue with local caching

## Issue Type

- [ ] Feature: Add a warning/message about local caching (FAQ?) due to "offline-first" feature.

## Description

I ran into something very weird today. Basically, I didn't get the same output when hitting F5 and ctrl+F5. F5 showed the old website, and ctrl+F5 showed me the latest -correct- version. I couldn't figure out why, local caching was disabled through the browser console. It's only when @gihrig gave me this solution that I finally figured out (or think so) what was wrong:

`Chrome dev tools > Application > Clear Storage > Clear site data`

I think that because of the "offline-first" feature, *sometimes*, the local caching gets screwed. And the only way to fix it is to clear it.

Basically, a simple warning in the docs somewhere would be enough, I don't really know what's the best place for it (maybe when we talk about offline-first? Assuming it's indeed related to offline-first.).

## Steps to reproduce

Couldn't really reproduce this error, happened once, hasn't happened again since.

# Versions

- Node/NPM: 6.7.0/3.10.3
- Browser:  53.0.2785.143 m (64-bit)

gihrig:
To reproduce (just a guess) :
In new clone with Example app code still in working dir:
`npm run start:production`
Open in Chrome browser.
Stop server
_now example app and its service worker are loaded in browser_

`npm run clean`
`npm start`
Open in Chrome browser
_Loading from server will get cleaned app. Loading from offline-first will get example app._

Also see #852 especially this [comment](https://github.com/mxstbr/react-boilerplate/issues/852#issuecomment-239920419).

Vadorequest:
Yep, this seems to be exactly what I have done! :)

I guess we should simply add it to the react-boilerplate FAQ, because people will likely run into this. If it's related to the production mode, maybe that's where it should be stated.

gihrig:
@Vadorequest see [#1075 comment](https://github.com/mxstbr/react-boilerplate/issues/852#issuecomment-251825900)

gihrig:
Discussion moved to #852 - Closing this.

gihrig:
See updated [FAQ](https://github.com/mxstbr/react-boilerplate/blob/dev/docs/general/faq.md#issue-with-local-caching-when-running-in-production-mode-f5--ctrlf5--cmdr-weird-behavior)

GGAlanSmithee:
A couple of questions, if you don't mind :)

[...]

PR #1081, opened 2016-10-06
FIX 852, 1074 - Add doc about local caching issue in the FAQ.

Vadorequest:
## Add doc about local caching issue in the FAQ

Add documentation to the FAQ about the F5/Ctrl+F5 AKA local caching issue.
See https://github.com/mxstbr/react-boilerplate/issues/852#issuecomment-252037351

### DIFF
--- a/docs/general/faq.md
+++ b/docs/general/faq.md
@@ -56,11 +56,11 @@ The fix is to kill the process and rerun `npm start`.
     > e.g. given the output from the example above, `YOUR_PID` is `28344`, hence
     that would mean you would run `taskkill /F /PID 28344`
 
-## Issue with local caching when running in production mode (F5 / ctrl+F5 weird behavior)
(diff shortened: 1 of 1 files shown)


GGAlanSmithee:
A couple of question, if you don't mind :)

Is this only a problem when running locally, or would you get the same problem when deploying under the same circumstances? Also, is the cache-related problem only present when updating during the same browser session?

Thanks!

Vadorequest:
@gihrig Done! Btw you can edit directly the file from github, faster than asking for review when it's just typos ;)

gihrig:
>  Btw you can edit directly the file from github, faster than asking for review when it's just typos ;)
@Vadorequest Didn't know that, thanks!

gihrig:
@GGAlanSmithee These are great questions, but I would like to have them in #1074 where others are more likely to find them when searching.

Please add to #1074, just copy and paste.

gihrig:
Great job @Vadorequest Thanks! 🎉

PR #1084, opened 2016-10-09
remove unnecessary `fromJS` call

nktssh:
`fromJS` will convert deep structures (object to Map, array to List). Since we already have one `fromJS` call - we need no second call


gihrig:
Please open all pull requests against `dev` branch.

mxstbr:
Thank you!

Issue #1090, opened 2016-10-12
PR/Discussion: Yarn Support


amilajack:
There are a large number of install issues related to the installation of npm packages. Yarn claims to guarantee installation safety:

> Yarn is able to guarantee that an install that worked on one system will work exactly the same way on any other system.

Think we should give this a go by recommending it as an alternative installation method and committing a `yarn.lock` file.

EugeneHlushko:
Hi @amilajack, i just opened a pull #1092

GGAlanSmithee:
While yarn looks promising, is it not a bit early to recommend it _over_ npm?

Watching the yarn repo, there were about 100 updated issues in the time frame of an hour. Some bug reports, some feature requests, some PRs. 

Maybe change the wording in #1092 from;

> We recommend to use yarn for dependency management

to

> As an alternative to `npm`, you can use `yarn` for dependency management, which might make the installation process easier for different platforms

And in time, when yarn is a bit more mature the wording can be changed?

EugeneHlushko:
Hi @GGAlanSmithee, reword sounds very reasonable lets move this discussion into PR, i will update the wording accordingly.

amilajack:
> And in time, when yarn is a bit more mature the wording can be changed?

Great idea, I definitely agree.

andykenward:
Does yarn work on travis-ci? 

I understand that it is faster on your local and machine and yarn produces a `yarn.lock` file so all the dependency version are the same even the ones from other packages. But if you are building and deploying via travis-ci etc and it doesnt use yarn and the `yarn.lock` file it seems pointless to use yarn at the moment.

sedubois:
@andykenward https://github.com/yarnpkg/yarn/issues/920

amilajack:
@andykenward see  https://twitter.com/kentcdodds/status/786282951210262528

andykenward:
So if `yarn` is used then the `travis.yml` needs to be amended? Or perhaps the docs need this defined in them for #1092

amilajack:
It isnt necessary. For now we should generate and commit both a `yarn.lock` and `npm-shrinkwrap.json` files for safer installation so that both npm and yarn users can benefit. We can get significantly faster build times by using yarn but for now I think we should wait for yarn to stabilize.

gihrig:
[...]

PR #1092, opened 2016-10-12
introduce yarn

EugeneHlushko:
Hi @mxstbr, lets add yarn.lock and yarn setup option if you don't mind.


README.md
```diff
@@ -98,7 +98,8 @@ There’s also a <a href="https://vimeo.com/168648012">fantastic video</a> on ho
 ## Quick start
 
 1. Clone this repo using `git clone --depth=1 https://github.com/mxstbr/react-boilerplate.git`
-1. Run `npm run setup` to install dependencies and clean the git repo.<br />
+1. We recommend to use yarn for dependency management, do a `yarn install` before doing `npm run setup-yarn`. Check out <a href="https://yarnpkg.com/en/docs/install" target="_blank">here</a> how to install yarn on your OS.
```
sedubois: Can just run `yarn` instead of `yarn install`

sedubois:
And the same way `package.json` is under VCS, I guess `yarn.lock` should be too.

README.md
GGAlanSmithee: added a comment in #1090

EugeneHlushko: Hi @sedubois yes you can, but its more obvious for beginners of what it does with npm syntax don't you think so?

sedubois: @EugeneHlushko then just add an explanation of what it does

EugeneHlushko: Do you think the new wording solves this? See the next commit please

EugeneHlushko: Okay lets hear from more people and see how it goes, i've tested yarn of few projects in the morning without any issues and very happy with it! :)

sedubois:
NB: target branch should be `dev`.

package.json
```diff
@@ -30,6 +30,7 @@
     "pagespeed": "node ./internals/scripts/pagespeed.js",
     "presetup": "npm i chalk shelljs",
```
EugeneHlushko: Not necessarily, yarn can be installed with `npm i -g yarn` as per instructions in the readme file (we have a link there to latest install instructions for different OS).

kane-c: I mean for installing chalk and shelljs :)

EugeneHlushko: Oh ye, good point. Pushed a new commit

README.md
```diff
@@ -98,7 +98,8 @@ There’s also a <a href="https://vimeo.com/168648012">fantastic video</a> on ho
 ## Quick start
 
 1. Clone this repo using `git clone --depth=1 https://github.com/mxstbr/react-boilerplate.git`
-1. Run `npm run setup` to install dependencies and clean the git repo.<br />
+1. As an alternative to npm, you can use yarn for dependency management, which might make the installation process easier for different platforms to do so: `USE_YARN=true npm run setup`. If you dont have `yarn` installed check out <a href="https://yarnpkg.com/en/docs/install" target="_blank">here</a> how to install yarn on your OS or do a `npm i -g yarn`
```
sedubois: typo: "dont"

EugeneHlushko: Thanks @sedubois

sedubois: I'd also recommend to squash all your commits, it's not very clean to have 9 commits for 7 lines of code. The current commits can be squashed with `git rebase -i HEAD~9` and next commits you can `git commit --amend` and `git push --force`.

EugeneHlushko: @sedubois thanks for the squash tip man but github has added a squash feature for the person who merges PR into the target branch within one click :) Since then i prefer not to squash commits so Peer Reviewer can decide himself which message to set and which commits to squash

README.md
```diff
@@ -98,7 +98,8 @@ There’s also a <a href="https://vimeo.com/168648012">fantastic video</a> on ho
 ## Quick start
 
 1. Clone this repo using `git clone --depth=1 https://github.com/mxstbr/react-boilerplate.git`
-1. Run `npm run setup` to install dependencies and clean the git repo.<br />
+1. As an alternative to npm, you can use yarn for dependency management, which might make the installation process easier for different platforms to do so: `USE_YARN=true npm run setup`. If you don't have `yarn` installed check out <a href="https://yarnpkg.com/en/docs/install" target="_blank">here</a> how to install yarn on your OS or do a `npm i -g yarn`
+1. (Skip this if you have done setup with `yarn`). Run `npm run setup` to install dependencies and clean the git repo.<br />
```
mxstbr: Instead of talking about this long and wide, what about detecting if people have yarn installed in the setup script and then just using it if they do? Why should they need to manually do `USE_YARN=true`? (which, btw, only works on unix based systems, on Windows it's different)

EugeneHlushko: Hi @mxstbr, detecting is a good idea, but don't you think user should have an option to force npm/yarn installation?

mxstbr: Sure, good point, let's leave that in there. It should be a tiny sidenote at the end of the instructions that people are able to do `USE_YARN`. It's not a very _quick_ start if you have to read a novel beforehand 😉

EugeneHlushko:
@mxstbr, this version looks cleaner. Tell me if you want me to rebase this into 1 commit, i don't do this anymore by default since you can do it while merging to develop as peer reviewer.

mxstbr:
I can squash merge and that's totally fine. I still don't see any automatic detection of yarn though? I thought we were going to use yarn if it's installed, but still let people optionally opt-out with `USE_YARN=false`?

EugeneHlushko:
I understood this vice-versa, will change that in next commit 👍

EugeneHlushko:
@mxstbr, updated version is here now!

internals/scripts/setup.js
```diff
@@ -68,6 +52,33 @@ function deleteFileInCurrentDir(file, callback) {
 /**
  * Installs dependencies
  */
-function installDeps(callback) {
-  exec('npm install', addCheckMark.bind(null, callback));
+function installDeps() {
+  exec('yarn --version', function (err, stdout, stderr) {
+    if (parseFloat(stdout) < 0.15 || err || process.env.USE_YARN === 'false') {
```
mxstbr: What is this for: `parseFloat(stdout) < 0.15`? Can we add a comment above stating why we don't want versions lower than 0.15?

EugeneHlushko: Morning @mxstbr 
Yarn from now on by default will be installed as 0.15 and above and all the tests that i have done were done with this version. With previous versions there might be unexpected behaviors. Do you think a comment like this will do?
`we expect yarn ^0.15 because previous versions were not tested and released before open sourcing yarn`

sedubois:
yarn 0.16.0 got released. Shall we merge this? IMHO no need for version checks, seems out of scope for this boilerplate.

EugeneHlushko:
@sedubois yarn finally accepted a pull to fix git+ssh links with private repos in new release :tada: 
a version check can remain if everyone agree, we can reuse this later when(if) version check is going to be needed.

mxstbr:
Let's get rid of the version check for now and add one when somebody reports a bug. (which I doubt)

We'll need one eventually, since they're going to adapt the lock file format which is a breaking change, but let's do that then and not now!

amilajack:
We need to find a way of updating the lock file when npm users install dependencies. Any recommendations?

sedubois:
@amilajack AFAIK if people want to use NPM, they should simply understand they're not using the benefits of the yarn.lock file (as has been the case so far), so their results won't be as guaranteed (can't predict precise dep versions). Maybe simply adding a disclaimer in the docs?

PR #1095, opened 2016-10-12
Remove warning about offline-plugin in dev

sedubois:
See https://github.com/mxstbr/react-boilerplate/issues/1025.


jeremyadavis:
FYI, this fix goes away on `npm run clean` because it isn't in the app.js template

Issue #888, opened 2016-08-19
Code Splitting/Async route issue with query string for same route


just4fun:
Hi there,

In the example app, I can search repos for GitHub user. But in my case, I want to share the search result to another person, so query string is needed to indicate specific result.

For example, when I input my GitHub account and press enter in home page, I want the url will be `http://localhost:3000?user=just4fun`. In this way, React Router will notice route changed, which will leads async route for home route be triggered (it has already been triggered when we access home page). And worse still, our sagas will be injected again.

```javascript
path: '/',
name: 'home',
getComponent(nextState, cb) {
  const importModules = Promise.all([
    System.import('containers/HomePage/reducer'),
    System.import('containers/HomePage/sagas'),
    System.import('containers/HomePage'),
  ]);

  const renderRoute = loadModule(cb);

  importModules.then(([reducer, sagas, component]) => {
    injectReducer('home', reducer.default);
    injectSagas(sagas.default);

    renderRoute(component);
  });

  importModules.catch(errorLoading);
},
```

Another side, as for the sagas,

```javascript
export function* githubData() {
  // Fork watcher so we can continue execution
  const watcher = yield fork(getReposWatcher);

  // Suspend execution until location changes
  yield take(LOCATION_CHANGE);
  yield cancel(watcher);
}
```

When url updated with query string, `LOCATION_CHANGE` action will be dispatched, it will leads the `watcher` be cancelled.

In conclusion, when I press enter, I did

- update url with query string for enter key handler
- get query string in `mapStateToProps`, then dispatch `loadRepos` action with query string in `componentWillMount` (case: input url directly) and `componentWillReceiveProps` (case: common search)

but nothing happen, b/c

- sagas has been injected again
- watcher in previous sagas has been cancelled

Is there any way to handle this case? Or is there any step I did not in proper way?

justingreenberg:
a simple solution would be to filter location actions by passing function to `take` effect, ie...

```js
export function* githubData() {
  // Fork watcher so we can continue execution
  const watcher = yield fork(getReposWatcher);

  const locationFilter = (action) =>
    action.type === LOCATION_CHANGE && !action.payload.query.user

  // Suspend execution until location changes * without user query *
  yield take(locationFilter);
  yield cancel(watcher);
}
```

now `LOCATION_CHANGE` actions will not cancel watcher:

```js
// eg in homepage component
this.props.changeRoute({ query: { user: 'justin' } })
```

[...]

Issue #826, opened 2016-08-04
Switch pages making sagas repeat inject


rangerch:
importModules.then(([reducer, sagas, component]) => {
                        injectReducer('demoPage', reducer.default);
                        injectSagas(sagas.default);
                        initedStatus.demoPage = true;
                        renderRoute(component);
                });

this code will repeat processing when switch page.

mxstbr:
Yes sir, this is intentional. Sagas should be using the watcher pattern as detailled in the example app to make sure they start and stop based on the route they're on!

@justingreenberg we should add a note of this to the docs, this question comes up really frequently.

skosno:
@mxstbr Is there any reason not to wrap sagas in "injectSagas" with this logic by default? Do you expect that some exported sagas should not cancel on location change?

mxstbr:
> Do you expect that some exported sagas should not cancel on location change?

Yes. This way, you can opt-into it, whereas the other way you're forced to have sagas only run when the route is open.

skosno:
Do you expect that there will be sagas that are injected asynchronously with routing but won't be suspended when routing changes (I was just trying to get real world scenarios)?

justingreenberg:
@skosno app-wise services, ie notifications/toast system, global fetch utilities, websocket manager, network connection status, etc

skosno:
@justingreenberg wouldn't we load such sagas directly upon application boostrap instead of injecting them as async sagas? If I understand the logic here, we want to load sagas connected to particular route asynchronously and stop their execution when route changes. What I am unsure is if there would be sagas that are not loaded on application bootstrap but rather upon reaching particular route and are not suspended when that route changes.

justingreenberg:
a few (convoluted) examples: 
- tracking background service (ie. data importer) while letting the user continue using app
- maintaining socket subscriptions created using route params that updates some aggregates

i totally hear what you are saying re: mounting watchers during bootstrap (this is what i do for notification system), the question is how many open saga processing do you want running in the background at **all** times? also, nothing in the current setup prevents you from doing that anyway

skosno:
I definitely agree you do not want too many sagas open in the background, thanks for all the feedback.

chaintng:
Cancel saga after `LOCATION_CHANGE` doesn't work in the case i route to the same PATH but with appending query string.... Sagas then stop working, they won't cancel the previous sagas, but both (old & new injection).

```
export function* githubData() {
  // Fork watcher so we can continue execution
[...]

PR #1106, opened 2016-10-14
automatically cancel running sagas on location change

aldaty:
Currently sagas associated with a route are loaded asynchronously as many times, as the user of an application has entered that route. To circumvent this, developers have to wrap each of their sagas in a `fork - take(LOCATION_CHANGE) - cancel` block. 

Sagas are not meant to be duplicated by default - if a developer wants their saga to run as many times, as the user has visited a route, then redux-saga provides a way to do that using the "spawn" effect, and it would be best to explicitly state that this is happening in the application's code, not implicitly have it happen by NOT doing something in the code.

At least as long as there is no solution preventing the router from ever loading a saga that is already running, I think at least this much can be done to make sagas run as if they were not loaded asynchronously.

### DIFF
--- a/app/containers/HomePage/sagas.js
+++ b/app/containers/HomePage/sagas.js
@@ -2,8 +2,7 @@
  * Gets the repositories of the user from Github
  */
 
-import { take, call, put, select, fork, cancel } from 'redux-saga/effects';
-import { LOCATION_CHANGE } from 'react-router-redux';
+import { take, call, put, select } from 'redux-saga/effects';
 import { LOAD_REPOS } from 'containers/App/constants';
 import { reposLoaded, repoLoadingError } from 'containers/App/actions';
 
@@ -37,19 +36,7 @@ export function* getReposWatcher() {
   }
 }
 
-/**
- * Root saga manages watcher lifecycle
- */
-export function* githubData() {
-  // Fork watcher so we can continue execution
-  const watcher = yield fork(getReposWatcher);
-
-  // Suspend execution until location changes
-  yield take(LOCATION_CHANGE);
-  yield cancel(watcher);
-}
-
 // Bootstrap sagas
 export default [
-  githubData,
+  getReposWatcher,
 ];
--- a/app/containers/HomePage/tests/sagas.test.js
+++ b/app/containers/HomePage/tests/sagas.test.js
@@ -3,10 +3,9 @@
  */
 
 import expect from 'expect';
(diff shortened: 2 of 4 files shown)


sourcesoft:
I remember when I was first using this boilerplate this took me half a day to figure out why nested route run their sagas multiple times and what is actually happening.

Now I have to use cancel for every saga. Sometimes I have to wrap a watcher inside another watcher to make it work the way I want. This PR should be very helpful for guys like me that didn't know much about sagas when they first started using this boilerplate.

mxstbr:
Interesting, not sure what the full implications of this are at the moment. @justingreenberg could you review this change? What do you think?

justingreenberg:
@aldaty thank you for your work on this!

this has been an ongoing issue—please check out this discussion... https://github.com/mxstbr/react-boilerplate/pull/744#issuecomment-233738166

>Currently sagas associated with a route are loaded asynchronously as many times, as the user of an application has entered that route.

the idea is to load saga logic with chunked route bundle, which is cached. without going into too detailed review, there are a few issues off the top of my head:

 -  i'm not sure we should be explicitly creating side effects from internals... it's a little bit different than adding a utility or demonstrating a pattern (redux is global)
 - `LOCATION_CHANGE` could also mean `location.state` or `location.query` was updated, which will cancel running sagas that manage those processes (see #888)

i have often considered whether or not this boilerplate should provide internal "channels" (`effects.actionChannel`)  for interfacing with basic services such as fetch, maybe the progress bar indicator, etc. (you can see some discussion of this on #826)



@sourcesoft 
> This PR should be very helpful for guys like me that didn't know much about sagas when they first started using this boilerplate.

we really don't want things to be magic...

i learned a lesson adding saga injector to the demo app: it was definitely an elegant solution for our simple purposes, but certainly not meant to be a reference implementation for route-based data loading! i mean it's literally one line `sagas.map(store.runSaga)`

@mxstbr i am hesitant to recommend adding additional functionality to the saga setup right now. it would make more sense to add comments all over the code.

i have been thinking a lot about this and there are a few solutions we have previously discussed in #744 and on gitter :) there are some other solutions i have been playing with but have been swamped with work lately. i will try to put together some demo PRs in the next week or two after i fix jest!!

mxstbr:
> @mxstbr i am hesitant to recommend adding additional functionality to the saga setup right now. it would make more sense to add comments all over the code.

👍  I think we'll close this PR for now then, thanks for taking the time to submit it though @aldaty! Please chime into the ongoing discussions as we try to solidify this setup.

PR #1108, opened 2016-10-15
Redux-saga login flow with auth0-lock

sedubois:
Demonstrates login flow using redux-saga. Uses Auth0 but adapting to another platform shouldn't be difficult. Configured with Facebook in this example, but you can configure Auth0 the way you want, just point to your Auth0 with `AUTH0_CLIENT_ID` and `AUTH0_DOMAIN` environment variables.
- [x] redirect after login
- [x] split code between login and callback routes
- [x] state param to prevent XSRF attacks
- [x] logout
- [x] display login/logout button in header
- [x] display login form inline instead of popup
- [x] display user badge
- [ ] coverage
- [ ] document
- [ ] integrate with header nav (merge https://github.com/mxstbr/react-boilerplate/pull/1079)

NB (can be addressed separately): user profile not persisted in localStorage or backend, only in Redux memory store. Thus, time travel etc will probably break and can't see profile in new tabs.

screen shot 2016-10-20 at 16 00 53

screen shot 2016-10-21 at 15 12 41

screen shot 2016-10-22 at 18 09 02

### DIFF
--- a/app/containers/Login/lib.js
+++ b/app/containers/Login/lib.js
@@ -1,0 +1,46 @@
+import Auth0Lock from 'auth0-lock';
+import { storeSecret } from 'containers/Viewer/lib';
+import { ORIGIN } from 'components/Window/constants';
+
+export const LOCK_CONTAINER_ID = 'lock-container';
+
+let lock;
+
+export function createAndShow(nextPathname) {
+  lock = createLock(nextPathname);
+  lock.show();
+}
+
+function createLock(nextPathname) {
+  const secret = createNonce();
+  storeSecret(secret);
+  return new Auth0Lock(process.env.AUTH0_CLIENT_ID, process.env.AUTH0_DOMAIN, {
+    auth: {
+      redirectUrl: `${ORIGIN}/login/callback`,
+      responseType: 'token',
+      params: {
+        state: JSON.stringify({
+          secret,
+          nextPathname,
+        }),
+      },
+    },
+    container: LOCK_CONTAINER_ID,
+    // other options see https://auth0.com/docs/libraries/lock/v10/customization
+  });
+}
+
+function createNonce() {
+  let text = '';
+  const charset = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+  for (let i = 0; i < randomLength(); i += 1) {
+    text += charset.charAt(Math.floor(Math.random() * charset.length));
+  }
+  return text;
+}
+
+function randomLength() {
+  const minLength = 30;
+  const maxLength = 50;
+  return Math.floor((Math.random() * maxLength) + minLength);
+}
--- a/app/containers/LoginCallback/sagas.js
+++ b/app/containers/LoginCallback/sagas.js
@@ -1,0 +1,33 @@
+/**
+ * 1. sets up auth library and waits for auth token
+ * 2. gets user profile
+ */
+
+import { call, put, take } from 'redux-saga/effects';
+import { browserHistory } from 'react-router';
+import { loginCallback, getProfile } from './lib';
+import { loginCallbackRequest, loginCallbackError } from './actions';
+import { setViewer } from 'containers/Viewer/actions';
+
+function* loginCallbackSaga() {
+  try {
+    const { nextPathname } = yield call(loginCallback);
+    const profile = yield call(getProfile);
+    yield put(setViewer(profile));
+    return nextPathname;
+  } catch (err) {
+    // TODO display error to user
+    yield put(loginCallbackError(err));
+    return '/';
+  }
+}
+
+function* init() {
+  yield take(loginCallbackRequest().type);
+  const nextPathname = yield call(loginCallbackSaga);
+  browserHistory.push(nextPathname);
+}
+
+export default [
+  init,
+];
--- a/app/components/Header/index.js
+++ b/app/components/Header/index.js
@@ -2,15 +2,25 @@ import React from 'react';
 
 import A from 'components/A';
 import Img from 'components/Img';
+import ViewerWidget from 'containers/ViewerWidget';
 import Banner from './banner.jpg';
 import styles from './styles.css';
 
 class Header extends React.Component { // eslint-disable-line react/prefer-stateless-function
+  static propTypes = {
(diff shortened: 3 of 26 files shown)


app/containers/LoginCallback/sagas.js
```diff
@@ -0,0 +1,27 @@
...
+ * 2. gets user profile
+ */
+
+import { call, put } from 'redux-saga/effects';
+import { browserHistory } from 'react-router';
+import { loginCallback, getProfile } from './lib';
+import { loginSuccess, loginError } from './actions';
+
+function* loginCallbackSaga() {
+  // TODO take the proper event so that this saga is run exactly once (when loading route)
```
sedubois: @mxstbr I tried to `take(LOCATION_CHANGE)` but that doesn't work. Do you know what I'm missing?
Current code works but the saga will loop twice (visible when adding some `console.log`).

app/utils/auth.js
```diff
@@ -0,0 +1,40 @@
+const AUTH_SECRET = 'auth-secret';
```
sourcesoft: I haven't used `auth0-lock` yet, quick question: You're calling `checkSecret` to check `AUTH_SECRET` string in the callback handler,  Is `AUTH_SECRET` a public key? I guess the const can be accessed on the client so is it 100% safe and the right way to do this?

app/containers/LoginCallback/sagas.js
sourcesoft: I think you can `fork` the saga inside a watcher and `cancel` when the thread ends. How about making a watcher and use a simple takeLatest and see if it runs twice

app/utils/auth.js
sedubois: @sourcesoft If you look at `checkSecret` you'll see it loads the secret via `popSecret` from `localStorage`. `AUTH_SECRET` is just the key to access `localStorage`. The secret is a random string generated in `nonce` method. Does that answer the question?

sourcesoft: sorry I wasn't paying attention to nonce part

app/containers/LoginCallback/sagas.js
sedubois: @sourcesoft thanks! Didn't think of `cancel`. I tried just doing
```
    const task = yield call(loginCallbackSaga);
    yield cancel(task);
```

But this whole code will still be executed again, which defeats the purpose. I couldn't use `fork(takeLatest, LOCATION_CHANGE, loginCallbackSaga)` either, it seems it doesn't receive the `LOCATION_CHANGE` action. And anyway I see in the logs that `LOCATION_CHANGE` is fired twice when doing the auth callback, one initial time coming from a `null` prev state to `/login/callback` next state, but also immediately a second time going from `/login/callback` to `/login/callback`. The only difference is that in the prev state, the `hash` is populated (with the auth idToken and state query params), and in the next state it's empty. I think auth0 must be doing a redirect onto itself as it's parsing this data, then when done it passes this parsed data to the `lock.on('authenticated', ...)` callback.

So I think I need to create and listen to an event that would be fired once `lock.on('authenticated', ...)` callback is called. (another way would be to somehow make the LOCATION_CHANGE listener work, and check whether the hash is empty, but that would be really hacky)

sourcesoft:
````
export function* loginSaga() {

}

// You can escape using this watcher
export function* loginWatcher() {
  let action;
  while (action = yield take(SOME_ACTION_TYPE)) {
    yield call(loginSaga, action);
  }
}

export function* loginObserver() {
  const loginWatcher = yield fork(loginWatcher);
  const action = yield take(LOCATION_CHANGE);
  // Check condition of action if it's needed
  yield cancel(loginWatcher); // cancel task instance
}
````
I don't know if it helps but this is a pattern I did once when I encountered a similar problem. You can escape having loginWatcher and have 2 generators. It might seem bit hacky I guess.

Comment (sedubois) [inline app/containers/LoginCallback/sagas.js]: OK somehow it's because I `browserHistory.push` within that saga, need to extract

File: app/containers/Login/lib.js [hunk 3]
```diff
@@ -0,0 +1,45 @@
...
+  lock = createLock(nextPathname);
+  lock.show();
+}
+
+function createLock(nextPathname) {
+  const secret = createNonce();
+  storeSecret(secret);
+  return new Auth0Lock(process.env.AUTH0_CLIENT_ID, process.env.AUTH0_DOMAIN, {
+    auth: {
+      redirectUrl: `${window.location.origin}/login/callback`,
```
sourcesoft: I guess `window.location.origin` works only in chrome and FF not IE. It needs a polyfill:
````
if (!window.location.origin) {
  window.location.origin = window.location.protocol + "//" 
    + window.location.hostname 
    + (window.location.port ? ':' + window.location.port : '');
}
````

**Comment by sedubois** [comment]
@mxstbr what do you think?
Review comments from anyone, before I set out to add tests?

File: app/components/UserWidget/index.js [hunk 4]
```diff
@@ -0,0 +1,59 @@
...
+import { Link } from 'react-router';
+import { createStructuredSelector } from 'reselect';
+import { removeToken, loggedIn, loggedOut } from 'components/AuthUser/lib';
+import { logout } from 'components/AuthUser/actions';
+import { selectGivenName, selectPicture } from 'components/AuthUser/selectors';
+import UserBadge from 'components/UserBadge';
+import styles from './styles.css';
+import buttonStyles from 'components/Button/styles.css';
+
+class UserWidget extends React.Component { // eslint-disable-line react/prefer-stateless-function
```
sourcesoft: I think this is a smart components connected to state, same almost goes for `AuthUser` with reducers and actions. Are they supposed to be in `components` folder?

app/components/AuthUser/constants.js
```diff
@@ -0,0 +1,2 @@
+export const SET_AUTH_USER = 'AuthUser/SET_AUTH_USER';
```
sourcesoft: I think action constants should be prefixed with `boilerplate/`.

app/components/UserWidget/index.js
sedubois: OK I'll move this one. And I'm not sure where the `AuthUser` folder should go? It has reducers etc, but nothing to render.

Issue #914, opened 2016-08-24
Feedbacks after deploying to Heroku


loicginoux:
Not sure if it can help but here is few points about deploying the app to heroku:
- because I am using Material-UI, and there is an error which looks like this

```
ERROR in ./~/iconv-lite/encodings/tables/gb18030-ranges.json
Module parse failed: MY_APP_PATH/node_modules/iconv-lite/encodings/tables/gb18030-ranges.json Unexpected token (1:9)
You may need an appropriate loader to handle this file type.
```

I needed to `{ test: /\.json$/, loader: 'json-loader' },` to the `module.loaders` of the file `webpack.dll.babel.js

 
- There is a confusion between the documentation which says to add `postinstall": "npm run build:clean` but the package.json contains already `"postinstall": "npm run build:dll",`. I think the correct one is build:dll...

- Nowhere in the deployement process described in your documentation, we call `npm run build` to actually build the ./build directory. Not sure if it's me or the doc which has a problem...
To have a better look here are the lines that can help:
My Procfile  `web: npm run start:prod`
In my package.json:

```
"prebuild": "npm run build:clean && npm run test",
    "build": "cross-env NODE_ENV=production webpack --config internals/webpack/webpack.prod.babel.js --color -p",
    "build:clean": "npm run test:clean && rimraf ./build",
    "build:dll": "node ./internals/scripts/dependencies.js",
    "start": "cross-env NODE_ENV=development node server",
    "start:tunnel": "cross-env NODE_ENV=development ENABLE_TUNNEL=true node server",
    "start:production": "npm run build && npm run start:prod",
    "start:prod": "cross-env NODE_ENV=production node server",
```

So we this, deploying to heroku works but the /build directory is not found.

- Then I tried to add to my Procfile  `web: npm run start:production`
But then Heroku couldn't start within 60 sec with a memory quota limit error:

```
2016-08-24T15:18:23.896122+00:00 app[web.1]: > cross-env NODE_ENV=production webpack --config internals/webpack/webpack.prod.babel.js --color -p
2016-08-24T15:18:23.896123+00:00 app[web.1]:
2016-08-24T15:19:14.086729+00:00 heroku[web.1]: Process running mem=531M(103.8%)
2016-08-24T15:19:14.086915+00:00 heroku[web.1]: Error R14 (Memory quota exceeded)
[...]

PR #1111, opened 2016-10-16
Issue #914: Tweak to deploy docs to include a heroku postbuild of static files.

nappels:
This PR adds a step to the deploy docs that instructs you to use heroku's `heroku-postbuild` script tag to run `npm run build` in order to build the static assets on the server. Without this step, you have to build the static files locally and track those generated files (BAD!).

### DIFF
--- a/docs/general/deployment.md
+++ b/docs/general/deployment.md
@@ -2,17 +2,19 @@
 
 ## Heroku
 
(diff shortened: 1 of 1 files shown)


KSMorozov:
@nappels works as a charm :100: , thank you! :lion:

PR #1123, opened 2016-10-18
Redux logger + let npm reorder packages

sedubois:


mxstbr:
Not a fan of `redux-logger`, the Redux DevTools do the same thing while not being annoying when you don't want that. Closing, thanks though!

PR #1126, opened 2016-10-19
Update ImmutableJS docs for grammar and add more reference links to ImmutableJS docs

spentay:
Hi!

First off - Thanks for the great work with this project! We're using the the boilerplate as a basis for our own project and the solid foundations have made dev work a lot easier for us ❤️

I wanted to add some guidelines to the docs in our project regarding ImmutableJS so first took a look at the existing .md file.

I found that the wording around the 'Records' section was a little confusing so I've updated it in this PR.
I'll be making these changes as well as adding the other changes that I originally wanted to include in our project but wanted to offer these changes to this repository while I was at it!

Let me know if there are things you disagree with or that could be improved upon and I'll do my best to polish this up!

cc'ing @ralomucho, @mxstbr & @oliverturner for your opinions, 
TIA 😃


docs/js/immutablejs.md
```diff
@@ -11,66 +11,78 @@ for a good explanation of the more intricate benefits it has.
...
-  myData: 'Hello World!',
+  myData: {
+  	message: 'Hello World!'
+  },
 });
 ```
 
-To react to an incoming actions our reducers can use the `.set` and the `.setIn`
-functions.
+When actions are dispatched and our reducers need to set new state they can do so by using setter methods such as [`.set`](https://facebook.github.io/immutable-js/docs/#/Map/set) and [`.update`](https://facebook.github.io/immutable-js/docs/#/Map/update).
```
oliverturner: I'd prefer it if we said something like "our reducers need to _return_ the new state"

docs/js/immutablejs.md
```diff
@@ -11,66 +11,78 @@ for a good explanation of the more intricate benefits it has.
...
 
 const initialState = new StateRecord({}); // initialState is now a new StateRecord instance
-                                          // initialized with myData set by default as 'Hello World!'
+                                          // initialized with myData.message set by default as 'Hello World!'
 ```
 
-Now, if you want to access `myData`, you can just write `state.myData` in your reducer code.
+Now, if you want to access `myData`, you can just write `state.myData` in your reducer code and to access the `message` property you can write `state.myData.message` as you would in a plain javascript object.
 
+It's worth noting that although dot notation can now be used to read properties the same does not apply to setting properties. Any attempts to set a property on a `Record` using dot notation will result in errors.
```
oliverturner: I think we should add the explanation that you _can_ set/update a property on a Record, but  only using `set/merge` to return a new instance of the Record type, and only if the Record type has that key in its definition.

Another caveat we should probably raise is that Record instances will throw if a `size` key is defined.

oliverturner:
@spentay These are great additions: thank you so much!

docs/js/immutablejs.md
spentay: > Another caveat we should probably raise is that Record instances will throw if a size key is defined.

Did not know this! Thanks @oliverturner I'll get on adding that in!
Same goes for set/merge :)

oliverturner: Yeah... Record extends `Collection` and so inherits the methods from `Iterable` but you'd never know looking at the docs (which btw, I often use the canonical example of why automated docs suck!)

Relatedly, the `size` thing is a doozy: really common key in incoming payloads, really baffling (and destructive) behaviour. I say relatedly because the error derives from the collision with the built-in `Collection.size` property!
Again nothing in the docs. I asked Lee Byron directly on Twitter what the deal was and he said the v4 would fix it... but there doesn't seem to have been any activity on Immutable for months now, so... look out I guess?

mxstbr:
Awesome, thanks!

spentay:
Thanks Max! & thanks again for the review @oliverturner <3

docs/js/immutablejs.md
spentay: I missed these points @oliverturner - GH notifications must have slipped past my radar!
I'll be keeping a watch on Immutable anyway so will watch out for that specifically! Then, all being well, we can just bin that caveat :D

PR #1128, opened 2016-10-19
Update Redux DevTools Extension's usage according to new API

zalmoxisus:
[`window.devToolsExtension` is being deprecated](https://github.com/zalmoxisus/redux-devtools-extension/issues/220) in favour of a more clear API. There's no need for `window.devToolsExtension.updateStore` workaround anymore, which invokes misconceptions like https://github.com/zalmoxisus/redux-devtools-extension/issues/231.

[Here](https://medium.com/@zalmoxis/improve-your-development-workflow-with-redux-devtools-extension-f0379227ff83)'s a blog post with more details. Also [there's an npm package](https://github.com/zalmoxisus/redux-devtools-extension#13-use-redux-devtools-extension-package-from-npm) to make things even easier.

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -50,15 +50,6 @@ import { translationMessages } from './i18n';
 const initialState = {};
 const store = configureStore(initialState, browserHistory);
 
-// If you use Redux devTools extension, since v2.0.1, they added an
-// `updateStore`, so any enhancers that change the store object
-// could be used with the devTools' store.
-// As this boilerplate uses Redux & Redux-Saga, the `updateStore` is needed
-// if you want to `take` actions in your Sagas, dispatched from devTools.
-if (window.devToolsExtension) {
-  window.devToolsExtension.updateStore(store);
-}
-
 // Sync history and store, as the react-router-redux reducer
 // is under the non-default key ("routing"), selectLocationState
 // must be provided for resolving how to retrieve the "route" in the state
--- a/app/store.js
+++ b/app/store.js
@@ -9,7 +9,6 @@ import createSagaMiddleware from 'redux-saga';
 import createReducer from './reducers';
 
 const sagaMiddleware = createSagaMiddleware();
-const devtools = window.devToolsExtension || (() => (noop) => noop);
 
 export default function configureStore(initialState = {}, history) {
   // Create the store with two middlewares
@@ -22,13 +21,17 @@ export default function configureStore(initialState = {}, history) {
 
   const enhancers = [
     applyMiddleware(...middlewares),
-    devtools(),
   ];
+
+  // If Redux DevTools Extension is installed use it, otherwise use Redux compose
+  /* eslint-disable no-underscore-dangle */
+  const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
+  /* eslint-enable */
 
   const store = createStore(
     createReducer(),
     fromJS(initialState),
-    compose(...enhancers)
+    composeEnhancers(...enhancers)
   );
 
   // Extensions


somghosh:
hello @mxstbr @zalmoxisus  - I am having some issues with the React Boilerplage + Chrome Devtools.

It was working fine till sometime back (maybe a week or so). What I am seeing now is slightly new display in the "Inspector" (just cosmetic change), but the actions on the left are showing up changes in the state that they in actuality are not causing

E.g. is a case below. I was using Boilerplate 3.0 and this was happening, I made the changes suggested by @zalmoxisus and still this is happening. Any advice?


devtools_error

zalmoxisus:
@somghosh, is the problem only with the Diff? If you look into `State` tab is the state changed? Alternatively, try `redux-logger` to make sure it's a problem of devtools.

Usually weird stuff with logging changes happens when the state is mutated. You can check that by adding `redux-immutable-state-invariant` middleware.

oakley808:
I noticed this too and thought I was going crazy. Sometimes things just don't show up in state. (until a later action triggers it to show up) 
I just noticed this ~2 days ago.  Thanks for the tip on `redux-immutable-state-invariant` @zalmoxisus I'll give that a try.

zalmoxisus:
@oakley808, when having such problems also try Log Monitor. Not showing the updated state could be an issue caused by optimisations in the Inspector Monitor.

oakley808:
@zalmoxisus Log Monitor in Redux DevTools seems to show the same misalignment. (state showing up later than the action that it happened in) 

Sadly I couldn't integrate `redux-immutable-state-invariant` into react-boilerplate without it throwing an error.  Have you seen an example of how to do this in this project?

zalmoxisus:
@oakley808, could you please provide some steps to replicate the misalignment?

zalmoxisus:
Did you blacklisted any actions in the extension's settings or by providing `actionsBlacklist`, `actionsWhitelist` or `predicate` parameter?

oakley808:
Nope. No blacklist/whitelist, or predicate custom settings. I'll look using into `redux-logger`.

somghosh:
Hi @zalmoxisus  - is there a way in the interim period I can use the previous version of the your amazing redux devtool? a version even a week back was working fine. Thanks. How can I roll back in Chrome extensions?

zalmoxisus:
@somghosh Chrome store doesn't keep previous versions, you have to [build](https://github.com/zalmoxisus/redux-devtools-extension#1-for-chrome) it [downloading the release tag](https://github.com/zalmoxisus/redux-devtools-extension/releases), but better help us with a repro example so I could fix it. You could try [our demos](https://github.com/zalmoxisus/redux-devtools-extension#demo) to reproduce (the source is [here](https://github.com/zalmoxisus/redux-devtools-extension/tree/master/examples)).

zalmoxisus:
@somghosh also I guess if you set `latency` parameter to `0` (as in [the example from here](https://github.com/zalmoxisus/redux-devtools-extension/releases/tag/v2.9.0)) it will help. Could you please try and confirm that?

somghosh:
I totally agree @zalmoxisus  - here is my analysis

1. Redux logger works perfectly fine and shows as expected

devtools1

2. Redux devtools has erroneous reporting - see the DIFF for hte same action (last action)

devtools2

devtools3

3. It seems that Devtools is reporting the difference CORRECTLY for the action that is 3rd BEFORE in the queue of actions

I am not sure if this is because of a simple typo in the code somewhere or not

E.g. see these screens


devtools4
devtools5

somghosh:
@zalmoxisus  - I tested the basic counter example and the router example - both seem to work fine with the Redux Devtool tab in Chrome

somghosh:
@zalmoxisus  just put the latency param as well, but I am getting the same error as above screenshots, the diff of Nth action matches the what it should be for N-3th action

`/**
 * Create the store with asynchronously loaded reducers
 */

import { createStore, applyMiddleware, compose } from 'redux';
import { fromJS } from 'immutable';
import { routerMiddleware } from 'react-router-redux';
import createSagaMiddleware from 'redux-saga';
import createReducer from './reducers';
import createLogger from 'redux-logger';
import Immutable from 'immutable';

const logger = createLogger()

const sagaMiddleware = createSagaMiddleware();
// const devtools = window.devToolsExtension || (() => noop => noop);
const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({
                                                                       latency: 1000,
                                                                     }) || compose;

export default function configureStore(initialState = {}, history) {
  // Create the store with two middlewares
  // 1. sagaMiddleware: Makes redux-sagas work
  // 2. routerMiddleware: Syncs the location/URL path to the state
  const middlewares = [
    sagaMiddleware,
    routerMiddleware(history),
    logger,
  ];

  const enhancers = [
    applyMiddleware(...middlewares),

  ];

  const store = createStore(
    createReducer(),
    fromJS(initialState),
    composeEnhancers(...enhancers)
  );

  // Create hook for async sagas
  store.runSaga = sagaMiddleware.run;

  // Make reducers hot reloadable, see http://mxs.is/googmo
  /* istanbul ignore next */
  if (module.hot) {
    System.import('./reducers').then((reducerModule) => {
      const createReducers = reducerModule.default;
      const nextReducers = createReducers(store.asyncReducers);

      store.replaceReducer(nextReducers);
    });
  }

  // Initialize it with no other reducers
  store.asyncReducers = {};
  return store;
}
`

zalmoxisus:
@somghosh, well, since I cannot reproduce that, I'm relying only on the introduced `latency` in the latest.  I added some tweaks and just published `2.9.1`. Please give it a try and let me know if the problem still occurs.

zalmoxisus:
Seems that it doesn't help. Let's continue the investigation in https://github.com/zalmoxisus/redux-devtools-extension/issues/249 as it's not related in some way with the code in `react-boilerplate`.

PR #1134, opened 2016-10-22
Improve <List> tests

chaintng:
enzyme API document state that
to use `find()` in ReactWrapper,
the component must be rendered using `mount()` not `render()`

https://github.com/airbnb/enzyme/blob/master/docs/api/ReactWrapper/find.md
https://github.com/airbnb/enzyme/blob/master/docs/api/mount.md

reference https://github.com/mxstbr/react-boilerplate/issues/1119

### DIFF
--- a/app/components/List/tests/index.test.js
+++ b/app/components/List/tests/index.test.js
@@ -1,12 +1,12 @@
 import React from 'react';
-import { render } from 'enzyme';
+import { shallow } from 'enzyme';
 
 import ListItem from 'components/ListItem';
 import List from '../index';
 
 describe('<List />', () => {
   it('should render the component if no items are passed', () => {
-    const renderedComponent = render(
+    const renderedComponent = shallow(
       <List component={ListItem} />
     );
     expect(renderedComponent.find(ListItem)).toBeDefined();
@@ -17,10 +17,11 @@ describe('<List />', () => {
       { id: 1, name: 'Hello' },
       { id: 2, name: 'World' },
     ];
-    const item = ({ name }) => <div>{name}</div>; // eslint-disable-line react/prop-types
-    const renderedComponent = render(
-      <List items={items} component={item} />
+
+    const renderedComponent = shallow(
+      <List items={items} component={ListItem} />
     );
-    expect(renderedComponent.find(items)).toBeDefined();
+    expect(renderedComponent.find(ListItem)).toBeDefined();
+    expect(renderedComponent.find(ListItem).length).toEqual(2);
   });
 });
--- a/app/components/List/tests/index.test.js
+++ b/app/components/List/tests/index.test.js
@@ -12,16 +12,19 @@ describe('<List />', () => {
     expect(renderedComponent.find(ListItem)).toBeDefined();
   });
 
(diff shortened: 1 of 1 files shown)


app/components/List/tests/index.test.js
```diff
@@ -1,15 +1,22 @@
...
 describe('<List />', () => {
   it('should render the component if no items are passed', () => {
-    const renderedComponent = render(
+    const renderedComponent = mount(
       <List component={ListItem} />
     );
+    expect(renderedComponent.find(List)).toExist();
+  });
+
+  it('should not render the items if no items are passed', () => {
```
spentay: I feel like "expect renderedComponent => ListItem to exist" does not prove "It should not render the items if no items are passed"

Might it not make more sense to check for length/count of the list items to prove this test?
If no items are passed in (empty array) then we would expect to find 0 listItem components.

I might be missing something here though - if so I apologise in advance! :D

app/components/List/tests/index.test.js
```diff
@@ -1,15 +1,22 @@
...
+    const renderedComponent = mount(
```
spentay: It would be a little better performance-wise if we could test this using shallowRender instead of mount.
I can't test this right now but I'm fairly certain that we could write this test effectively using shallowRender

app/components/List/tests/index.test.js
```diff
@@ -18,9 +25,10 @@ describe('<List />', () => {
       'Hello',
       'World',
     ];
-    const renderedComponent = render(
+    const renderedComponent = mount(
       <List items={items} component={ListItem} />
     );
-    expect(renderedComponent.find(items)).toExist();
+    expect(renderedComponent.find(ListItem)).toExist();
+    expect(renderedComponent.find(ListItem).length).toEqual(items.length);
```
spentay: I'd recommend using `.toEqual(2)` here, it's always better to be explicit when checking equality in tests.
In this case the 'items' array is explicitly set and there are no side effects. It's also quite clear that no values are added or removed from the items array but that's not always the case...
Imagine if for some crazy reason the length of items changed during the mounting of the component, leaving items.length as 10, we would be checking that the expected value is equal to 10 without knowing.
I hope that makes sense, it's quite difficult to explain!

mxstbr:
What @spentay said!

samit4me:
@chaintng Are you able to resolve the conflicts and use shallow instead?

chaintng:
@samit4me done 👍

chaintng:
Hi @samit4me 
Thanks a lot for your review and very good explanation.
I try to follow you and i think i have better understanding.
I push a new commit, could you please help me review again ?

Thank you so much.

samit4me:
This is great @chaintng, nice work LGTM ✨ 

After looking at this test again, it seems that the component getting created in the second test seems unnecessary and might even distract you from what's actually getting testing, what do you think? 

```
// this line 
const component = ({ item }) => <ListItem>{item.name}</ListItem>;
```

It might make sense to simply pass the `ListItem` directly. Anyway, it looks good either way and it resolves the original issue. Let me know what you think and then we can get this merged.

PR #1148, opened 2016-10-25
Added node_modules caching for appveyor ci

amilajack:
Since we'll have 4 CI builds running in parallel, thought some caching might speed things up

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -40,6 +40,10 @@ test_script:
   # run build and run tests
   - npm run build
 
+# Cache node_modules for faster builds
+cache:
+  - 'node_modules'
+
 # remove, as appveyor doesn't support secure variables on pr builds
 # so `COVERALLS_REPO_TOKEN` cannot be set, without hard-coding in this file
 #on_success:


sedubois:
Maybe adding an `npm prune` (or whatever the yarn equivalent is), to avoid bad surprises?

amilajack:
If i remember correctly, `npm prune` strips out all unused deps?

sedubois:
Yes, `npm prune` prunes all deps which are in node_modules but not referenced in package.json.

OTOH it seems like yarn intends to prune automatically, but there might be a [bug](https://github.com/yarnpkg/yarn/issues/696) currently. Maybe we could live with it for the time being, knowing the issue could fix itself "soon".

amilajack:
Should this be added to `travis.yml` as well>

sedubois:
I'm not familiar with the current CI setup. As I mentioned in previous comment, if test automation is done using `yarn` instead of `npm`, maybe we can forget about prune completely?

amilajack:
I think we should use yarn when its more stable

sedubois:
@amilajack any idea why coveralls got stuck on this branch and others?

amilajack:
I was wondering the same. I'm not sure. I don't have any experience with it.

sedubois:
@amilajack I have no handle on its configuration so can't do anything myself

mxstbr:
This LGTM, thanks!

PR #1149, opened 2016-10-26
Add withMutations to ImmutableJS docs

spen:
Hi All,
I've noticed a lot of .set().set().set() type chaining of ImmutableJS data structures over a number of projects - this approach is less than ideal performance-wise.  
I'm currently looking at adding something similar to the changes below to a project that I'm working on at the moment but thought that the community at large might benefit from being made aware of performance disadvantages of chaining multiple mutations and the solution that withMutations brings.

I also have another PR open for this particular file that fixes up some of the wording of the rest of the document... they can be landed irrespective of each other but it may be helpful to take a look at this one while reviewing this PR - #1126

### DIFF
--- a/docs/js/immutablejs.md
+++ b/docs/js/immutablejs.md
@@ -80,7 +80,7 @@ Because a new instance is created and returned each time a value is set or updat
 To avoid creating multiple instances of the data structure, we can utilise ImmutableJS's [`.withMutations`](https://facebook.github.io/immutable-js/docs/#/Map/withMutations), batching all of our changes in to a single mutation.
 
 ```JS
-const stateMap = Map({
+const stateMap = Map({ // eslint-disable-next-line new-cap
   a: 'A',
   b: 'B',
   c: 'C'
--- a/docs/js/immutablejs.md
+++ b/docs/js/immutablejs.md
@@ -74,31 +74,31 @@ const initialState = new StateRecord({}); // initialState is now a new StateReco
 
 Now, if you want to access `myData`, you can just write `state.myData` in your reducer code.
 
-## Making Complex Mutations Efficiently
+## Performance Tip: Batching Complex Mutations
 
 Because a new instance is created and returned each time a value is set or updated on an immutable data structure, complex mutations that aim to make changes to multiple parts of the data structure can cause a performance penalty.  
 To avoid creating multiple instances of the data structure, we can utilise ImmutableJS's [`.withMutations`](https://facebook.github.io/immutable-js/docs/#/Map/withMutations), batching all of our changes in to a single mutation.
 
 ```JS
 const stateMap = Map({
-    a: 'A',
-    b: 'B',
-    c: 'C'
+  a: 'A',
+  b: 'B',
+  c: 'C'
 });
 
 // Creates 3 new Map instances
 const newState = stateMap
-					.set('a', 1)
-					.set('b', 2)
-					.set('c', 3);
+	.set('a', 1)
+	.set('b', 2)
+	.set('c', 3);
 ...
 
 // Creates only 1 new Map instance - more performant
-const newState = stateMap.withMutations( intermediateState => {
+const newState = stateMap.withMutations((intermediateState) => {
 	intermediateState
 		.set('a', 1)
 		.set('b', 2)
-		.set('c', 3)
+		.set('c', 3);
 });				
 ```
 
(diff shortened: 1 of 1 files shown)


spentay:
Looks very interesting! I'll take a closer look and give these a test run this evening :)
Thanks heaps @justingreenberg

mxstbr:
This looks like a great addition to the docs though! What do you think @justingreenberg?

docs/js/immutablejs.md
```diff
@@ -74,3 +74,33 @@ const initialState = new StateRecord({}); // initialState is now a new StateReco
 
 Now, if you want to access `myData`, you can just write `state.myData` in your reducer code.
 
+## Making Complex Mutations Efficiently
```
justingreenberg: I would maybe recommend something like "Performance Tip: Batching complex mutations"

docs/js/immutablejs.md
```diff
@@ -74,3 +74,33 @@ const initialState = new StateRecord({}); // initialState is now a new StateReco
...
+```JS
+const stateMap = Map({
+    a: 'A',
+    b: 'B',
+    c: 'C'
+});
+
+// Creates 3 new Map instances
+const newState = stateMap
+					.set('a', 1)
```
justingreenberg: nit: please use 2 spaces to match project style guidelines

code style looks mostly good, i would just quickly paste these blocks into project file for linting just be to sure

docs/js/immutablejs.md
```diff
@@ -74,3 +74,33 @@ const initialState = new StateRecord({}); // initialState is now a new StateReco
...
+					.set('b', 2)
+					.set('c', 3);
+...
+
+// Creates only 1 new Map instance - more performant
+const newState = stateMap.withMutations( intermediateState => {
```
justingreenberg: nit: likewise, please wrap callback arguments `stateMap.withMutations((intermediateState) => {` as it is also standard to this codebase

spentay:
Thanks for the feedback @justingreenberg! I'll get on this in the next hour or so :)

spentay:
Still one issue with linting - the `Map()` from `Immutable` - any thoughts or ideas here?

screen shot 2016-11-03 at 6 01 38 pm

mxstbr:
```
// eslint-disable-next-line new-cap
```

😉

spentay:
Worth adding to the example - I was hesitant to add that in case it caused confusion!
If so I can have that in there in a jiffy!

mxstbr:
No that's perfectly fine in this case.

spentay:
Sorry Max... perfectly fine to add or perfectly fine to omit? :P

mxstbr:
Haha sorry, it's worth adding in this case! Sorry for being unclear.

docs/js/immutablejs.md
```diff
@@ -74,3 +74,33 @@ const initialState = new StateRecord({}); // initialState is now a new StateReco
...
+const newState = stateMap.withMutations((intermediateState) => {
+	intermediateState
+		.set('a', 1)
+		.set('b', 2)
+		.set('c', 3);
+});				
+```
+
+From the [official docs](https://facebook.github.io/immutable-js/docs/#/Map/withMutations): 
+> Note: Not all methods can be used on a mutable collection or within withMutations! Only set and merge may be used mutatively. >
```
justingreenberg: If you implement this for the demo app, you will notice this is not accurate and that withMutations works for batching some deep operations as well, in our case setIn 🤔

spentay: Well that's interesting! I'm going to explore this over the weekend, see if it's been discussed in depth over at the Immutable repo... I think it's worth including so that people don't naively wrap all of their mutations and hit errors but depending on findings it could perhaps come with a caveat.

spentay:
@mxstbr no problem! I just wanted to be doubly sure :P
 
@justingreenberg Thanks for putting together the bench! Very interesting to see just how much of a difference it makes! It's obviously tiny in terms of actual time when isolated to one instance but in a large app these ~300%s could add up to become sizeable!

spen:
Hi guys, sorry that this took a little while to make such a tiny update! I wanted to spend some time investigating @justingreenberg's point about `withMutations` and `setIn`.
I couldn't get it to work just now but will look further in to it (the REPL I was using is a little outdated so testing in RBP example will be more accurate). For now thought, I thought I'd simply add the ignore.  
It might be that ImmutableJS's docs need a slight update on this point - I'd rather see the change here first and update the note here after to avoid any confusion :)

KarandikarMihir:
Closing because stale.

PR #1151, opened 2016-10-26
Support socket connection

kwent:
Allow server to connect through a socket.


mxstbr:
Why should we add this by default? What's the benefit?

kwent:
@mxstbr I just found myself pushing an app in production and we use Nginx. Nginx is listening on a port and relaying traffic to a socket. So i needed my node app to listen to a socket and this was not supported.

Maybe some other people will run into this issue.

Regards,

mxstbr:
Thanks for this great reference PR, I'm certain people will get inspiration from it! As of right now, I don't think we'll merge it by default, since it's a step not everybody will need.

Thank you for taking the time to contribute this back to `react-boilerplate`, we really appreciate it!

PR #1155, opened 2016-10-27
Display QRCode for ngrok url

posva:
When running `npm run start:tunnel` display a QR code with the ngrok URL
to make it easier to test your app on mobile devices

Please tell me if you find this feature useful as I do before I add a test 🙂 

Edit: here's a screenshot
screen shot 2016-10-27 at 11 21 20


sedubois:
Nice idea! If we merge this it could need some tests yes, and also add a console message to the user? Also personally I don't even know how I should scan QR codes from my phone, it would probably help to update the docs to mention and explain this cool feature.

posva:
Glad you like it! I'll add the info log, a test and some docs in some hours

sedubois:
Would be nice to know what @mxstbr thinks, even if he's quite busy ATM 😉

posva:
Then I'll wait for his input 😄

mxstbr:
Not a fan of QR codes personally, to be completely honest. What benefit does this have? I have to have a separate app installed, then open that app and scan that QR code – might as well enter the short URL we have?

posva:
Yes, but it takes more time to manually enter the url. It's even worse if you make any mistake given the nature of ngrok urls

mxstbr:
I guess that makes sense, I still don't like just throwing that huge blob up into a users terminal though. Is there a nicer way to do this? Maybe with some sort of link that, when clicked on, shows the QR code?

posva:
I did a quick search and found about a service that does that: https://chart.googleapis.com/chart?cht=qr&chl=github.com&chs=360x360&choe=UTF-8&chld=L|2 (replace github.com with the actual url)
I still think directly having the QR code is better because not all terminals allow clicking on the console and it extends the time to reach the website on your phone.

That link can be added below the proxy link. It's a bit long though

nhducit:
If we display QR code in console user have to scan QR right after start dev server because webpack will display a bunch of building message.

How about add QR code in "/about" page, or display it in the footer.

posva:
@nhducit No, it won't. It looks as in the screenshot

nhducit:
@posva  the QR code is displayed when we run `npm run start` right?  whenever we modify source code console display new webpack log message then have to scroll up to see the QR code.

posva:
It display a new line when you changes something. Try it to actually see it 🙂

KarandikarMihir:
The only thing holding me back is the size of that QR code. Is there any way we can reduce it?

posva:
@KarandikarMihir Unfortunately no, it's already using Unicode chars to make it look smaller. None of my teammates found it big while using it. Since it's displayed only in the tunnel mode and only when the server is launched, I didn't have any issue about it. Keep in mind that if the QR code is too small it may be hard to scan it.

samit4me:
When I first saw this I was like woah that's pretty cool a QRCode in the console, but after thinking about it  I tend to agree with @mxstbr and @KarandikarMihir, it's fairly large and distracting and it seems like it's actually more effort to use than copying and pasting the link. Thanks for sharing, today I learned you can put a QRCode in the console, much awesome 😍

posva:
> it's actually more effort to use than copying and pasting the link

Unfortunately, you cannot copy/paste the link from a mobile device (without any extra app)

It's something my team really enjoyed so I shared it. I'm sure people will find useful to have access to a qr code of any kind (it can be a link to generate the QR code as said below).

The PR being outdated I'm closing this. More than 3 months have passed 😆

PR #1161, opened 2016-10-29
Support 'browser' property

episage:
Webpack will respect `browser` property in imported modules. 

See `https://github.com/defunctzombie/package-browser-field-spec` for details.


mxstbr:
Awesome, thanks!

Issue #1055, opened 2016-09-28
XAMPP shows blank page, even after successful build


ferroariel:
Once built the project (npm run build), and put files inside a folder on my xampp local installation (just prior to move them inside a folder on my production apache server), the pages renders empty, no JS errors, all files loaded.

My stack is:
* RBP v3.0.0
* npm v3.10.6
* node v4.4.7

gihrig:
3.0.0 is pretty old, can you try a new clone of the project in the same setup to see if that works?

Also consider trying a fresh clone of 3.0.0 (assuming you have a lot of development on 3.0.0) to see if that works.

Since 3.0.0 there have been many breaking changes in dependencies. If you have ever run an `npm install` since your original clone, it's certain that some dependencies have changed. Several of these are breaking changes, in spite of semver.

investlift:
Is there any error message in the console?

amilajack:
Why are you running this in XAMPP?

KarandikarMihir:
@amilajack What's your point?

amilajack:
@KarandikarMihir Makes it pretty hard to debug if he says it doesn't work in xampp. I don't know if he ran the boilerplate with `npm start` and it worked or not.

@ferroariel do you get any issues when running this with `npm serve`?

KarandikarMihir:
@amilakack True. But `npm start` would always work, because it uses webpack dev server. I managed to deploy my project on IIS server, which is principally same as XAMPP. No extra efforts were needed other than url rewriting. I would just dump build folder in IIS virtual directory. Can't think of why it would fail. 🤔

amilajack:
Seems like a XAMPP issue than an issue with this project. Cant confirm that yet tho.

ferroariel:
I want to let you know that it was my fault. It was related to the way React Router works. 
I was using route paths like "/"and "/search". So after building and moving to the final destination in my xampp server, the real URl end up being  "http://localhost/myapp". 
So the problem was that the route path string I used. I changed both routes to "/myapp/" and "/myapp/search". 
As long as I did it, everything worked OK. 
Also I took note of  [this post](// http://stackoverflow.com/questions/33376414/adding-a-base-url-to-an-app-using-redux-router-react-router) on SO, to better solve this problem when putting deployed files on a sub directory. 
Thanks for your comments!

PR #1175, opened 2016-11-02
Remove DevTools from Production

ansumanshah:
Added Env variables check to DevTools code based on the new API

Sorry for [#1055](https://github.com/mxstbr/react-boilerplate/pull/1005)


mxstbr:
Do we need to do this in `internals/templates/store.js` too? (that's the one that gets put there after running `npm run clean`)

mxstbr:
Thank you, this looks great!

zalmoxisus:
Just to notice that allowing Redux DevTools Extension in production isn't wrong and [rather helpful](https://github.com/zalmoxisus/redux-devtools-extension/issues/185), though agree that better to allow it explecitely.

PR #1178, opened 2016-11-02
Add shouldComponentUpdate to FeaturePage (demo)

justingreenberg:
companion to #811

the `FeaturePage` component class slipped through the cracks of #811, since it used to be a functional component when the pull was first opened... but when i thought about it, this page doesn't really fit the bill for `PureComponent` either. it doesn't need any external props or state to render itself. this is idiomatic use case for `shouldComponentUpdate`—the component only needs to render 1x. the demo is a cool way to expose these simple patterns which might not be obvious to someone new to react... for example, they might ask themselves why using this optimization on a parent component doesn't break it's descendants. in this case, why do the `react-intl` child components still update (ie when toggling locale)? 📚


mxstbr:
That's a great idea @justingreenberg, love it!

Issue #1181, opened 2016-11-03
i18n select box not updated


themre:
Hi,
I noticed that combo box for language selector is not updated.

## How to reproduce
Select `de` for language and german language is not selected in the combo box although content is in german. 

## Other
Btw, how can we put language in the URL so we can send it to someone to look in specific language? Thanks! Very nice boilerplate.

mxstbr:
Arg, that slipped through with the v3.3 release! Would appreciate a PR fixing that.

gihrig:
Also, 
- language selection is lost on Not Found page
- language selection lost when navigating via address bar

Should the language selection be persisted via cookie? I see that Europe is kinda down on cookies, but is there a better way?

mxstbr:
> Should the language selection be persisted via cookie? I see that Europe is kinda down on cookies, but is there a better way?

I don't quite understand why we would need to do this, it should be saved in the redux state?

gihrig:
> it should be saved in the redux state?

That sounds ideal, provided that the language selection persists across browser open/close or new tabs. Presently that is not the case. OTOH, I suppose persisting redux state may be beyond the scope of the boilerplate?

As an example: 
1. Open the home page at `localhost:3000`
2. Select German language
3. Open the Features page by editing the URL to `localhost:3000/features`

The Features page comes up in English.

gretzky:
Closing due to inactivity

PR #1184, opened 2016-11-03
Update styles.css

stmoreau:
Just a small fix on style of the Footer select element to be inligned with the rest of the footer.

### DIFF
--- a/app/components/Footer/styles.css
+++ b/app/components/Footer/styles.css
@@ -4,3 +4,6 @@
   padding: 3em 0;
   border-top: 1px solid #666;
 }
+footer section:nth-child(2) {
+    margin-top: 11px;
+}


stmoreau:
And by the way in the home page when you change the language to 'de' and the page refreshes, the select element is still showing by default 'en'.
To go back to to 'en' you have to click 'de' (that makes no change) and then again on 'en'.

mxstbr:
We switched to `styled-components`, so the `styles.css` files no longer exist! Would you mind resubmitting this change against an up-to-date `dev` branch? Thanks!

Also, for that second issue we have #1181

zamrq:
@mxstbr , any critical reason to change style.css to styled-components ?

Issue #1185, opened 2016-11-03
i18n.js - Can't resolve 'containers/App/constants' after 'npm run clean'


adjnor:
-Node: 7.0.0 (also tried with 6.9.1)
-NPM: 3.10.8

Steps to reproduce:
1. git clone --depth=1 https://github.com/mxstbr/react-boilerplate.git
2. cd react-boilerplate
3. npm run setup
4. npm run clean
5. npm start (or npm run start)

Results in:

```
ERROR in ./app/i18n.js
Module not found: Error: Can't resolve 'containers/App/constants' in '/Users/MyUsername/Work/react-boilerplate/app'
 @ ./app/i18n.js 8:0-58
 @ ./app/app.js
 @ multi main
```

Fix is adding back a 'constants.js' file in 'containers/App/' with the line:
`export const DEFAULT_LOCALE = 'en';`

connorbode:
Was just going to report this; I can confirm that after running `npm run clean`, the resulting project is missing `app/containers/App/constants.js`.  This was a bit confusing when just starting out.

mxstbr:
Ugh, would appreciate a PR fixing this!

adjnor:
I'm on it!

mxstbr:
Released as v3.3.1, thanks for the quick fix @adjnor, amazing!

sebscholl:
Hey, this issue is popping back up for me.

Inside the /containers/App/ directory there is a constants.js file with 

`export` const DEFAULT_LOCALE = 'en';

However the same error is popping up

`ERROR in ./app/i18n.js
Module not found: Error: Can't resolve './translations/en.json' in 'c:\Users\sebsc\Development\rampex\rampex-frontend\app'
 @ ./app/i18n.js 13:0-59
 @ ./app/app.js
 @ multi main`

gihrig:
@sebscholl I just did a clone, yarn install, npm run clean on current `master` and `translation/en.json` is present.

![image](https://cloud.githubusercontent.com/assets/1481063/22311317/d568f922-e307-11e6-9c3d-65a13c9a9240.png)

It contains an empty array.

What version of the boilerplate are you using?

sebscholl:
@gihrig I didn't use yarn install but npm... maybe there's something in there?

gihrig:
I have been using `yarn install`exclusively for a couple of months. The resulting dependency tree has been very solid, `npm install` not so much...

[...]

PR #1186, opened 2016-11-03
Fix 'npm run clean' breaking (#1185)

adjnor:
-Added a constants.js file to templates with the DEFAULT_LOCALE export
-Added a line in clean.js to copy the templates/constants.js file to containers/App/

This resolves the issue of npm run clean breaking because of constants.js not being found since it was being deleted on npm run clean.


mxstbr:
Awesome, thank you so much!

PR #1187, opened 2016-11-03
Fix i18n button not updating (#1181)

adjnor:
Added a value prop to the Toggle component, being passed down to the Select component to explicitly state which option is selected. The problem seemed to be that when selecting a new language, the component would re-render the options in the order they are passed down via the values prop, thus causing the error.

### DIFF
--- a/app/components/Toggle/index.js
+++ b/app/components/Toggle/index.js
@@ -20,7 +20,7 @@ function Toggle(props) {
   }
 
   return (
-    <Select onChange={props.onToggle}>
+    <Select value={props.value} onChange={props.onToggle}>
       {content}
     </Select>
   );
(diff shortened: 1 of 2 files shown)


mxstbr:
This LGTM! @gihrig @sedubois I'm away for the weekend, can one of you merge to dev and cherry pick the commit to master?

gihrig:
I'm still seeing the problem :-/

I did:
`git clone https://github.com/adjnor/react-boilerplate.git react-bp2 && cd react-bp2`
`git checkout -b adjnor-fix-i18n-button`
`yarn`
`npm start`
`localhost:3000`
Change language to `de`: Language changes, but button still shows `en`

I have cleared Application > Clear storage > Clear site data, then opened new tab.

When changing language, I get the following console error:
image
etc...

adjnor:
@gihrig, when you checkout my branch it should be 'git checkout fix-i18n-button' otherwise I think you're creating a new branch.
As for the errors in the console, it's unrelated to this PR, it comes from react-intl complaining when a message has not been passed for the chosen locale

gihrig:
Yes, I copied the URL from Github, git complained, so I edited without really looking, My bad!

So yes, works as expected now. Thanks for contributing :tada:

Issue #1180, opened 2016-11-03
Document how to keep project up-to-date with react-boilerplate


samit4me:
I've seen at least two people asking how to keep their project up-to-date with react-boilerplate.

Would you be happy for me to create a new section under [Documentation](https://github.com/mxstbr/react-boilerplate#documentation)?

My rough outline would be something like what is already documented in the [CONTRIBUTION](https://github.com/mxstbr/react-boilerplate/blob/master/.github/CONTRIBUTING.md#pull-requests) guidelines!

gihrig:
Previously discussed in #600, #889, #1047.

As those issues attest, this is not an easy task. The ideal would be to split the boilerplate and the application into separate modules as Create React App does. I seem to recall @mxstbr discussing this somewhere, but I'm not finding it now. :-/ In any case, this is a large task, that when accomplished will move this boilerplate lightyears ahead.

mxstbr:
I want `react-boilerplate` to be a fork of `create-react-app`. This will a) give us some goodies in configuration that we don't currently have, but b) it will allow updating more easily.

The problem is that doing so requires a huge rewrite of the whole thing, and I don't have time to do that at the moment.

That being said, let's caution people in the FAQ that while it's possible to update, it's very hard for now. I don't want to set the expectation that it's simple or easy to do, which it isn't right now.

ansumanshah:
I am trying to create a universal boilerplate using `create-react-app` and `react-native` with `styled-components` taking some bits on project structure from this awesome project and `este` would be happy to submit a PR/Fork if needed

samit4me:
@gihrig @mxstbr Thanks for the info, that makes a lot of sense, I too really like the idea of separating the boilerplate from the application 👍 Your totally right, it is not easy to update. Will try and write up something in FAQs.

GGAlanSmithee:
This is an interesting topic. For the sake of discussion, this is how I am doing it:

I have my own template proejct which has this (along with a few other projects) as git submodules, that I keep around as references. This template project contains mostly configuration and build related code, and almost no application code. When there is an interesting development in any of the reference projects, such as this, I patch it to my template project.

I then have a couple of actual projects which contains application code, that are based on the template project. I try to avoid making changes to anything _but_ application code in these, and when the template project updates, i rebase these projects on it.

This workflow adds a bit of manual labor and overhead, but it's a nice intermediate step which allows me to keep my projects up to date with evolving tech.

nonotest:
[...]

PR #1192, opened 2016-11-04
Faq staying updated with boilerplate

samit4me:
Keeping your project up-to-date with react-boilerplate can be difficult and there is *no recommended approach* at the moment. There is a long term goal to change this (see #1180), so this PR is more or less to document the official response to the common question "how to keep my project up-to-date or in sync with the boilerplate". Would love to hear feedback!

Also added a TOC as I feel this makes it much easier to navigate.

### DIFF
--- a/docs/general/faq.md
+++ b/docs/general/faq.md
@@ -1,5 +1,27 @@
 # Frequently Asked Questions
 
+## Table of Contents
+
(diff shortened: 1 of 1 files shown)


mxstbr:
LGTM, thanks!

Issue #1189, opened 2016-11-04
Cannot run generate after "npm run clean" of pristine clone of 3.3.1


jacobk:
`npm run generate` throws an error when run after `npm run clean`

**My environment**
```shell
</tmp/react-boilerplate> (master)
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.11.5
BuildVersion:	15F34
</tmp/react-boilerplate> (master)
$ npm --version
3.10.8
</tmp/react-boilerplate> (master)
$ node --version
v5.0.0
</tmp/react-boilerplate> (master=)
$ yarn --version
0.16.1
```

**How to reproduce**
1. Clone 3.3.1
2. Run `npm run setup`
3. Run `npm run clean`
4. Run `npm run generate`
5. Error

**Reproduce error log**
```shell
</tmp>
$ git clone --depth=1 https://github.com/mxstbr/react-boilerplate.git
Cloning into 'react-boilerplate'...
remote: Counting objects: 328, done.
remote: Compressing objects: 100% (280/280), done.
remote: Total 328 (delta 49), reused 201 (delta 33), pack-reused 0
Receiving objects: 100% (328/328), 1.80 MiB | 1.60 MiB/s, done.
Resolving deltas: 100% (49/49), done.
Checking connectivity... done.
</tmp>
$ cd react-boilerplate/
</tmp/react-boilerplate> (master=)
$ git describe --abbrev=0 --tags
v3.3.1
</tmp/react-boilerplate> (master=)
$ npm run setup

> react-boilerplate@3.3.1 presetup /private/tmp/react-boilerplate
> npm i chalk shelljs

react-boilerplate@3.3.1 /private/tmp/react-boilerplate
├─┬ chalk@1.1.3
│ ├── ansi-styles@2.2.1
│ ├── escape-string-regexp@1.0.5
│ ├─┬ has-ansi@2.0.0
│ │ └── ansi-regex@2.0.0
│ ├── strip-ansi@3.0.1
│ └── supports-color@2.0.0
└─┬ shelljs@0.7.5
  ├─┬ glob@7.1.1
  │ ├── fs.realpath@1.0.0
  │ ├─┬ inflight@1.0.6
  │ │ └── wrappy@1.0.2
  │ ├── inherits@2.0.3
  │ ├─┬ minimatch@3.0.3
  │ │ └─┬ brace-expansion@1.1.6
  │ │   ├── balanced-match@0.4.2
[...]

PR #1197, opened 2016-11-05
Node version support update

samit4me:
If using Node v5 or older and you followed the steps listed in **issue #1189**, it would fail on both macOS and Windows.

According to [Node Green](http://node.green/#let), Node v5 should support the use of `let`, but it appears that this is not true given the error. To resolve this I've added `use strict` to the top of `internals/generators/component/index.js`.

As this boilerplate supports the last three major versions of Node.js, see [FAQs](https://github.com/mxstbr/react-boilerplate/blob/master/docs/general/faq.md#im-using-node-v012-and-the-server-doesnt-work), this change also includes doc updates along with an improved `setup.js` that will exit on old versions of Node.

### DIFF
--- a/internals/scripts/setup.js
+++ b/internals/scripts/setup.js
@@ -25,7 +25,6 @@ cleanRepo(function () {
     interval = animateProgress('Installing dependencies');
   }, 500);
 
-  process.stdout.write('Installing dependencies');
   installDeps();
 });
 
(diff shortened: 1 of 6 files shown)


internals/scripts/setup.js
```diff
@@ -25,7 +25,6 @@ cleanRepo(function () {
     interval = animateProgress('Installing dependencies');
   }, 500);
 
-  process.stdout.write('Installing dependencies');
```
samit4me: This was removed as it appears as a duplicate as shown below:
`Installing dependencies... (This might take a while)Installing dependencies`

mxstbr:
Awesome, thank you!

package.json
```diff
@@ -7,7 +7,7 @@
     "url": "git://github.com/mxstbr/react-boilerplate.git"
   },
   "engines": {
-    "npm": ">=3"
+    "npm": ">=5"
```
MariusRumpf: npm is still version 3 with node 5 and upwards, I think you meant node >= 5 here.

Issue #1196, opened 2016-11-05
Feature page: Error message when change language from English to German


nhducit:
## Description
There are some error messages in console when change language from English to German
<img width="960" alt="screen shot 2016-11-05 at 11 40 30 am" src="https://cloud.githubusercontent.com/assets/4246176/20027691/4915bb6c-a34d-11e6-8c39-41fcba41bb7a.png">

## Steps to reproduce
git clone --depth=1 https://github.com/mxstbr/react-boilerplate.git
npm run setup
open http://localhost:3000/features
change language from English to German
# Versions
v3.3.1
- Node/NPM: v6.9.1
- Browser: Version 54.0.2840.71 (64-bit)

Error message:
```
index.es.js:976 [React Intl] Missing message: "boilerplate.components.Header.features" for locale: "de", using default message as fallback.
index.es.js:976 [React Intl] Missing message: "app.components.LocaleToggle.en" for locale: "de", using default message as fallback.
index.es.js:976 [React Intl] Missing message: "app.components.LocaleToggle.de" for locale: "de", using default message as fallback.
```

adjnor:
These errors are from the react-intl library, a simple way to fix them is to edit the de.json file (in the translations folder) and add translations (or any text) to the empty strings. AFAIK it's the intended behavior when translations are missing.

Issue #1191, opened 2016-11-04
Localization issues after running clean command


web2style:
First of all, there's an issue with importing `DEFAULT_LOCALE` in `i18n.js`. It should be `import { DEFAULT_LOCALE } from './containers/App/constants';` instead of `import { DEFAULT_LOCALE } from './containers/App/constants';`. The current one thrown an error when running `npm run generate language`. Here is a PR for this: https://github.com/mxstbr/react-boilerplate/pull/1190

After fixing this issue and successfully running `npm run generate language` (let's say it is `da`) we have the following code in the bottom of `i18n.js` file:

```
export const translationMessages = {
  en: formatTranslationMessages('en',
  'da',
 enTranslationMessages),
};
```

but it should be:

```
export const translationMessages = {
  en: formatTranslationMessages('en', enTranslationMessages),
  da: formatTranslationMessages('da', daTranslationMessages),
};
```

Unfortunately I'm not good at regexps, so I can't fix the generator myself (https://github.com/mxstbr/react-boilerplate/blob/master/internals/generators/language/index.js#L48-L53 or https://github.com/mxstbr/react-boilerplate/blob/master/internals/generators/language/format-translation-messages.hbs)

But nevertheless after fixing this issue the localization still doesn't work. After adding appropriate translations to `app/translations/da.json` and changing `DEFAULT_LOCALE` to `da`, I still get English translations rendered in the app. It must be Danish, right? I think @chaintng is the right person to ask this question. There's pretty sure something else I haven't found.

Also, there's an issue with chunked polyfill for browsers without Intl support. So in my case after adding `da` language I got:

```
// Chunked polyfill for browsers without Intl support
    ...
    .then(() => Promise.all([
      System.import('intl/locale-data/jsonp/de.js'),
      System.import('intl/locale-data/jsonp/da.js'),
    ]))
    ...
```

I didn't add `de` language, I think it should be removed after running `npm run clean`.

ulrik59:
After checking, the "real" default language of the app is defined in the `/containers/LanguageProvider/reducer.js`: 

````js
const initialState = fromJS({
  locale: 'en', // this is the line to change
});
````

**Comment by web2style** [comment]
@ulrik59 Yes, the language changes for me if I change `locale`.

So I suppose this should be:

```
import { DEFAULT_LOCALE } from '../App/constants';

const initialState = fromJS({
  locale: DEFAULT_LOCALE
});
```

samit4me:
[...]

PR #1200, opened 2016-11-05
Fix console error on lang change

samit4me:
This is a fix for Issues #1196, which throws "[React Intl] Missing message" errors to the console when changing language with the locale toggle in the application footer.

*Note: while this fixes this specific problem, there are outstanding still issues when generating new languages with the command `npm run generate language`  (see #1191) .*


mxstbr:
Thank you!

PR #1204, opened 2016-11-06
Replace start:prod with independent start:production

outdooricon:
The docs are a little confusing on how to get a prod server started. `start:prod` has a undeclared dependency on `build`. So instead, switching the docs to reference the more independent `start:production` cmd is easier for people to follow.


samit4me:
Thank you for that, nice work 👌

mxstbr:
Thanks!

Issue #190, opened 2016-02-23
ApplicationCache on localhost


peter-gordon:
There's some fierce ApplicationCache happening on localhost:3000 for me.

I'm now seeing the "Hello World" screen at http://localhost:3000, even with no node apps running, or running other projects from localhost:3000.

Clearing the Chrome ApplicationCache hasn't helped, along with a few other things I've tried. 
The result is I'm unable to use localhost:3000 until I can figure out how to resolve this.

mxstbr:
Fixed on the `v3.0.0` branch, also see here: https://github.com/mxstbr/react-boilerplate/issues/71

peter-gordon:
Thanks mx. A bit more info as it wasn't mentioned on #71 or #70 
I ended up resolving this by accessing the ServiceWorker internals control panel and unregistering from there: chrome://serviceworker-internals

tconroy:
I had this issue too -- thanks for the solution @peter-gordon that did it for me ( didn't want to wipe my whole app cache ). Cheers.

PR #1205, opened 2016-11-06
chore(selectors): refactor selectors

Dattaya:
- [x] Simplified regular selectors
 - [x] Gave names recommended by the `reselect` docs to selector factories (proposed by @kachkaev, thanks!)


spen:
Awesome work :) I feel this gives a lot more clarity so should help people avoid falling in to anti-patterns :D

I do have one more thing to ask / suggest...

I'm not particularly sold on the naming: `makeSelectLocationState`
I feel this is a little confusing since `make` & `select` are both verbs.
My suggestion would be something like:  
`create______Selector` 

so `makeSelectLocationState` becomes `createLocationStateSelector` 

I feel that might either be confused with createSelector() or could actually be more inline with it... honestly it's all fairly new to me so I could be way off the mark.

Dattaya:
@spen, makes sense, two consecutive verbs do sound weird, if more people are in favor of "createLocationStateSelector" I'll redo the naming in this PR.

Dattaya:
Everyone here (@spen, @gihrig, and I) is in favor of `create...Selector` names for selector factories, in #190 @jimbolla said that the naming in the reselect's docs is an established practice, so I'll leave the final say to @mxstbr. 
Otherwise it's should be good to go; I tested generated templates and didn't notice any issues.

spen:
Just FYI, issue link links to RBP's 190 - [Here's reselects Issue](https://github.com/reactjs/reselect/issues/190) where this was mentioned

kachkaev:
I'd stick to an established practice. New users of `react-boilerplate` will otherwise struggle because of a mismatch in naming. IMHO the goal of this boilerplate is to glue the best practices rather than reinventing the wheel :–)

Dattaya:
It would be nice to get two reviews from collaborators, because this one could've been merged long time ago.

jwinn:
I haven't pulled and tested locally, but the naming changes LGTM and seem inline with `reselect` conventions. 👍

mxstbr:
Same, let's :shipit: then!

Thanks @Dattaya!

gihrig:
Late to the show, but, I discovered a bug...

As of this PR, a cleaned project fails on `npm start`

To reproduce:

`git clone https://github.com/mxstbr/react-boilerplate.git react-bp && cd react-bp`
`git checkout 50fef95d9f2e589385174250e2855fbbdd6907fb`
`yarn` <-- note that `npm start` will work as expected here.
`npm run clean`
`npm start`

In the terminal I get:
image

In the browser console:
image

Dattaya:
Sorry about that, and thanks for the detailed report, @gihrig, more than enough information for me to fix it immediately.

alexidsa:
I'm confused about this change. makeXXX syntax is recommended in reselect docs not as a default approach but as a solution to a specific problem. Do we really need to push it for our app that even doesn't have selectors depending on props (that where selector factories shine)?

Dattaya:
@alexidsa, agreed, selector factories like `makeSelectUsername` make more sense as regular selectors. These too, because as you said we don't depend on props:

```js
createStructuredSelector({
  repos: makeSelectRepos(),
  username: makeSelectUsername(),
  loading: makeSelectLoading(),
  error: makeSelectError(),
})
```

but @mxstbr probably wanted to bring more awareness to this interesting approach.

Note that they were recreated on each call before, I just renamed them in this PR, if you want, you can create a new issue for discussion, or event better, send a PR.

alexidsa:
I agree that this approach is better than a previous one (before your PR we had factories without good naming).

> agreed, selector factories like makeSelectUsername make more sense as regular selectors

What do you mean by that? My point was quite the opposite. I think factories make sense only we actually need them (i. e. we depend on props) not as regular/ordinary selectors.

May be a good compromise would be use normal selectors as default but use factories in case when it's really needed. We need a component that is rendered multiple times for that. Repository component seems like a good option for that.

Dattaya:
> May be a good compromise would be use normal selectors as default but use factories in case when it's really needed.

Definitely! I don't know about other collaborators, but I completely support this idea.

alexidsa:
I'm still getting used to the codebase but I'll try to make a PR in near future.

tomasfrancisco:
I found an inconsistency between `selectors.js.hbs` and `selectors.test.js.hbs`.

Or the domain selector function should be named with 
`makeSelect{{ properCase name }}Domain` at `selectors.js.hbs`
or the import on respectively test should be 
`import { makeSelect{{ properCase name }}Domain } from '../selectors';` at `selectors.test.js.hbs`
instead of 
`import { select{{ properCase name }}Domain } from '../selectors';` at `selectors.test.js.hbs`.

What is the option that should be took?

PR #1207, opened 2016-11-07
Switch out hmre preset for react-hot-loader v3.

### LINKED ISSUE #1068: Switch to react-hot-loader v3
fall cleaning... maybe we should remove `babel-preset-hmre` as it relies heavily on the now deprecated https://github.com/gaearon/react-transform-hmr the question is: how important is maintaining local (react) component state between HMR reloads?

options:
- try to implement React Hot Loader 3 (but... https://github.com/gaearon/react-hot-boilerplate/issues/97#issuecomment-249862775)
- use vanilla webpack HMR (will lose local component state between HMR reloads)

edit: also see https://github.com/danmartinez101/babel-preset-react-hmre/issues/46

marnusw:
Intended to fix #1068, but not ready to merge yet since it seems there are still some issues with supporting async routes: gaearon/react-hot-loader#288.

When running the demo with this setup the static `components/Header` is hot reloaded, but the async `containers/HomePage`, for example, is not.

@MoOx You mentioned in #1068 that you've used RHL3 in other projects. Are you aware of a solution for this?


MoOx:
Sorry, currently I don't use async routes so I can't really help :(

app/app.js
```diff
@@ -92,7 +95,7 @@ const render = (messages) => {
 if (module.hot) {
   // modules.hot.accept does not accept dynamic dependencies,
   // have to be constants at compile-time
-  module.hot.accept('./i18n', () => {
+  module.hot.accept(['./i18n', 'containers/App'], () => {
```
neverfox: I don't this this is sufficient without also calling `require` on `App` in the callback. At least, it didn't work for me until I added that (using 3.4 and RHL@next branch).

MattKetmo:
Hello @marnusw I've just cloned your branch containing react-hot-loader v3 but I'm stuck making it work properly (even after removing the async routes part).

For instance after setup,
```
git clone --depth=1 git@github.com:marnusw/react-boilerplate.git -b reactHotLoader3
cd react-boilerplate
npm run setup
npm start
```

I tried to edit the `HomePage/index.js` render but I still get the message 
```
[HMR] The following modules couldn't be hot updated: (Full reload needed)
```

image

Does this work for you or did I miss something?
Thanks

marnusw:
@MattKetmo I don't have a clean `react-boilerplate` without async stuff to test this in, but I know I read somewhere that it helps to move the router into its own component rather than rendering it directly in `app.js`. This is what I'm using in an actual project where hot reloading is working:

```javascript
// containers/Router.js

import React, { PropTypes } from 'react'
import { applyRouterMiddleware, Router as ReactRouter } from 'react-router'
import { useScroll } from 'react-router-scroll'

import createRoutes from '../../routes'

const Router = ({store, history}) => (
  <ReactRouter
    history={history}
    routes={createRoutes(store)}
    render={
      // Scroll to top when going to a new page, imitating default browser behaviour
      applyRouterMiddleware(useScroll())
    }
  />
)

Router.propTypes = {
  store: PropTypes.object.isRequired,
  history: PropTypes.object.isRequired,
}

export default Router
```

Then in `app.js` replace the router like this:

```javascript
const render = (translatedMessages) => {
  ReactDOM.render(
    <AppContainer>
      <Provider store={store}>
        <LanguageProvider messages={messages}>
          <Router store={store} history={history}/>
        </LanguageProvider>
      </Provider>
    </AppContainer>,
    document.getElementById('app')
  )
}

// ...

// Hot reloadable translation json files
if (module.hot) {
  // modules.hot.accept does not accept dynamic dependencies,
  // have to be constants at compile-time
  module.hot.accept(['./i18n', './containers/Router'], () => {
    render(translationMessages)
  })
}
```

The `Router` could also take the store from the `context`, but passing the prop keeps it simple.

Reading the latest comments on gaearon/react-hot-loader#288 it seems that with Webpack 2.2.0 hot reloading might work with code splitting as well now. 

If the above helps solve your case and you have time to see if it keeps working with async parts then I could make time to clean up this PR and we might get it merged.

app/app.js
marnusw: As I recall it depends on whether you import the component with `require`/`import` initially, and I think whether it is the default or a named export also has an effect. I couldn't find the original info on that, but the current [example in `react-hot-boilerplate`](https://github.com/gaearon/react-hot-boilerplate/blob/next/src/index.js) also does it without an additional require.

justingreenberg:
@marnusw i would love to land this in v3.5, any chance you could take another look at this? 🙏

marnusw:
@justingreenberg let me see what I can do this weekend. I've just been using this without async routes ever since I opened this PR  so I've lost touch with the topic a bit, but I'll see.

marnusw:
I'm all for going with #1871 instead of this. I've taken another look and I couldn't, in quick time, figure out how to make react-hot-loader work with async loaded components. I think the limitation is in react-hot-loader itself, whereas the proposal by @Dattaya with the latest webpack hmr seems to just work. If everyone is in agreement this PR could be closed in favour thereof.

Dattaya:
just merged #1871, we can reconsider `react-hot-loader` when it's more stable.

PR #1208, opened 2016-11-07
Fix default Intl polyfill language

web2style:
When we remove example app, default language is English, not German.


mxstbr:
Good catch, thank you!

PR #1211, opened 2016-11-07
fix(docs): Fix typo in filename

RebootJeff:
Typo in filename broke a link in `docs/general/README.md`. This PR fixes that link and updates others so they all reference the correctly spelled filename.


samit4me:
@RebootJeff That's great, thank you for that! I have nobody but my keyboard to blame 😳

mxstbr:
Thanks!

PR #1212, opened 2016-11-07
fix(example): Use `createStructuredSelector` instead of `createSelector` in `RepoListItem`

Dattaya:
Currently the example app in the dev branch shows an error message when you try to fetch information:

image

`git bisect` pointed me to the Jest integration commit and there I noticed this change: https://github.com/mxstbr/react-boilerplate/commit/98f2d0d931ad0945961b66047b47f519734c9db2#diff-f66d28b72b146f6d22bdc2e7aab68001L57 that broke `RepoListItem` component; unless I'm missing something, it's better to replace `createSelector` with `createStructuredSelector` to fix this error.

### DIFF
--- a/app/containers/RepoListItem/index.js
+++ b/app/containers/RepoListItem/index.js
@@ -6,7 +6,7 @@
 
 import React from 'react';
 import { connect } from 'react-redux';
-import { createSelector } from 'reselect';
+import { createStructuredSelector } from 'reselect';
 import { FormattedNumber } from 'react-intl';
 
 import IssueIcon from './IssueIcon';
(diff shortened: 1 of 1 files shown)


Dattaya:
Need to figure out why tests haven't detected this.

gihrig:
Ran into this while preparing a dependency update.

I can confirm that the code patch fixes the issue in my environment :100:

Dattaya:
@gihrig, thank you for confirmation, much appreciated.
@mxstbr, the reason for false positives could be that props are directly passed and are not returned from the store (?):

```js
    const renderedComponent = renderComponent({
      item,
      currentUser: item.owner.login,
    });
```

Anyhow, I think this is ready to be merged, we can improve tests later.

mxstbr:
Thank you!

PR #1219, opened 2016-11-10
Fix wrong npm version introduced with d202b07

MariusRumpf:
Commit d202b07 introduced a not existing npm version of >= 5 which does not exist. The correct npm version is checked during installation which might otherwise lead to an error. Conclusion might be that this was meant to be node >= 5 like mentioned in #1197.


package.json
```diff
@@ -7,7 +7,7 @@
     "url": "git://github.com/mxstbr/react-boilerplate.git"
   },
   "engines": {
-    "npm": ">=5"
+    "npm": ">=3"
```
samit4me: Should we also add `"node": ">=5"`? Just so it's clearer?

gihrig: Agreed, but should be "node": ">=4". Reason: if Node 4 is to be deprecated other areas need to be adjusted as well (travis and appveyor configs, docs, other?). These changes are out of scope for this PR. I guess it could be argued that `engines: node` could even be out of scope.

samit4me: This was done as part of #1197, so we should be good!

gihrig: @samit4me You're right. My error 😊

Issue #1221, opened 2016-11-11
extract-intl script can't parse LocaleToggle/messages.js


Dattaya:
at least on Windows (dev branch):

```bash
> npm run generate language

> react-boilerplate@3.3.0 generate D:\Projects\react-boilerplate
> plop --plopfile internals/generators/index.js "language"

? [LANGUAGE] What is the language you want to add i18n support for (e.g. "fr", "de")? fr

D:\Projects\react-boilerplate\internals\generators\language\index.js:71
            throw err || stderr;
            ^
Error transforming file: app/containers/LocaleToggle/messages.js
SyntaxError: unknown: [React Intl] `defineMessages()` must be called with an object expression with values that are React Intl Message Descriptors, also defined as object expressions.
```

Does anyone else experience this issue?

samit4me:
Confirmed on macOS also, `git checkout dev` > `npm i` > `npm run generate language` > selected the default `fr` > hit enter and then got the same error you reported!

Dattaya:
@justingreenberg, I need your advice. #813 broke `extract-intl` script, here's an error message (it can't parse `app/containers/LocaleToggle/messages`):

```bash
Run extraction on all filesSyntaxError: unknown: [React Intl] `defineMessages()` must be called with an object expression with values that are React Intl Message Descriptors, also defined as object expressions.
  17 | }
  18 |
> 19 | export default defineMessages(
     |                ^
  20 |   getLocaleMessages(appLocales)
  21 | );
  22 |
    at File.buildCodeFrameError (D:\Projects\react-boilerplate\node_modules\babel-core\lib\transformation\file\index.js:431:15)
    at NodePath.buildCodeFrameError (D:\Projects\react-boilerplate\node_modules\babel-traverse\lib\path\index.js:140:26)
    at assertObjectExpression (D:\Projects\react-boilerplate\node_modules\babel-plugin-react-intl\lib\index.js:285:36)
    at PluginPass.CallExpression (D:\Projects\react-boilerplate\node_modules\babel-plugin-react-intl\lib\index.js:316:21)
    at newFn (D:\Projects\react-boilerplate\node_modules\babel-traverse\lib\visitors.js:276:21)
    at NodePath._call (D:\Projects\react-boilerplate\node_modules\babel-traverse\lib\path\context.js:76:18)
    at NodePath.call (D:\Projects\react-boilerplate\node_modules\babel-traverse\lib\path\context.js:48:17)
[...]

PR #1223, opened 2016-11-12
feat(ci): allow for linting of the templates

Dattaya:
Depends on:
  - [x] #1221
  - [x] #1234
  - [x] #1327
  - [x] #1329

TODO
 - [x] Add comment at the top describing the purpose of `generate-all-templates.js`
 - [x] Add inline comments where necessary
 - [x] `exit` in case of an error generating code from a template?
 - [x] Update `yarn.lock`

What do you folks think about it, do we need to lint and check for errors of the generators themselves and generated code?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,9 @@ node_js:
   - 6
   - 5
 
-script: npm run build
+script:
+  - node ./internals/scripts/generate-templates-for-linting
+  - npm run build
 
 install:
   - npm i -g npm@latest
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -37,6 +37,7 @@ test_script:
   # Output debugging info
   - node --version
   - npm --version
+  - node ./internals/scripts/generate-templates-for-linting
   # run build and run tests
   - npm run build
 
--- a/internals/generators/index.js
+++ b/internals/generators/index.js
@@ -5,6 +5,7 @@
  */
 
 const fs = require('fs');
+const path = require('path');
 const componentGenerator = require('./component/index.js');
 const containerGenerator = require('./container/index.js');
 const routeGenerator = require('./route/index.js');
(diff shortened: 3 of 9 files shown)


mxstbr:
Yes I'd love to land this! We really really need it!

gihrig:
@Dattaya Re: merge conflict:
image

I presume the eslint disable was a temporary measure? Or are you suggesting a different  handling?

The conflict is with #1327 which this PR "depends" on. My thought is that since we are generating a stub to be completed by the end user (developer), the eslint disable is more confusing, it looks more like there might be a reason for it that the dev will not immediately grasp. The `return null` is fine until one looks at a real sagas.js in the example app and sees there is no need for the return.

Now that I have looked at this more closely, maybe a better approach would be to remove the truly useless return and replace it with an informative comment e.g.

> // See example project containers/HomePage/sagas.js

That eliminates the lint error and makes the project more usable all at the same time :smile:

I have edited the conflict with that solution, feel free to disagree 😉

Dattaya:
@gihrig, many thanks for merging those two PRs and looking into this one ❤️, that means I will be able to finish it today. 

> I presume the eslint disable was a temporary measure?

Yes, that's correct, and I prefer your solution, because it's conciser, either way a dev would delete it.

> Now that I have looked at this more closely, maybe a better approach would be to remove the truly useless return and replace it with an informative comment e.g.

Sounds like a good idea to me.

Dattaya:
> Coverage increased (+0.007%) 

Good one, coveralls, 😁
Now my TODO list is empty, except the one TODO you can see in `generate-templates-for-linting` file, but it's not a big deal, `console.error` is synchronous on Mac according to the Node.js docs, but it also behaves synchronously on Windows in my experience, all error messages were successfully flushed into the terminal, it can be left as is until we experience any issues with it. 

Should be ready for review. If you have any questions about this PR, don't hesitate to ask them.

gihrig:
I imported your branch `lint-generated-templates` (88f35e95fd332ffaaeaa3dc4602e1ab5d5a74a30) into my RBP local project.

I pushed to my GitHub repo (setup for Coveralls, Travis and AppVeyor).

Both Travis and AppVeyor successfully completed the build.

Issue #1103, opened 2016-10-13
Wanted: Server-side rendering


asood123:
Hi! I work on [OpenCollective](https://www.opencollective.com) - we help other open source projects (including [react-boilerplate](https://opencollective.com/react-boilerplate) get funding. 

We'd like to rebuild our [frontend](https://github.com/opencollective/opencollective-website) using React-boilerplate. But, Server-side rendering is a must for us - mostly for search crawlers and faster performance.

We are willing to pay someone to implement SSR in react-boilerplate and help the community out in the process. Would anyone be interested? You can reply on this issue.

Needless to say that whatever you build has to meet @mxstbr's approval and be merged into react-boilerplate to be considered complete.

Update: removed my email address. Prefer to discuss this publicly.

sedubois:
(NB just so that it is known, SSR seems to mix very well with GraphQL. [Relay](https://github.com/denvned/isomorphic-relay), [Apollo](http://www.apollostack.com/) and [Relate](https://github.com/relax/relate) all seem to support it. E.g. the in-development [Relax CMS](https://github.com/relax/relax) created and uses Relate which has SSR, seems pretty neat. And I write this in brackets, because GraphQL clients are still on the experimental side 😉 see https://github.com/mxstbr/react-boilerplate/issues/1041 for more GraphQL discussion.)

sedubois:
FYI other threads about SSR: https://github.com/mxstbr/react-boilerplate/pull/625, https://github.com/mxstbr/react-boilerplate/issues/174. The PR looks stale but could give inspiration I guess. Maybe @datapimp could comment?

diegohaz:
I've implemented SSR on some React projects and I plan to do that on my boilerplate (https://github.com/diegohaz/arc). I'll see if I can do that in `react-boilerplate` and I send a PR. :)

peter-mouland:
if it helps, I have a project (much simpler than this, but still fully tested) which add SSR.

It shows how to add SSR using Koa or Express:

https://github.com/peter-mouland/react-lego

sedubois:
Looks like [react-starter-kit](https://www.reactstarterkit.com/) provides SSR.

sedubois:
It also _looks_ like [hjs-webpack](https://github.com/HenrikJoreteg/hjs-webpack) provides SSR. And anyway looks like a nice way to simplify the webpack configuration, automatically comes with dev server, transpiling, HMR, minifying and publishing.

[...]

PR #1236, opened 2016-11-15
Server side rendering

tomazy:
Implements #1103

I've managed to extract the SSR code from one of our apps based on this boilerplate. 
There are still a few issues to fix (see below) but I thought I could  ask you guys to help me with reviewing it. 

I'm including the CSS during the server side rendering but the styled components inject it anyway. It would be great if we could avoid it. Any ideas? Is there any "virtual-css" yet :) ?

TODO
- [x] Use hashes in assets' names (`vendor.css`, `main.js`)
- [x] Fix the content flash (this is not FOUC, but for some reason there is an ugly blink)
- [x] Tests
- [x] Split dev/production middlewares into separate modules
- [x] Handle 404
- [x] npm run clean
- [x] How do people with JS disabled access the german version?
- [x] npm run generate route


mxstbr:
The FOUC is probably because of `styled-components`? Ref https://github.com/styled-components/styled-components/issues/124

tomazy:
@mxstbr That's not FOUC what I'm experiencing here. The page loads and it looks like it should while the tab browser spinner suggests network activity. Once the spinner stops - the whole page becomes blank for half a second and the repaints again. This may have something to do with the code splitting and loading of the chunks. I'm looking into this right now

tomazy:
@mxstbr the flash was caused by incorrect handling of async routes when the a dynamic route was pre-rendered on the server. 	f901878 fixed that.

lucianlature:
@tomazy I'm using your PR locally and I get this error:
```
ERROR in [AssetsWebpackPlugin] Could not read output file [...]/react-boilerplate/server/middlewares/assets.generated.json
```

tomazy:
@lucianlature weird. `assetsPluginInstance` shouldn't be in `webpack.ssr.babel`. How can I reproduce it?

lucianlature:
Probably you already have the `assets.generated.json` in `middlewares` directory. Try to delete it and then run `npm run start`.

tomazy:
@lucianlature thanks for reporting this. It should be fixed now.

app/components/HtmlDocument/index.js
```diff
@@ -0,0 +1,70 @@
...
+
+        {/* app css */}
+        <style type="text/css" dangerouslySetInnerHTML={{ __html: css }} />
+      </head>
+      <body>
+        {/* Display a message if JS has been disabled on the browser. */}
+        <noscript>
+          If you are seeing this message, that means <strong>JavaScript has been disabled on your browser</strong>
+          , please <strong>enable JS</strong> to make this app work.
+        </noscript>
```
mxstbr: Let's change this message, since if JS is disabled one can actually see things now. This should be a popup or something saying that this site will work better with JS enabled.

app/containers/App/reducer.js
```diff
@@ -10,42 +10,73 @@
...
+const userData = combineReducers({
+  repositories,
+});
+
+export default combineReducers({
+  loading,
+  error,
+  currentUser,
+  userData,
+});
```
mxstbr: Let's create a `reducers/` folder and put the individual reducers into separate files within that.

mxstbr: Let's create a reducers/ folder and put the individual reducers into separate files within that.

app/components/HtmlDocument/index.js
mxstbr: We need to make this message look nice. Since users now actually see something when JS is disabled, this should be some sort of popup saying the website works _better_ with JS enabled, but it should be non-intrusive and disappear after a few seconds.

app/containers/HomePage/index.js
```diff
@@ -77,21 +59,29 @@ export class HomePage extends React.PureComponent { // eslint-disable-line react
...
+                    placeholder="mxstbr"
+                    value={this.props.username}
+                    onChange={this.props.onChangeUsername}
+                  />
+                </label>
+                <FormattedMessage
+                  {...messages.userLink}
+                  values={{
+                    link: (<Link to={`/users/${this.props.username}`}>{this.props.username}</Link>),
+                  }}
```
mxstbr: This user link needs to have some sort of design, right now it's a bit ugly in the middle of the page.

docs/general/server-side-rendering.md
```diff
@@ -0,0 +1,23 @@
...
+Servers side rendering should work out of the box.
+
+Here is how we do it.
+
+In order to be able to use our client code on server it has to be transpiled into node.js compatible code.
+Webpack takes care of that with `target` option set to `"node"`.
+
+The entry point for the server webpack build is `app/serverSideRenderAppToStringAtLocation.js`.
+Webpack will generate a couple of files and place them in `server/middleware`. The names of these files will
+start with `"generated."`. They shouldn't be added to the repository (git).
```
mxstbr: Change last sentence to "They shouldn't be checked into git"

internals/webpack/webpack.ssr.babel.js
```diff
@@ -0,0 +1,88 @@
...
+        presets: [
+          ['es2015', { modules: false }],
+          'react',
+          'stage-0',
+        ],
+        plugins: [
+          'transform-react-remove-prop-types',
+          'transform-react-constant-elements',
+          'transform-react-inline-elements',
+        ],
```
mxstbr: Why do we have a separate config here, can't this just use the `babelConfig` from the `package.json`?

tomazy: Not needed indeed. I'll remove it

app/containers/App/reducer.js
tomazy: What benefit would that bring? The reducers are very small here and to me it makes sense to keep them together and see what they do at a glance.

app/components/HtmlDocument/index.js
mxstbr: Yeah I guess, not sure how that would work, whatever you can make happen!

oliverturner:
My biggest concern is over performance: I haven't dived deeply into the SSR code, but it looks as though every request would trigger a fresh render every time, regardless of whether the content were static or not.

@tomazy Is there a caching strategy you have in mind that could reduce unneeded work on the server?

tomazy:
@oliverturner static assets are handled by the appropriate middleware. If the request is not matched then it will be handled by the SSR code. I don't think we should care about caching too much here.  You can always put your server behind a varnish, cloudflare, etc.

app/components/HtmlDocument/index.js
tomazy: That's a lot of work and I'm not sure if that should be a part of this PR. Maybe we could create a separate issue for this and some css wizards might help us. What do you think?

app/containers/App/reducer.js
sourcesoft: I think It's already hard and complicated to edit 3 different files (actions.js, constants.js, reducer.js) and keep track of importing and exporting to have a reducer working compared to easy-to-reason [compact code](https://github.com/infinitered/ignite/blob/master/ignite-base/App/Redux/TemperatureRedux.js) with [modularity](https://github.com/erikras/ducks-modular-redux).

app/components/HtmlDocument/index.js
wmertens: you could use 90s technology and store a "jsflash" message in a session that gets cleared from the second render onwards…

tomazy:
@mxstbr I've addressed all the Big issues that you posted in your review. 
With this PR merged the project is no longer a "client" only boilerplate. The server part is important to make it work (no `index.html`). It's a braking change and probably shouldn't go into v3. I'm not sure what we should do with the http server configuration files as they are not valid any more. 

We would have to implement something like https://github.com/kriasoft/react-starter-kit/blob/master/tools/render.js to make it serverless again.

isaachinman:
If this is going out in 4.0, what's the roadmap for that? Can we expect something production-ready by end of year? Happy to help if anything still needs to be done.

carlos-peru:
Any chance to answer @isaachinman about the ssr roadmap? Also, any idea how complex migrating to ssr (4.0?) will be?

mxstbr:
The main thing I need is people running this in production. I have no idea how this system behaves when you've built a large app with it that gets millions of views per day – does it crash every server you can get or is it totally fine and can still run on a 5$ DigitalOcean Droplet?

I don't think migration will be a thing here, I'm afraid. :( Looking at how intricate of a change this is for the tiny example app, I'm sure it can be done but you'll highly likely be on your own there.

mathieumg:
I've been following this since this Summer and SSR is the thing I'm looking forward to, to finally jump in this boilerplate and start using it heavily.

tomazy:
@mxstbr There is a number of changes in this PR that could be extracted into separate PRs, like the changes in the app.js, some reducers, the user page, etc. I could spend some time extracting these things so that we could have only the minimum patch in this PR to have SSR working. This could help to keep this branch from staling until it's time to merge it. And it would be easier to review it. Would do you think?

mxstbr:
That sounds very good to me!

jgrancher:
Any news on this PR? Sorry to bother, I just can't wait to use it! 💃

mxstbr:
Please use it @jgrancher!

Seriously, please use it and let us know how it works – I can't really merge this in good conscience until a few people have tried and everything works as it should.

zamrq:
untitled

try to use @tomazy  ssr branch

zamrq:
use yarn ssr:start
untitled

clioharp:
I think the issue that @zamrq is having is with the way to state is being set initially in app.js. Right now its ```const initialState = window.APP_STATE || {};```  but if do ```const initialState = {}```, you're just fine

zamrq:
@clioharp  thx for your solution
solved after clean

clioharp:
@tomazy the way you have webpack set up right now, you can't load SVGs. Not sure if that's something you want to change by adding `|svg` to the module loader section.

denisflorkin:
Hi @mxstbr and @tomazy 
I was wondering if there is any plan to keep the serverless option in place once/if the merged on master has happened ? Or, as an alternative build or branch maybe, or is the plan to get rid of serverless entirely and replace it with ssr and 'drop' support for serverless ?

> @mxstbr I've addressed all the Big issues that you posted in your review.
With this PR merged the project is no longer a "client" only boilerplate. The server part is important to make it work (no index.html). It's a braking change and probably shouldn't go into v3.

Thanks

ItsMrAkhil:
Any updates on this PR. sorry to bother. 

But I see there are bugs in PR.
1) I see no image banner.
2) No resourses found. 404 errors in console.
error

3) Whenever I hit enter in get repos input text it is getting reloaded not hitting github API.
4) Clicking on any link is not behaving like SPA. It is hitting the server to get whole data including js files.

erkieh:
@ItsMrAkhil which port do you use?
I get those same errors if I use the SSR service port 3001 instead of the 3000.

tomazy:
@ItsMrAkhil what command do you use to run the app?

ItsMrAkhil:
@tomazy @erkieh I've been using port 3001 instead of 3000. It's a bit confusing. Now it's clear.

@tomazy Why we are using two port? Please, someone, let me know.

But I have an error in console. 

`combineReducers.js:29 Unexpected property "home" found in previous state received by the reducer. Expected to find one of the known reducer property names instead: "route", "global", "language". Unexpected properties will be ignored.`

datoml:
@ItsMrAkhil 
I think thats the problem with server side rendering and async reducer.
On the service side the reducer was loaded and the state was injected for the client.
On the client side however you want to init a state where the home reducer wasnt loaded at this time.
So you get this error.

To solve this we would have to load the reducer on the client too before the init state happens.

ItsMrAkhil:
@datoml thanks, I understood.

ItsMrAkhil:
Sorry to bother. Is there any progress on this PR.

gihrig:
@ItsMrAkhil What it needs most is people to try it out and report back on their findings.

tomazy:
It's difficult for me to find time to work on this PR now. I may find some time in Feb. I'd like to extract as much as possible from this PR into separate smaller PRs so that it would be easier for the maintainers to review and merge it eventually. If anyone interested in SSR would like to help with breaking this PR into smaller chunks then it would help to push it further. For example the reducers and app.js extractions would be quite easy to do.

declanelcocks:
@ItsMrAkhil @tomazy @datoml @erkieh Did any of you manage to fix the issue that you posted?

This one:
`Unexpected property "home" found in previous state received by the reducer. Expected to find one of the known reducer property names instead: "route", "global", "language". Unexpected properties will be ignored.`

I'm also trying to test the SSR codebase and running into this roadblock.

Also get this error on load:

`Uncaught (in promise) TypeError: Cannot read property 'pathname' of undefined at matchRoutes (matchRoutes.js:227)`

tomazy:
@declanelcocks 

Regarding:
`Unexpected property "home" found in previous state received by the reducer. Expected to find one of the known reducer property names instead: "route", "global", "language". Unexpected properties will be ignored.`

This is a warning from `redux-immutable` and I think it can be ignored. 
You see this because of the nature of the code splitting and asynchronous loading of routes. 
On the server the `home` route is resolved and the state initialised with this sub-state. When the client loads the server state it doesn't know about the `home` route yet and that's why you see this warning.

declanelcocks:
@tomazy 

> This is a warning from redux-immutable and I think it can be ignored.

Understood, thanks. I did assume that it was just a warning as it just says the property `home` was unexpected, with nothing crashing.

The bigger issue for me then seems to be:

`Dynamic page loading failed TypeError: Cannot read property 'pathname' of undefined` and `Uncaught (in promise) TypeError: Cannot read property 'pathname' of undefined` which are both coming from `react-router`'s `matchRoutes.js` file. Seems that even when I try to submit the form to load some repos, it refreshes the page completely, and the same when switching between different routes. This is all on `:3000` of course, where the behaviour is expected to be the same as before.

tomazy:
@declanelcocks

> `Dynamic page loading failed TypeError: Cannot read property 'pathname' of undefined and Uncaught (in promise) TypeError: Cannot read property 'pathname' of undefined`

I wasn't experiencing this at the time of writing this PR so I'd assume something has changed in the dependencies (probably the router package(s)). Unfortunately I don't have time right now to look into this. I hope you can try to resolve this. The navigation should work fine once this one is fixed.

declanelcocks:
@tomazy This was my first thought as well, but seems that my `package.json` file is virtually identical to the one in this PR, including the version numbers. I also compared the latest `dev` branch but seems nothing has changed there with the routing. Will keep investigating but if you/anyone have any ideas then would be appreciated. Other than these issues, the PR seems pretty solid to me.

declanelcocks:
@tomazy Found this issue with `react-router` and `react-router-redux` https://github.com/ReactTraining/react-router/issues/4092 which just required me to update both of those packages. Sorry for wasting your time with that one.

On that note, seems pretty solid to me so far, and will keep testing it out. As someone mentioned before, the use of `3000` and `3001` is the only confusing part for me as a user/developer using it. When I make a `curl` request for `3000` or `3001` both are fully rendered responses so the difference is not clear to me!

tomazy:
Thanks @declanelcocks .

During the development when you edit the source code the client will get reloaded by HMR but we also need to force the server side rendering code to reload. That's why we run an auxiliary service on port $PORT + 1 (3001 in your case). The dev middleware will use this service to do the SSR job. It shouldn't be accessed directly - it's only for the dev server to have access the freshest app code for SSR. You should use the app in exactly the same way as previously. 

It unclear to me why this confusion arises. Have I said anywhere anything that causes it?

declanelcocks:
Thanks for the explanation, I just saw that another person asked the question so was also curious about why it's being loaded. Maybe I just thought it was misleading because in the console you write something like `ssr ready @ localhost:3001` so I thought it was a useable page.

datoml:
@declanelcocks 
I don't use this boilerplate anymore.
Can't really help you on this one.

gihrig:
@tomazy Thanks for keeping this up to date. I believe it's important!

Sadly my project is not to a point where I'm ready for SSR, otherwise I would take a more active roll in this PR. 😢 

PS. CI is failing due missing tests...
image

zamrq:
@gihrig would you like to review about @tomazy pr ?
thanks

duminhtam:
Anybody help to solve these conflict?

cristobal-io:
Hi @tomazy, great work done here, thanks.

I'm giving a good shoot at your branch, trying to migrate our app that is actually running with the boilerplate 3.3.3 and I'm running into some issues.

## Connection Refused

First, in a raw install of your branch,  when we first `npm start` the app, it happens the following, when it says that the app is ready to load, I go to the localhost:3000 and the following error appears:

```
[ssr service] ready @ http://localhost:3001
webpack built f396a5290348f7f5ba58 in 17748ms
{ Error: connect ECONNREFUSED 127.0.0.1:3001
    at Object.exports._errnoException (util.js:1026:11)
    at exports._exceptionWithHostPort (util.js:1049:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1090:14)
  code: 'ECONNREFUSED',
  errno: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 3001 }
[ssr service] ready @ http://localhost:3001
```

at the browser it displays the following information:
```
Proxying failed for page rendering service.

        The service maybe restarting so try again in a second.

        Check the console for more information.
```

Then everything works fine at the app. I've checked and I don't have any other app running that could generate any conflict. I guess some sort of race condition.

## Sass

At our app we are using sass, importing a small sass file into the components that we need, I know it's not the best approach but it's what we are doing until we move onto styled components.

The actual instructions were working with the normal react-bootstrap version, so I tried to include the following:

```
    }, {
      test: /\.scss$/,
      exclude: /node_modules/,
      loaders: ['style-loader', 'css-loader', 'sass-loader'],
    }, {
```
at the config webpack.ssr.babel.js under loaders, but I am getting an error saying the following:

```
ReferenceError: window is not defined
    at /Users/username/Code/tmp/react-boilerplate-ssr/server/middlewares/generated.serverEntry.js:3109:30
    at /Users/username/Code/tmp/react-boilerplate-ssr/server/middlewares/generated.serverEntry.js:3104:47
    at module.exports.module.exports (/Users/username/Code/tmp/react-boilerplate-ssr/server/middlewares/generated.serverEntry.js:3126:68)
    at Object.<anonymous> (/Users/username/Code/tmp/react-boilerplate-ssr/server/middlewares/generated.serverEntry.js:3354:37)
    at __webpack_require__ (/Users/username/Code/tmp/react-boilerplate-ssr/server/middlewares/generated.serverEntry.js:21:30)
    at Object.<anonymous> (/Users/username/Code/tmp/react-boilerplate-ssr/server/middlewares/generated.serverEntry.js:1286:69)
    at __webpack_require__ (/Users/username/Code/tmp/react-boilerplate-ssr/server/middlewares/generated.serverEntry.js:21:30)
    at Object.module.exports.prevRoutingState (/Users/username/Code/tmp/react-boilerplate-ssr/server/middlewares/generated.serverEntry.js:1375:77)
    at __webpack_require__ (/Users/username/Code/tmp/react-boilerplate-ssr/server/middlewares/generated.serverEntry.js:21:30)
    at Object.<anonymous> (/Users/username/Code/tmp/react-boilerplate-ssr/server/middlewares/generated.serverEntry.js:2304:73)
```

Any idea how to enable `sass` files?

Thanks.

cristobal-io:
## Sass (solution)

I've found a solution using the [isomorphic-style-loader](https://github.com/kriasoft/isomorphic-style-loader)

I've added at the `webpack.ssr.babel.js` config file the following:

```
    }, {
      test: /\.scss$/,
      exclude: /node_modules/,
      loaders: ['isomorphic-style-loader', 'css-loader', 'sass-loader'],
    }, {
```

Hope it helps.

tomazy:
@duminhtam here you go

tomazy:
@cristobal-io thanks for reporting this. It may be connected to the fact that we're restarting the ssr proxy service in dev mode whenever you change a watched file. I agree we should handle it better. Let me know if you can reliably reproduce that in this branch, then I'd try to handle it gracefully.

brunorzn:
@tomazy I have the same "Connection Refused" issue than @cristobal-io - 
it happens really often (more than 50% of the time I think), and by changing the same file. 

It doesn't seem to happen when no browser client is connected on localhost:3000

nhducit:
@tomazy please notice we may get memory leak when using react-helmet.
```
To use on the server, call rewind() after ReactDOMServer.renderToString or ReactDOMServer.renderToStaticMarkup to get the head data for use in your prerender.

Because this component keeps track of mounted instances, you have to make sure to call rewind on server, or you'll get a memory leak.
```


https://github.com/nfl/react-helmet

tomazy:
@nhducit do you think this is not enough: https://github.com/react-boilerplate/react-boilerplate/pull/1236/files#diff-de02853908e1278058760e3cde3d71ffR68 ?

nhducit:
@tomazy I think it's enough. I searched "rewind" in your repo and cannot found it. But I searched in master branch. My bad :)

arthurCormack:
A question regarding 404s and containers that get data from APIs: What happens when the url path matches a route that contains path parameters, which are used to call the API, and the API then returns no results? I would then like to be able to return a 404 status code from the server instead of a 200. 

Imagine that there is a route defined, as something like: `http://classicalfm.ca/news/arts `
which maps to a container/saga/reducer  that gets data from a REST API ( `http://classicalfm.ca/wp-json/zm-content/v1/getposts?category=arts` ), which would return some JSON. There might be successful results from the API, in which case, we render some articles. Hooray. But what about something like http://classicalfm.ca/news/completegarbage - there are no results for that from the API. 
The best thing to do in that case, insofar as SEO (google!) is concerned, is to return a status code 404. 

I'm looking at the is404 method in serverEntry.js, and all that it seems to do is  `routes.some((r) => r.name === 'notfound')` which I assume just determines whether or not there is a route defined to handle the url path of the request. What if the action that was dispatched when no results were returned set something in state, that could be inspected to determine whether there were results for that url or not, which then was used to determine is404? What do you think the best approach might be?

Best regards,

Arthur

declanelcocks:
@arthurCormack I'm not saying my approach is a good way, but with SSR when I require data for a specific page I like to add a `static get { ... }` to the page which tries to fetch the required data. On the server, it will check for any `static` methods and try to fetch the data required by the page.

I'm not sure I understood your comment correctly though, because you're saying that if the data fails to load, it should show a `404`? In previous projects, I've used the above approach to fetch the data and if there is no data or it doesn't load then I render an error message or empty state on the page. Are you saying this kind of approach is bad for SEO? 🤔

app/components/HtmlDocument/index.js
gihrig: [pure css modal popup](https://www.google.com/search?num=100&q=pure+css+modal+popup&oq=pure+css+modal+popup&gs_l=serp.3..0.23823.25028.0.27824.5.5.0.0.0.0.108.450.4j1.5.0....0...1c.1.64.serp..0.5.444...0i7i30k1j0i13k1.duhWlABm9Hc) is indeed a thing, well, several things 😄

arthurCormack:
@declanelcocks 
Q: "Should we return a 404 if the data fails?" 
A (part 1): https://webmasters.googleblog.com/2011/05/do-404s-hurt-my-site.html
A (part 2): Maybe, but not necessarily so. It depends. I think the question we have to ask first, is whether or not the page/post/ url path exists or not.

In the case of a SPA that is retrieving content data from an API, then it would make sense for that API to determine and be the authority about whether or not the content for that url path actually exists, and by extension, whether or not that url path exists. 

Not all data from APIs should be considered essential for determining whether or not a given page exists. On some sites, there might be multiple elements/comtainers that require data to be loaded. Some of it might be critical for displaying the content at a given url path (like the actual content of a post/article) and some of it it might not be essential, like stuff for a widget in a sidebar. 

For users/browsers, loading things that are rendered client-side, I think its fine to return a not found page with a status code of 200. But for an indexing spider, crawling a site, the 404 can be informative and helpful. It could determine whether a non existing page gets indexed, and indexed, but non-existing urls, could have a negative impact on Link Authority and SEO score. 

The truth is I'm actually not so clear about what the actual impact of not returning a 404 status code for not found pages could be. My boss seems to think its critical though.

nhducit:
I'm trying to use this branch for my project. But get problem with client global objects (window, document-for GA stuff, location) in server-side. Are there any one know how to deal with these object in server-side?

FaustoC:
@nhducit i've included this in my project very well, and it's alredy in production with optimal performance.
for production i simply run the start script in package.json.
For window, document and so on, i just modified my code, if i'm in server enviroment, i don't use window, ga, ecc... so that the page will be rendered with the helmet (so sharing on facebook/twitter wont be useless, like in the normal repo) and when on the browser the js will rerun and use windows to do all the things we need.

Also, i'm very interested in using this PR directly from the boilerplate, i think ssr resolve many problems that make the boilerplate unusable for normal projects where the target client browse with ""normal"" (old) browser or likes to share pages.

nhducit:
@FaustoC yup, this PR works. Do you have any problems when run `npm run start` to start render server in production? because it's using NODE_ENV=development.
furthermore Could you share with me how to deploy react-boilerplate with this PR? just run `npm run start` then config nginx point to port `3001`?

FaustoC:
@nhducit

-  in prod we execute : 
1 ) npm run build
2 ) npm run start:prod (cross-env NODE_ENV=production node server)

- Yes we have a nginx going for 3100 post

-  i use a simple plugin like so
new CopyWebpackPlugin([{
    from: path.join(process.cwd(), 'public'),
    to: 'public',
  }]),

nhducit:
@tomazy 
I got these warning message, from image-loader. I want to help but I don't know which branch I should create PR
```
pebble commented on 28 Jan
DEPRECATED. Configure gifsicle's interlaced option in it's own options. (gifsicle.interlaced)
DEPRECATED. Configure optipng's optimizationLevel option in it's own options.
DEPRECATED. Configure gifsicle's interlaced option in it's own options. (gifsicle.interlaced)
DEPRECATED. Configure optipng's optimizationLevel option in it's own options.
```

Current:
```
module.exports = {
  progressive: true,
  optimizationLevel: 7,
  interlaced: false,
  pngquant: {
    quality: '65-90',
    speed: 4,
  },
};
```
should be:
```
module.exports = {
  mozjpeg: {
    progressive: true,
  },
  optipng: {
    optimizationLevel: 7,
  },
  gifsicle: {
    interlaced: false,
  },
  pngquant: {
    quality: '65-90',
    speed: 4,
  },
};
```

tomazy:
@nhducit I guess this one: https://github.com/tomazy/react-boilerplate/tree/ssr

gihrig:
@nhducit I believe the errors you are seeing are due to updating `image-webpack-loader` to 3.x see #1561.

As such, your fix is welcome, but it is not specifically related to this PR. Please open a separate PR for this against `dev` branch.

If your project is using `image-webpack-loader` 2.x please advise.

Thanks for tracking this down 😄

howardya:
The latest ssr branch by @tomazy works like a charm with the master!

nhducit:
@FaustoC Do you have any problem with rendering server? I test my application with https://artillery.io/ and found that ther rendering server consumes a lot RAM. 
- Have to try to test how many concurrent user a server can serve?

FaustoC:
@nhducit The process itself fills ~30MB of Ram.
As for the concurrent users, i dunno, we have from 2 to 3 request per minute going to the server, the rest is all serve from the CDN (~50 per minute).
As for the GA ecc... in HtmlDocument, it's an actual html file rendered from js, add your script tags normally at the end of the body.

nhducit:
@FaustoC 
- I use tool to test my local server (running in production mode) (start: 70MB). after hundreds requests, the rendering server consumes a lot ram and it take a long time to reduces to ~100MB. I guess I have memory leak here (30MB). I will research deeper when I have time.
- I figure out how to deal with GA and facebook stuff.

FaustoC:
@nhducit i'm going to load test the dev server without cdn in a day or two to prepare for another ssr project, i'll try to gather data to see if i've memory problems too

nhducit:
@FaustoC if you have a staging server you can use this tool to perform stress tests https://artillery.io/

TheRusskiy:
Anyone has the following problem?
Any time something is changed inside the app I am getting (ok, not any time, but 90% of times)
```
{ Error: socket hang up
    at createHangUpError (_http_client.js:254:15)
    at Socket.socketOnEnd (_http_client.js:346:23)
    at emitNone (events.js:91:20)
    at Socket.emit (events.js:185:7)
    at endReadableNT (_stream_readable.js:974:12)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
    at process._tickCallback (internal/process/next_tick.js:98:9) code: 'ECONNRESET' }
```

And in the browser:

> Proxying failed for page rendering service. The service maybe restarting so try again in a second. Check the console for more information.

This pretty much eliminates hot reloading.

May be we can run both the main server and the rendering server in one script (in dev)? Gonna make it easier to wait for server to start.

TheRusskiy:
I think it was a false alarm, I did not realise I was changing `React.PureComponent` which caused full page reload, if you change `React.component` there isn't such an issue.

TheRusskiy:
I am not 100% sure how good of an idea that is but what I do now for the case when SSR server hasn't reloaded yet is just reloading the page automatically.
Here's the pull request: https://github.com/tomazy/react-boilerplate/pull/2

FaustoC:
@tomazy is it possible to render just the helment on the server, and leave the whole page loading "normally" in the browser?

nhducit:
@FaustoC I think you can hack it by detect server environment and only put necessary data.

serverEntry.js
```
  const doc = renderToStaticMarkup(
    <HtmlDocument
      --appMarkup={appMarkup}
      lang={state.language.locale}
      --state={state}
      head={Helmet.rewind()}
      --assets={assets}
      css={css}
     -- webpackDllNames={webpackDllNames}
    />
  );
```

arthurCormack:
[at] anyone-with-a-clue: A question about using socket.io in SSR/client universal app. 
I'm using redux-socket.io, which was fairly straightforward to get working with react boilerplate (non SSR). I basically just added it to the middleWares, inside of store.js, like so: 
```const middlewares = [
    sagaMiddleware,
    socketIoMiddleware,
    routerMiddleware(history),
  ];```
but (predictably) ... there are errors on the server: `TypeError: Cannot read property 'emit' of undefined`.
So, my question is, what might the best approach be towards including a socket.io client in a  Universal React Boilerplate based app?

**Comment by declanelcocks** [comment]
Potential issue when trying to import `antd` as a library:

```
Dynamic page loading failed .../node_modules/antd/lib/style/index.less:1
(function (exports, require, module, __filename, __dirname) { @import "./themes/default";
                                                              ^
SyntaxError: Invalid or unexpected token
    at createScript (vm.js:53:10)
    at Object.runInThisContext (vm.js:95:10)
    at Module._compile (module.js:543:28)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (.../node_modules/antd/lib/button/style/index.js:3:1)
(node:24742) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): SyntaxError: Invalid or unexpected token
(node:24742) DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

I followed the example solution from https://github.com/react-boilerplate/react-boilerplate/issues/755 to add `antd` as a UI library to the project. On the `master` branch it works fine, but if I try on the `ssr` branch it throws this error. I also added the `webpack` settings to the `webpack.ssr` file but no luck.

For anyone wanting to use a UI library it would be good to get some recommendations on how to do this with SSR.

**Comment by lichenglu** [comment]
Hi, @FaustoC 

Do you mind sharing more on this approach?

> For window, document and so on, i just modified my code, if i'm in server enviroment, i don't use window, ga, ecc...

Thank you!

**Comment by nhducit** [comment]
@lichenglu you can detect server or browser via environment variable. define something like `process.env.IS_BROWSER`

**Comment by TheRusskiy** [comment]
Try "process.browser"

**Comment by FaustoC** [comment]
@lichenglu exatly as nhducit said, check if `typeof window !== 'undefined'` or find other suitable ways to check if you are in browser env. (we are using the simple check above, but it can surely be optimized)

**Comment by arthurCormack** [comment]
@tomazy,
I have a question about the the SSR and CSS. 
I have noticed that there are tests for .css in the webpack.base.babel.js but not in the webpack.ssr.babel.js

I have to import some plain .css in some of my components, and, as a result, I'm seeing a whole quagmire of errors, about those css imports when the SSR attempts to build.

After studying the code (for a few days!), I've realized that there is a css attribute that is passed into the <HtmlDocument /> component, which comes from styleSheet.rules ( from 'styled-components/lib/models/StyleSheet'). So, I think that this collects all of the rules that have been created with styled, and passes them on to get rendered out. 

What happens to the styles that weren't made with styled; that were imported from components, or from libraries in node_modules? 
How could we adjust the webpack.ssr.babel.js in order to parse properly and include those other .css files. 
Thanks again, and best regards.

**Comment by arthurCormack** [comment]
Sorry to be a pest, but I think that the global-styles might be getting embedded twice. I think it appears both in the serverEntry.js and the app.js. I can see the same two <style> tags appearing in the <head>.

**Comment by tomazy** [comment]
@arthurCormack The project already uses plain css - search for `sanitize.css`.

> I think that the global-styles might be getting embedded twice: I can see the same two (global-styles) style tags appearing in the head.

I'm aware of that. Unfortunately I don't know a good solution for that. I guess [styled-jsx](https://github.com/zeit/styled-jsx) does something to prevent this.
This is what happens: on the server we will render all the style tags that are necessary for the page (**once**), but once the page is loaded in the browser the styled components library will inject them again :( I don't follow `styled-components` and I don't know if they have solved it (looks that the issue https://github.com/styled-components/styled-components/issues/124 is still open)

Do you have any issue with this double style tags? Does styling in your app get broken because of that? I'd like to know if there are real world side effects of it.

**Comment by wmertens** [comment]
Render the server styles to a style tag with id, and then just before
rendering the app on the client side, remove the server css.

**Comment by arthurCormack** [comment]
@tomazy, re: would there be an issue with double style tags: It depends on what goes in the global styles. If there isn't much in it, then its not a problem. 
But if it's a lot of .css, it's expensive having to load it twice. 

In my case I'm working with designers that haven't quite grokked the while notion of localized styles, and want to easily port their work from older paradigms: wordpress, drupal, etc. 

But Its not really a show-stopper for me. It's easy enough to side-step the issue by wrapping the megalithic css in an App level styled component ... in effect namespacing it, but at the top level. 
It works. 

But I like @wmertens of doing a switcheroo, using the id of the global style tag, when the client renders.

**Comment by arthurCormack** [comment]
Is there a way that we could pass server variables, such as the hostname that a request was made to, from the server, into a variable that would be accessible to react/redux/sagas?

In my current setup ( not SSR), I have the react-boilerplate SSR sitting behind an nginx proxy, which sends requests that begin with `/wp-admin` or `/wp-json` (an api) to an upstream LAMP/Wordpress server, and all other requests to the node/SSR app server.  It cleanly handles CORS issues by having the API and the app making request to that API share the same domain. 
And it works on my local dev environment ( I have NGINX, and WP Server running locally on my Mac) , my staging server, and my production server, because I use relative URLS when I define the API URLs.

But when I try to make it work with Server Side Rendering, the sagas that are making asynchronous calls require an absolute URL, not a relative URL! What I'd like to do is establish a baseURL, which would be determined, either by Node, or by some argv or process.env ...  
and then construct an absolute URL by concatenating the relative URLs of the API endpoints to that baseURL.

But how to do this ... any ideas? I've made some attempts with `new webpack.DefinePlugin`, but so far, the correct approach eludes me.

Best Regards.

**Comment by tomazy** [comment]
@arthurCormack try adding your variables [here](https://github.com/react-boilerplate/react-boilerplate/pull/1236/files#diff-785338b656a733ff4ae5a2938dffb9feR22) and then you could put them somewhere into your app state [here](https://github.com/react-boilerplate/react-boilerplate/pull/1236/files#diff-de02853908e1278058760e3cde3d71ffR81).

**Comment by nvartolomei** [comment]
In case this is merged, can we have it under a feature flag?

At first I was all in on ssr, but unfortunately sometimes this has a high cpu footprint and response times together with user experience can degrade.

**Comment by arthurCormack** [comment]
@nvartolomei regarding SSR and performance, etc:
It's an unrealistic expectation that there wouldn't be significantly higher CPU / memory, time demands on the SSR server: It's doing a lot more work, the demands placed upon it are significantly higher. But the payoff is so worth it imho, in terms of SEO, link authority, and user experience.

Making things fast is going to require some additional elements and strategies. Caching is key. And a smart strategy  that determines what, when and where to cache is essential if you want to get things blazingly fast. Why would you want to make your SSR server render the exact same stuff over and over again?  Separate the stuff that is not cacheable from the stuff that is, and handle them separately.

**Comment by arthurCormack** [comment]
@tomazy 
`Unexpected property "home" found in previous state received by the reducer. Expected to find one of the known reducer property names instead: "route", "global", "language". Unexpected properties will be ignored.`
I'm seeing this error in the console when I run fresh builds of pull/1236/head.

**Comment by nhducit** [comment]
@arthurCormack https://github.com/react-boilerplate/react-boilerplate/pull/1236#issuecomment-269594460

**Comment by tomazy** [comment]
@arthurCormack see also [this comment](https://github.com/react-boilerplate/react-boilerplate/pull/1236#issuecomment-274404770)

@nhducit I really appreciate your help with answering the questions in this PR! Thank you!

**Comment by arthurCormack** [comment]
Righto - apologies for asking a question that had already been answered.

**Comment by TheRusskiy** [comment]
to me it seems like SSR server is not really starting in parallel, compilation for it happens AFTER web server has compiled

**Comment by nvartolomei** [comment]
@arthurCormack you are thinking in terms of mostly static sites. I think of highly dynamic websites with lots of data, which not necessarily need to be indexed by search engines, think Facebook, Twitter, SaaS projects. I'd rather ship partial UI and progressively load data to user, making parts of application available to user than block on server, spend resources and ship complete UI too late.

Because there are lot of different use cases and requirements, I proposed to have this one under a flag. Though this change will not affect me directly as I always can customize it for my specific needs.

**Comment by acanimal** [comment]
@tomazy @mxstbr Any plan to merge this PR??

**Comment by DeividasK** [comment]
Hey all,

I am trying out this PR and am running into problems with sagas. The sagas seem to start on the server (even if I check for window and window.document objects), so I get errors. Also, I suspect that my sagas are not working on the browser either.

Anyone encountered any of these issues?

**Comment by nhducit** [comment]
@DeividasK Could you provide example repo? I don't have any issues with saga when running this PR in production

**Comment by DeividasK** [comment]
@nhducit it turns out it was me not being able to see that I hooked up the wrong saga. Sorry for raising the alarm.

**Comment by arthurCormack** [comment]
I think that we have a problem with the server-side `Helmet` rendering. It seems as though the `<Helmet>` in `<App>` is what gets rendered, and not the `<Helmet>` in the route containers (which is supposed to override the default in `<App>`. Why does this matter? Because of Google, Facebook and Twitter, for starters. 
With Facebook, when we go to share a link on our wall/timeline, facebook scrapes that url, and parses meta tags, like og:title, og:description, og:url, og:image. But it will only pick up what the server initially renders. So ... all pages, will wind up having the same values for these tags (from the default `<Helmet>` in `<App>`), instead of their own specific values. The same kind of problem affects twitter and google. 

Am I correct about this diagnosis? If so ... what approach would you recommend? I've read the Helmet docs, but haven't found much about SSR. Has anybody out there encountered this problem?
I've thought about connecting the `App` to state, and plucking values for title, description, image out of there, after having set those values by the route reducers that run sagas, and retrieve content data from an API.  I'm hoping that there might be an easier and tidier way to make Helmet render, after all the sagas have run, and the route containers have rendered.

**Comment by nhducit** [comment]
@arthurCormack I'm using this branch for production site. 
I don't understand what is your problem with `React Helmet`? Could you provide more information (sample repo is great)? 
Here are my website. https://www.carousellmotors.com/. I don't any problems with facebook or google crawler. Please help to check if I get the same problem with you or not.

**Comment by tomazy** [comment]
@nhducit which part of your website uses SSR? The root url doesn't seem to use it when visited with js disabled.

**Comment by nhducit** [comment]
I'm sorry. I only enablre ssr for crawler only. @tomazy use chrome
extension to change the useragent to googlebot ỏ facrbook

**Comment by arthurCormack** [comment]
Here is the site that I'm currently on: http://radiospa-stg.gm333zxxti.us-east-1.elasticbeanstalk.com/
If you try clicking a link to a post, and inspecting the meta tags in the `<head>` of the source that is originally sent by the server ( not rendered client-side ) and comparing it to the client-side rendered `<head>` you'll see that they are not the same. The `<helmet>` stuff in the route container does not seem to override the `<helmet>` stuff from `<App>`.

**Comment by arthurCormack** [comment]
Guys, I'm sorry for the noise about the Helmet stuff. I think it's a problem that is particular to me and my codebase, and not the SSR branch. I've conducted some experiments agains a vanilla version of the branch, and I can't reproduce the problem there. Apologies.

**Comment by DeividasK** [comment]
Ok, *I think* I have found a bug. On every refresh of the page, the new items is added to the browser history. This happens on my project as well as on a fresh install of this pull. Steps to reproduce the bug:

- git clone git@github.com:react-boilerplate/react-boilerplate.git
- cd react-boilerplate
- git fetch origin pull/1236/head:ssr
- git checkout ssr
- npm install
- npm start

Navigate to http://localhost:3000. You should see an extra item added to the browser history. Additional history items are added correctly when navigating the page. However, when you refresh the page a new item of a current page is added.

**Comment by JuHwon** [comment]
first of all: awesome work!

I migrated this PR in my project yesterday. Im still not 100% happy though and mby you guys can help me out here.

1. I'd like to render things differently when it is rendered by the server. Just for the sake of feasibility  i implemented something like this in the HomePage: 
```javascript
  renderWelcomeMessage = () => {
    if (typeof window === 'undefined') {
      return <p>Hello from the Server.</p>;
    }
    return <p>Hello from the Client.</p>;
  }
```
Sadly the client rerenders the whole page at the beginning which results in the following behaviour:
apr-11-2017 12-19-56

2. I normally fetch the required data (by dispatching an action) of a container in componentWillMount. So i actually refetch everything when the site arrives at the client.

Are there any solutions/best practices to avoid such a behaviour? Any ideas/thoughts?

TheRusskiy:
@JuHwon 
I cache many things in redux store and as a side effect it eliminates fetching data second when I am on client after server rendering.
So I would have something like this:
```
HomePage:
componentWillMount() {
    this.props.loadPage('home');
  }
...
HomePage/sagas:
function* pageSaga() {
  let reviews = yield select((state) => state.getIn(['reviews', 'home']));
  if (reviews) {
    reviews = reviews.toJS();
  } else {
    const response = yield call(routes.reviews.index);
    reviews = requestResult.data.reviews;
  }
  yield put(pageLoaded('home', { reviews }));
}

export function* pageFlow() {
  yield takeEvery((action) => action.type === LOAD_PAGE && action.page === 'home', pageSaga);
}
```
In response to `pageLoaded` action with `home` reducers would save reviews in `'reviews' -> 'home'`.
I omitted error handling from this example.

pruhstal:
This PR ever gonna get merged? ;)

JuHwon:
thanks for the reply @TheRusskiy 

i implemented my own SsrProvider with a HOC to achieve my goals. 
Now i got 2 props ( `isServer` and `isServerRendered` ) injected. So i can do specific renderings when the page is server rendered.
```javascript
  renderWelcomeMessage = () => {
    if (this.props.isServerRendered) {
      return <p>Hello from the Server.</p>;
    }
    return <p>Hello from the Client.</p>;
  }
```

Also now i can have one class method which only gets called on the server and not again on the client when it is a server rendered page. 
```javascript
  static loadInitialData = ({ props }) => {
    ...
  }
```

mby this helps anyone who is having the same issue ;)

DeividasK:
For anyone who has encountered [the same issue](https://github.com/react-boilerplate/react-boilerplate/pull/1236#issuecomment-292966086) as me, the fix is to add `adjustUrlOnReplay: false` to the options of `syncHistoryWithStore` method.

`app/setup/syncHistoryWithStore.js`

```js
import { syncHistoryWithStore } from 'react-router-redux';
import { makeSelectLocationState } from 'containers/App/selectors';

// Sync history and store, as the react-router-redux reducer
// is under the non-default key ("routing"), makeSelectLocationState
// must be provided for resolving how to retrieve the "route" in the state
export default function syncHistory(history, store) {
  return syncHistoryWithStore(history, store, {
    selectLocationState: makeSelectLocationState(),
    adjustUrlOnReplay: false, // <-- change this to false, as it is true by default
  });
}

**Comment by designbyadrian** [comment]
@tomazy Which one of your react-boilerplate branches should I checkout if I want to use your awesome updates early?

**Comment by nhducit** [comment]
@designbyadrian you can use this pull request

**Comment by tomazy** [comment]
@designbyadrian yeah, here the right place.

**Comment by designbyadrian** [comment]
I can't access the pull request. Could you link me your branch, please?

*Adrian von Gegerfelt*
Design by Adrian

**Comment by nhducit** [comment]
@designbyadrian http://stackoverflow.com/questions/27567846/how-can-i-check-out-a-github-pull-request

**Comment by oyeanuj** [comment]
Hey @tomazy @gihrig ! Just wanted to get a sense from you for when do you see this being merged and what parts need to happen before the merge? Thanks for all the work!

**Comment by gihrig** [comment]
> when do you see this being merged

Needs input and approval from community [maintainers](https://github.com/react-boilerplate/react-boilerplate/blob/master/MAINTAINERS).

>what parts need to happen before the merge

A decision needs to be made as to merging SSR into the main project (and writing docs on how to remove it if not wanted), or to maintain as a separate branch.

**Comment by davidjbradshaw** [comment]
@gihrig would it be possible to just have two starting points for the app?

- index.html for non universal
- index.js for universal

The people can choose and even later switch between the two.

**Comment by gihrig** [comment]
@davidjbradshaw That's an interesting idea.

I would prefer to see the distinction made in the build process. e.g `npm run build` vs `npm run build:ssr` or some such thing. 

The plain `build` would have to strip out the ssr code, rendering pretty much the same app it does now. I'm not sufficiently familiar with the webpack build processes to guess at how much work that would involve, but I don't think we would want to ship all the ssr code to production when it was not going to be used.

**Comment by nhducit** [comment]
@gihrig I'm using this PR for production website but I only use SSR for crawler (google, facebook,...). I generate index.html file and then use nginx to detect user-agent:
- crawler => SSR
- normal browser => serve index.html file

I added HtmlWebpackPlugin to webpack.prod file
```
    // Minify and optimize the index.html
    new HtmlWebpackPlugin({
      template: 'app/index.html.ejs',
      minify: {
        removeComments: true,
        collapseWhitespace: true,
        removeRedundantAttributes: true,
        useShortDoctype: true,
        removeEmptyAttributes: true,
        removeStyleLinkTypeAttributes: true,
        keepClosingSlash: true,
        minifyJS: true,
        minifyCSS: true,
        minifyURLs: true,
      },
      inject: true,
    }),
```

**Comment by gihrig** [comment]
@nhducit That looks great, thanks for sharing 👍 

> I generate index.html file ...

This is a static index.html, a copy of that generated by SSR?

**Comment by nhducit** [comment]
@gihrig in original PR by @tomazy, he removed HtmlWebpackPlugin, that's mean index.html file is not generated when run `npm run build`,  in favor of `HtmlDocument` component, this file is used for SSR.
But in my use case, I need to generate `index.html` file that's why I readd `HtmlWebpackPlugin ` plugin. 

my generated index.html file.
```
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="utf-8">
  <meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1,user-scalable=0">
  <meta name="mobile-web-app-capable" content="yes">
  <link rel="shortcut icon" type="image/png" href="https://dhyaco2i7c70y.cloudfront.net/favicons/favicon_32x32.png" />
  <title>Carousell Motors</title>
  <link href="/vendor.d9480866ae378ddd957a3327065511b0.css" rel="stylesheet">
</head>
<body>
<noscript>If you're seeing this message, that means <strong>JavaScript has been disabled on your browser</strong>,
  please <strong>enable JS</strong> to make this app work.
</noscript>
<div id="app"></div>
<link href="https://fonts.googleapis.com/css?family=Source+Sans+Pro" rel="stylesheet">
<script type="text/javascript" src="/main.483217690189c5adb125.js"></script>
</body>
</html>
```

**Comment by gihrig** [comment]
@nhducit 

>=> If we don't want to use SSR for production. We just need to add HtmlWebpackPlugin.
>=> But the SSR feature is still there when we developing. I haven't tried to disable it, because I need SSR for my project

Thank you for clarifying that ✨

**Comment by nhducit** [comment]
But there are a problem I havn't resolved yet that is `generate.serverEntry` have source map file. But it doesn't work for me on console.
example: My code has an error on App.js:12 file. command line show there is an error at serverEntry:32131:123 instead of App.js 12.

**Comment by gihrig** [comment]
@nhducit Have seen #1559 ?

**Comment by tgrant59** [comment]
Love the work so far, can't wait till this gets merged. I'm starting to use this branch on a new project. I've found that `npm run test:watch` is broken, both before and after running `npm run clean`

Error Message:
```
node[3588] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-22)
2017-05-20 17:01 node[3588] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-22)
2017-05-20 17:01 node[3588] (FSEvents.framework) FSEventStreamStart: register_with_server: ERROR: f2d_register_rpc() => (null) (-22)
events.js:163
      throw er; // Unhandled 'error' event
      ^

Error: Error watching file for changes: EMFILE
    at exports._errnoException (util.js:1033:11)
    at FSEvent.FSWatcher._handle.onchange (fs.js:1375:11)
```

Should be reproducible just by cloning this branch, running `npm run setup` and giving it a shot

**Comment by tgrant59** [comment]
Also, there is still a font flicker when refreshing the application. It's got to have something to do with the loading of google fonts

**Comment by davidjbradshaw** [comment]
@gargol So I guess the best solution here is to have the option to start the dev server in SSR and normal modes with `npm start` and `npm run start:ssr`. 

Plus build options to build standard client mode (`build`), SSR (`build:ssr`) and mixed mode (`build:hybrid`).

In the normal mode I expect that you'll end up with a tiny amount SSR code, but not enough to worry about.

**Comment by Proxiweb** [comment]
How do you load datas from api to get filled server side rendered page ? I tried to start a saga but I haven't found the way to make this work...

**Comment by nhducit** [comment]
@Proxiweb where do you call an action to start saga? I call actions in `componentWillMount`. note that in server react component *don't* trigger `componentDidMount`

**Comment by RayProud** [comment]
@tomazy can you help (or tell me about) with 
```
Unexpected property "home" found in previous state received by the reducer. Expected to find one of the known reducer property names instead: "route", "global", "language". Unexpected properties will be ignored.
(anonymous) @ combineReducers.js:29
computeNextEntry @ VM269:2
recomputeStates @ VM269:2
(anonymous) @ VM269:2
dispatch @ createStore.js:171
createStore @ createStore.js:245
(anonymous) @ VM269:3
(anonymous) @ VM269:3
(anonymous) @ VM269:2
(anonymous) @ applyMiddleware.js:31
(anonymous) @ VM269:2
createStore @ createStore.js:57
configureStore @ store.js?f39c:35
(anonymous) @ app.js?86a9:54
./app/app.js @ main.js:791
__webpack_require__ @ main.js:687
fn @ main.js:106
27 @ main.js:2696
__webpack_require__ @ main.js:687
(anonymous) @ main.js:782
(anonymous) @ main.js:785
combineReducers.js:29 Unexpected property "home" found in previous state received by the reducer. Expected to find one of the known reducer property names instead: "route", "global", "language". Unexpected properties will be ignored.
(anonymous) @ combineReducers.js:29
computeNextEntry @ VM269:2
recomputeStates @ VM269:2
(anonymous) @ VM269:2
dispatch @ createStore.js:171
dispatch @ VM269:2
(anonymous) @ middleware.js:22
(anonymous) @ middleware.js:73
dispatch @ VM269:2
handleLocationChange @ sync.js:99
syncHistoryWithStore @ sync.js:108
syncHistory @ syncHistoryWithStore.js?e172:8
(anonymous) @ app.js?86a9:58
./app/app.js @ main.js:791
__webpack_require__ @ main.js:687
fn @ main.js:106
27 @ main.js:2696
__webpack_require__ @ main.js:687
(anonymous) @ main.js:782
(anonymous) @ main.js:785
```

I just run your branch and the console showed me this message.

**Comment by nhducit** [comment]
@RayProud https://github.com/reactjs/redux/issues/1994#issuecomment-249916713
because we are using code splitting, reducers are added to store dynamically. This is just a warning, it's ok to ignore it

**Comment by Proxiweb** [comment]
@nhducit I tried to fire sagas in serverEntry's renderHtmlDocument, after sagasDone(). I didn't try in componentDidMount because it's a client side only method. 

Do you know a way to load datas ?

**Comment by RayProud** [comment]
@Proxiweb you can use `componentWillMount()` method.

**Comment by tomazy** [comment]
@Proxiweb have a look how `UserPage` loads the data in this PR.

**Comment by tomazy** [comment]
I've just added a script that will generate static html pages for routes defined in `./static-pages.js`. This happens automatically with `npm run build`. 

@nhducit Could you please see if that would work for you app?

**Comment by nhducit** [comment]
@tomazy cool, I will try it in tonight (my timezone).
FYI: I updated my project to helmet 5.0.0, they renamed `helmet.rewind` to `helmet.renderStatic`

File: internals/templates/setup/ensureIntlSupport.js [hunk 6]
```diff
@@ -0,0 +1,10 @@
+export default function ensureIntlSupport() {
+  if (window.Intl) return Promise.resolve();
+  return new Promise((resolve) => {
+    resolve(System.import('intl'));
```
RayProud: `System.import` is deprecated now, isn't it?
Should we use `import` instead?

https://github.com/webpack/webpack/issues/2163

leifdalan:
I was able to fix this by moving `babel-polyfill` to the webpack entry point rather than _inside_ the entry point.

in `webpack.babel.ssr.js`:
```
  entry: [
    'babel-polyfill',
    path.join(process.cwd(), 'app/serverEntry.js'),
  ],
```
And remove it from `serverEntry.js`.

dnson:
Hi all,
I am using the react-boilerplate with ssr. But I have a problem, when I was trying to import the scss file in the container I got this error `page1.scss doesn't export content`. Please help 
Thank you

kunal-arora:
@RayProud I am making 2 api calls using sagas, And want to render the result of both the calls on server side. How should I approach this?
[you can review the code here.](https://github.com/kunal-arora/pr_react_redux/blob/feature/ssr_v2/app/containers/HomePage/sagas.js)

designbyadrian:
@dnson Your problem is related to how you generate your CSS. Though have you tried just generating one CSS at the start with `ExtractTextPlugin` instead of dynamic loading of styles?

designbyadrian:
@kunal-arora You can dispatch as many actions (to be taken by sagas) as you want in `componentWillMount`

designbyadrian:
Does somebody have an example of a working Dockerfile for an SSR-seasoned react-boilerplate app?

JuHwon:
@designbyadrian whats your issue? [here is a dockerfile i am using.](https://gist.github.com/JuHwon/93c212fb4cd2b8e9970762f5e37ffdff) works for me.

designbyadrian:
@JuHwon Thanks a lot! I'll ponder the similarities to mine.

My issue is that some JS chunks, local images, and the stylesheets *sometimes* are returned as 404. An image src URL which doesn't resolve on the HTML page, can be pasted into a new window and work, and because some JS chunks don't load, the whole site is broken. Works perfectly on localhost.

It looks like the resources are being returned as the 404 page defined by react-router, which makes it harder to determine where the problem actually lies.

designbyadrian:
@tomazy I can't help but notice that SSR isn't started in prod? Excuse my ignorance, but could you explain why? I'm looking for a solution to have webpages rendered for the purpose of SEO.

kunal-arora:
@designbyadrian Hey thanks for replying, I cannot dispatch that action from `componentWillMount` since I am getting the data from the first action and passing it to the other action.
If possible please review the code [here](https://github.com/kunal-arora/pr_react_redux/blob/master/app/containers/HomePage/sagas.js) and inform me if I am doing anything wrong.

designbyadrian:
@kunal-arora If you need to call two actions in a row every time, why not have just one saga that makes those two API calls in a row?

RWOverdijk:
Yes please. But for now, something like [roast](https://www.roast.io) works fine, too. https://gaping-dad-7012.roast.io/features

kunal-arora:
@designbyadrian I do not have to call those two actions in a row every time , I just need to call those two actions in a row on page load only.
For example -
The first saga loads up all the categories and the later one loads the products of the category id passed in the action.

designbyadrian:
@kunal-arora Yes, so create a new saga that does those two, and one unique action that's only fired at start up.

kunal-arora:
@designbyadrian Thanks for that. It seems like thats the best solution right now.

tomazy:
@designbyadrian 
> @tomazy I can't help but notice that SSR isn't started in prod? Excuse my ignorance, but could you explain why? I'm looking for a solution to have webpages rendered for the purpose of SEO.

In production mode the SSR works directly from the server process. During development, however, we have this additional service to support hot/live reloading. Can you test your production with JS disabled?

darkenvy:
When viewing the SSR version of the boilerplate on :3001, (index), main.js, reactBoilerplateDeps.dll.js and vendor.[xyz].css all share the same HTML. ie: the files don't load correctly. 

This is with no changes to the code and running npm run start. Ideas on whats happening here?

jahglow:
@tomazy Hi, I was using your SSR solution and one thing I can't understand is why it doesn't render the App component contents to string in my case, even though it fetches data from sagas for a correct route and puts it in the SSR html string. What it generally does is creates two react-empty tags instead of Helmet and AppFrame for some reason (I mean these two inside a div in [this file](https://github.com/jahglow/eurobloha/blob/firebaseSSR/app/containers/App/index.js) ). Can you take a look and say where I went wrong?
image

designbyadrian:
@tomazy Sorry for the delay. Yes, SSR works properly on production, as you said!

matteocng:
@darkenvy you should not use `document.location.href` (it reloads the page, bypassing `React & Redux`). When you need to change the location programmatically, you can use the `router` object, which is in `this.props` almost everywhere excluding `<Header />` and `<Footer />` (if you need it there => `<Header router={props.router} />` in `containers/App/index.js`):
 
   ```js
   this.props.router.push('/some/path')
   ```
When you need to trigger a navigation event outside a component you might use [the `browserHistory` singleton](https://github.com/ReactTraining/react-router/blob/v3/docs/guides/NavigatingOutsideOfComponents.md) after checking that it is not `falsy`, but personally I would avoid doing this whenever possible in an `SSR` scenario.

   ```js
   import browserHistory from 'react-router/lib/browserHistory';
   
   if(browserHistory) {
      // Go to /some/path.
      browserHistory.push('/some/path')

      // Go back to previous location.
      browserHistory.goBack()
   }
   ```

justingreenberg:
unfortunately, this will not be landing :( the consensus amongst users and maintainers was that RBP will move forward as a baseline for "for creating large, high-performance enterprise-grade frontend applications with advanced/custom requirements, statically delivered by CDN." please see #1776 for discussion and vote results

@tomazy would you be interested in maintaining an SSR fork? if so, please feel free to create a PR with link to your fork in documentation

on behalf of all RBP users and maintainers, thank you sincerely for your time and effort 🙏  i will leave this PR open so we have a place to discuss any plans to maintain an SSR fork

mathieumg:
Thanks @justingreenberg for letting us know, even though this is very disappointing. Given that the poll was very close and had a low sample size, it would have been appreciated if a link to it was posted here as well as in #1746 to try to gather a larger sample size, as we don't receive backpings by email. This must also be frustrating for @tomazy, who has constantly been syncing up with master since last Fall. :disappointed:

hadifarnoud:
I agree with @mathieumg here. we did not have a chance to vote on this.

kunal-arora:
Hey, I am trying to make a responsive app using this PR, I am using [react-responsive](https://github.com/contra/react-responsive/) for loading my responsive components. Everything seems to be working right after the first render. I am trying to fix the first render and the solution for that seems to be in [this](https://github.com/contra/react-responsive/issues/42), i.e to pass the screen width from the server to the app, but I am having a really hard time implementing that, would love if anyone of you guys can give me some direction on how to do that. I am trying to implement this using a library called [mobile-detect](https://github.com/hgoebl/mobile-detect.js) that gives us the user-agent of the user.

peter-mouland:
hey, @kunal-arora sounds like this should be discussed in a separate issue.

As a quick response, I don't know your use case but I'd recommend against the server using screen-sizes/header/user-agent detection for a number of reasons.  Mainly because it'll mean your server has to do more work than is needed (e.g. is also hard to add caching), but also because the client (i.e. CSS) can be used to do most/all responsive components.  

I've been build responsive websites for (over 6) years and have always been able to do it with CSS (even when team members started using JS I've normally been able to refactor the JS out of it and just use CSS for better perf).

Sorry for not really answering your question!

wmertens:
@kunal-arora how about https://github.com/Yaska/redux-mediaquery/blob/master/README.md#server-side-rendering ?

@peter-mouland while you are right that in many cases, CSS is better, sometimes you want to render completely different components on mobile vs desktop, and then CSS would only be able to hide parts instead of only rendering the needful.

kunal-arora:
@wmertens thanks for the resource, I will look into it. 

Thanks for replying @peter-mouland, as @wmertens stated there are few cases where I want to render a completely different component for mobile and desktop but not enough to make totally different apps for mobile and desktop.

kunal-arora:
Hey @wmertens , the link you provided is doing the same thing using action providers. For ssr they say `If you combine this with a library like https://github.com/hgoebl/mobile-detect.js, you can decide on the server whether to render for a phone or for desktop, or even for search bots.`

So my problem is how to do that, How to get the screen size using plain js or to get user-agent using mobile detect from server and pass it on to the app. I know few things like we can pass it in the render() function in app.js and then it will be available to htmlDocument component, but that wont give us the screen size or user-agent of each user I think.

wmertens:
(Last reply of mine in this thread, this should go in another issue
somewhere or on stackoverflow.)

With redux-mediaquery you get the screen size/type using redux @connect. So
you can just subscribe the components to e.g.  isPhone (client and SSR) or
innerWidth (client only)

hadifarnoud:
we're trying to use your fork @tomazy. would you please write a small tutorial on how it works?

tomazy:
@hadifarnoud see: https://github.com/tomazy/react-boilerplate/blob/68360c446a065fc2d8976c7bf50a2bddf0bd4a24/docs/general/server-side-rendering.md

hadifarnoud:
thank you very much @tomazy. this is more of a description on how things work. is there any instructions for ssr? I mean is there anything else we should do (other than what boilerplate readme has)

kunal-arora:
Hi @tomazy , I am trying to make a responsive layout using your ssr branch. For that I am using [react-responsive-redux](https://github.com/modosc/react-responsive-redux), In their docs they mention that we have to [Add mobile detection and dispatch a redux action during your request handler](https://github.com/modosc/react-responsive-redux#es6-example-1), But in your ssr branch's [serverEntry.js](https://github.com/tomazy/react-boilerplate/blob/ssr/app/serverEntry.js) file there is no 'request' parameter to use for mobile detection. Please help me with this, I am kinda stuck here. Thanks!

nhducit:
@kunal-arora I think you should look for `req` in this file https://github.com/tomazy/react-boilerplate/blob/ssr/server/middlewares/handleSSR.js

kunal-arora:
thankyou @nhducit I am looking into it.

kunal-arora:
Hey @nhducit , For ssr with responsive layout, I am using  [react-responsive-redux](https://github.com/modosc/react-responsive-redux) in which the whole Implementation of media queries is done based on a variable in state called [fakeWidth](https://github.com/modosc/react-responsive-redux/blob/master/lib/components.js) . To get it working according to the docs we need to [dispatch a redux action during request handler](https://github.com/modosc/react-responsive-redux#server-setup), In my project I do that in the [serverEntry.js](https://github.com/gitserverhu/ustraa_v2_react/blob/features/mobile-detect/v1.0/app/serverEntry.js) file.

Everything works as expected and the state is added to the store, But when I use the prebuilt components of the package this error shows up 
`TypeError: Cannot read property 'fakeWidth' of undefined
    at Function.mapToProps (/var/www/html/ustraa_v2.1/node_modules/react-responsive-redux/lib/components.js:62:50)`

I think this comes up because there is no state available to the components when `renderAppToString(store, renderProps);` is being called.

oyeanuj:
@kunal-arora For SSR support, I've recommend https://github.com/ctrlplusb/react-universally, probably the most solid starter kit I've found.

kunal-arora:
Hey @tomazy , I need some help figuring this out,

Like I mentioned earlier I am using  [react-responsive-redux](https://github.com/modosc/react-responsive-redux) in which the whole Implementation of media queries is done based on a variable in state called [fakeWidth](https://github.com/modosc/react-responsive-redux/blob/master/lib/components.js) . To get it working according to the docs we need to [dispatch a redux action during request handler](https://github.com/modosc/react-responsive-redux#server-setup), In my project I do that in the [serverEntry.js](https://github.com/gitserverhu/ustraa_v2_react/blob/features/mobile-detect/v1.0/app/serverEntry.js) file.

Now I am using [selectors.js](https://github.com/gitserverhu/ustraa_v2_react/blob/features/mobile-detect/v1.1/app/containers/App/selectors.js) to get the fakeWidth that is present in the state. But it always throws an error `responsive.get is not a function`. I am clueless why this is happening.

nhducit:
@kunal-arora i'm sorry, I haven't use  `react-responsive-redux`. Implement and maintain SSR is really hard. I think you should consider to use other boilerplates or project like: `next.js`

kunal-arora:
@nhducit true that. But my issue is pretty general and I think someone with more experience should be able to provide a solution to that.

nhducit:
@kunal-arora why don't you post this question on stackoverflow?

kunal-arora:
yes I did that.

philipcressler:
@tomazy When I pull your SSR branch, I'm getting the following errors on page load:

screen shot 2017-10-13 at 1 22 17 pm

Any idea how to fix? Thank you for your effort in creating an SSR branch of this great boilerplate.

nhducit:
@philipcressler https://github.com/react-boilerplate/react-boilerplate/issues/1901

kunal-arora:
Hello Guys,
I am using dynamic params to load my component, which looks something like [this](https://github.com/gitserverhu/ustraa_v2_react/blob/feature/mobileCategoryPage/v1.6/app/routes.js#L47)
```javascript
{
      path: '/:urlKey',
      getComponent(nextState, cb) {
        if (nextState.location.pathname === previousPath) {
          return;
        }
        const importModules = Promise.all([
          import('containers/CatalogPage/reducer'),
          import('containers/CatalogPage/sagas'),
          import('containers/CatalogPage'),
        ]);

        const renderRoute = loadModule(cb);

        importModules.then(([reducer, sagas, component]) => {
          injectReducer('catalogPage', reducer.default);
          injectSagas(sagas.default);

          renderRoute(component);
        });

        previousPath = nextState.location.pathname;
      },
 }
```
So when I use `<Link to="/all" />` then the response comes and it renders smoothly but from `<Link to="/all" />` when I go to `<Link to="/justone" />` nothing happens.  Any Help for this ?

juciccio:
Hey guys, does someone have an example of how to add server side rendering? I have a project made with version 3.4 and I need server side meta tags for Facebook and Twitter sharing.

hadifarnoud:
you're not alone @juciccio. if anyone writes a comprehensive tutorial, that would be awesome!

davidjbradshaw:
@juciccio & @hadifarnoud I expect the easiest solution would be to create a new project based on this branch, then lift and shift your project code into it.

The other option would be to merge this branch into your project, but I expect you will have a lot of conflicts to resolve in code that your not familiar with.

nhducit:
I integrated this branch to react-boilerplate 3.4 project and this is exact what I did.
```
@juciccio & @hadifarnoud I expect the easiest solution would be to create a new project based on this branch, then lift and shift your project code into it.
```

hadifarnoud:
if a few of us (I'm not a react developer myself but will make sure our dev join the effort) maintains a forked copy of this boilerplate with ssr, that would be much appreciated since @tomazy is not updating it anymore. 

ssr is an important feature for a lot of us. still hope @mxstbr and maintainers of this repo reconsider.

nhducit:
> ssr is an important feature for a lot of us. still hope @mxstbr and maintainers of this repo reconsider.

I agree, but SSR is really a big and complex PR. if we official support it then we need to do benchmarks and you can see we lack of maintainer add SSR even make this project hard to maintain. 
btw `next.js` is really good project and well maintained

ItsMrAkhil:
@nhducit 
I agree SSR is complex. But at least for the first attempt if we can ignore the hot reloading with SSR we may do it. 
Or else let's restrict the SSR for some routes. We can keep a config file where we can have regex. only for those routes activate SSR that to only in production mode. (So no Hot reloading no hot module replacement. No server restart.)
I don't know whether this is a good idea or not. But I think it is possible.

arthurCormack:
I've used this SSR as the foundation of a couple production sites now, and I'm pretty happy with the results so far. http://classicalfm.ca is one of them.  I integrate a wordpress json api, for content, and return appropriate http codes (200, 301, and 404s) from the SSR based upon calls to the content api.

zalabinc:
I tried to combine this project structure in order to run in next.js
I do not do much change, file structure for project is ideal for scalling.
I've used this code structure as the foundation with nextJS,

Work  like charm :+1: 
React Boilerpate + NextJS = :1st_place_medal: 

https://github.com/zalabinc/nextjs-boilerpate/blob/master/app/utils/configureStore.js

ItsMrAkhil:
@zalabinc Good attempt.

hadifarnoud:
it would be great if someone creates a clean fork with latest version of RB merged. there are enough people here willing to put time in maintaining it.

davidjbradshaw:
@hadifarnoud that sounds very close to volunteering to take it on.

hadifarnoud:
I'd love to @davidjbradshaw but I'm not a developer myself. I would've done it long ago if I could

hadifarnoud:
all the ssr tasks in package.json are run in NODE_EVN=development. should I change them to production? why was it set to dev anyway?

tomazy:
@hadifarnoud They are there to support hot reloading etc and you don't need that in production

hadifarnoud:
thanks @tomazy , can you post the recommended production version here please?

tomazy:
@hadifarnoud You just use the default scripts as documented:

`npm run start:production` - this will run all the tests, build the assets and start the server.

If you only care about building the assets before deployment to a target server then run:

`npm run build`

and on the target server:

`npm run start:prod`.

You don't need to do anything special to enable SSR in production. It should just work.

juciccio:
I was using react-boilerplate 3.4 with postcss and now I integrated this branch to my project, everything works fine but now i can't have css in each component the imports are not working anymore. Any suggestion?

kunal-arora:
Hello everyone, I have 2 questions related to routes, my `routes.js` files looks like this
```javascript
// These are the pages you can go to.
// They are all wrapped in the App component, which should contain the navbar etc
// See http://blog.mxstbr.com/2016/01/react-apps-with-pages for more information
// about the code splitting business
import { getAsyncInjectors } from 'utils/asyncInjectors';
import App from 'containers/App';

const errorLoading = (err) => {
  console.error('Dynamic page loading failed', err); // eslint-disable-line no-console
};

const loadModule = (cb) => (componentModule) => {
  cb(null, componentModule.default);
};

function createChildRoutes(store) {
  // create reusable async injectors using getAsyncInjectors factory
  const { injectReducer, injectSagas } = getAsyncInjectors(store); // eslint-disable-line no-unused-vars
  let previousPath = null;
  return [
    {
      path: '/',
      getComponent(nextState, cb) {
        if (nextState.location.pathname === previousPath) {
          return;
        }
        const importModules = Promise.all([
          import('containers/HomePage/reducer'),
          import('containers/App/sagas'),
          import('containers/HomePage/sagas'),
          import('containers/HomePage'),
        ]);

        const renderRoute = loadModule(cb);

        importModules.then(([reducer, appSagas, sagas, component]) => {
          injectReducer('homePage', reducer.default);
          injectSagas([...appSagas.default, ...sagas.default]);

          renderRoute(component);
        });

        previousPath = nextState.location.pathname;
      },
    },
    {
      path: '/:urlKey',
      getComponent(nextState, cb) {
        if (nextState.location.pathname === previousPath) {
          return;
        }
        const importModules = Promise.all([
          import('containers/CatalogPage/actions'),
          import('containers/CatalogPage/reducer'),
          import('containers/App/sagas'),
          import('containers/CatalogPage/sagas'),
          import('containers/CatalogPage'),
        ]);

        const renderRoute = loadModule(cb);

        importModules.then(([actions, reducer, appSagas, sagas, component]) => {
          injectReducer('catalogPage', reducer.default);
          injectSagas([...appSagas.default, ...sagas.default]);
          renderRoute(component);
          store.dispatch(actions.loadPageData(nextState.params.urlKey));
        });

        previousPath = nextState.location.pathname;
      },
    }, {
      path: '*',
      getComponent(nextState, cb) {
        import('containers/NotFoundPage')
          .then(loadModule(cb))
          .catch(errorLoading);
      },
    },
  ];
}

// Set up the router, wrapping all Routes in the App component
export default function createRootRoute(store) {
  return {
    component: App,
    childRoutes: createChildRoutes(store),
  };
}

```
First, `path: '/:urlKey',` this path is getting called even if I am on my homePage i.e `localhost:3000/` and if i console the path it show `/null` or sometimes `/false`. How should I prevent that from happening ? 

Second, how to block ` store.dispatch(actions.loadPageData(nextState.params.urlKey));` on the initial render of the app ?

mparker11:
@tomazy I'm getting an error from `generated.serverEntry.js` that says `ReferenceError: window is not defined`, which refers to line 184 of that generated file. Any idea what might be happening there? It's a bit surprising since that's a global JS object.

gretzky:
Closing this PR because it's gone off track and is out of date. Please feel free to open a new PR based on SSR discussion in #1103

Issue #1237, opened 2016-11-16
[styled-components] Adapting based on props throws ESLint error


KarandikarMihir:
Adapting style w.r.t. props requires to use arrow functions in template strings. Consider this example-

```
export const ProgressBarWrapper = styled.div`
  position: fixed;
  top: 0;
  left: 0;
  width: 100%;
  visibility: ${(props) => props.hidden ? 'hidden' : 'visible'};
  opacity: ${(props) => props.hidden ? '0' : '1'};
  transition: all 400ms;
  z-index: ${(props) => props.hidden ? '-10' : '9999'};
`;
```

This throws following error-
`Arrow function used ambiguously with a conditional expression  no-confusing-arrow`

GGAlanSmithee:
can't check, but if memory serves me right;

`${(props) => (props.hidden ? 'hidden' : 'visible')}`

will get rid of that lint message

samit4me:
This is certainly one of those ESLint rules that can get on your nerves! I usually do exactly what @GGAlanSmithee suggested, but this will require the ESLint option `allowParens` set to true, example:
```
/* eslint no-confusing-arrow: ["error", {allowParens: true}] */

export const ProgressBarWrapper = styled.div`
  ...
  visibility: ${(props) => (props.hidden ? 'hidden' : 'visible')};
  ...
```

Another option is to turn off the rule and then you can leave your code as it is:
`/* eslint no-confusing-arrow: ["off"] */`

If want to apply these rules to your entire project, add it to your `package.json` under the `eslintConfig` section.

Maybe this boilerplate could add one of the above rules globally, I would suggest `["error", {allowParens: true}] `.
What does everyone think???

http://eslint.org/docs/rules/no-confusing-arrow

mxstbr:
Let's just disable that rule because it's a style we encourage from the `styled-components` side, so it shouldn't be an issue in the boilerplate.

KarandikarMihir:
@GGAlanSmithee Your solution worked without setting `{allowParens: true}`. Do we still need to globally disable this rule? cc: @mxstbr @samit4me

samit4me:
It must depend on which ESLint plugin / editor you use because the [documentation](http://eslint.org/docs/rules/no-confusing-arrow) states that the `allowParens` is set to false by default. I use VSCode and if I do not specify the `allowParens: true` it shows errors with parenthesis.

I agree with Max on disabling this rule, it typically causes more pain the good.

[...]

PR #1238, opened 2016-11-16
Disable ESLint rule: no-confusing-arrow

samit4me:
Globally disable the ESLint rule (no-confusing-arrow) as per discussion in #1237


Dattaya:
@mxstbr asked for this change, so I don't even have to think much before giving this a :+1: : LGTM

gihrig:
Agreed, LGTM

Issue #1230, opened 2016-11-14
ESlint recommends against for ... of


gihrig:
After dependency update #1225 many new style errors are flagged. I worked around these by disabling the new rules, but, IMO the code should be corrected and the rules re-enabled.

To reproduce, enable the rules listed below in package.json and run npm run lint.

Error:
```
~/react-bp/app/i18n.js
28:3 error Using 'ForOfStatement' is not allowed no-restricted-syntax
react-bp/internals/templates/i18n.js
24:3 error Using 'ForOfStatement' is not allowed no-restricted-syntax
```
_eslint rule no-restricted-syntax in package.json <--- disabled_

see:
https://github.com/airbnb/javascript#iterators--nope
http://eslint.org/docs/rules/no-restricted-syntax
http://eslint.org/docs/2.0.0/rules/no-restricted-syntax
https://github.com/airbnb/javascript/blob/8c2df7493aebbf7a77778134d571e5b1f8fbd0dc/packages/eslint-config-airbnb-base/rules/style.js#L236

gihrig:
I'm closing this as it is fixed by #1243, which I have tested and found to be working.

If anyone has any further concerns about this please comment on that PR.

PR #1243, opened 2016-11-16
Satisfy ESLint rule: no-restricted-syntax

samit4me:
Satisfy ESLint rule (no-restricted-syntax) as per #1230


gihrig:
LGTM - passes all my tests :100:

package.json
```diff
@@ -153,8 +153,7 @@
       "redux-saga/yield-effects": 2,
       "require-yield": 0,
       "import/no-webpack-loader-syntax": 0,
-      "import/first": 0,
-      "no-restricted-syntax": 0
```
Dattaya: One question: does anyone iterate over generators with the `for ... of` loop. As far as I remember, Airbnb recommends against generators, this is why they don't care about the `for ... of` loop.

Dattaya:
LGTM

PR #1249, opened 2016-11-17
Add host command line parameter to start

ifedotov:
Proposing simple change to add a "host" command line parameter to start script.


docs/general/commands.md
```diff
@@ -66,11 +66,11 @@ npm run start:production
 Starts the production server, building the app for optimal performance: assets are
 minified and served gzipped.
 
-### Port
+### Host and Port
 
-To change the port the app is accessible at pass the `--port` option to the command
-with `--`. E.g. to make the app visible at `localhost:5000`, run the following:
-`npm start -- --port 5000`
+To change the host and/or port the app is accessible at pass the `--host` and `--port` option to the command
```
samit4me: Maybe a comma after "accessible at" and put another "and/or" between --host --port.

`To change the host and/or port the app is accessible at, pass the `--host` and/or `--port` option(s) to the command`

docs/general/commands.md
```diff
@@ -66,11 +66,11 @@ npm run start:production
...
+with `--`. E.g. to make the app visible at `my-local-hostname:5000`, run the following:
+`npm start -- --port 5000 --host my-local-hostname`
```
samit4me: Maybe put the `--host` option before `--port` so it matches the order above.

server/index.js
```diff
@@ -19,11 +19,14 @@ setup(app, {
...
-// get the intended port number, use port 3000 if not provided
+// get the intended host and port number, use localhost and port 3000 if not provided
+const host = argv.host || process.env.HOST || 'localhost';
 const port = argv.port || process.env.PORT || 3000;
 
+console.log(argv);
+
 // Start your app.
-app.listen(port, (err) => {
+app.listen(port, host, (err) => {
```
samit4me: Again a small nitpick, but for consistency maybe put the host before the port everywhere if possible.

samit4me:
@ifedotov would you mind updating the docs and merging or rebasing the latest `dev` branch?

ifedotov:
Yep, will do thanks.

server/index.js
ifedotov: This is actually internal express thing (i.e app.listen() is an express call). It has host parameter after post. That's why I didn't change logger to have host before port.

samit4me: Yep that makes sense, thanks for clarifying.

PR #1261, opened 2016-11-24
fix(eslint config): resolve numerous invalid lint errors seen in Atom…

### LINKED ISSUE #1233: ESlint errors in Atom editor
While working on dependency update #1225 I observed many linting errors in the Atom editor.

These errors do not appear when linting from the terminal, only in Atom.
The errors first appear in `98f2d0d` feat(core): Move to Jest for testing (#980)

Atom: 1.11.2
linter-eslint 8.0.0
OS X 10.11.6

To reproduce:
Open `/app/app.js`  (or other .js files) in editor

image

gihrig:
Resolves #1233 ESlint errors in Atom Editor. See that issue for discussion.

Corrected `package.json` to conform to this [#980 comment](https://github.com/mxstbr/react-boilerplate/pull/980#issuecomment-247111015)


package.json
```diff
@@ -159,7 +159,7 @@
     "settings": {
       "import/resolver": {
         "webpack": {
-          "config": "./internals/webpack/webpack.dev.babel.js"
+          "config": "./internals/webpack/webpack.prod.babel.js"
```
samit4me: My first thought was "this should be `webpack.base.babel.js`" as that is where the resolve is actually declared! But that does not work 😕 There seems to many issues open around this, do you know why it works with prod and no other webpack file @gihrig ?

I've tested before this change on Atom and VSCode on Windows and see the reported errors. Then applied your fix and can confirm it works on both 👍

samit4me: Okay, I just read your comment on the original issue, should have read that first lol. So basically this works but we don't know why? I certainly cannot figure out why either #webpackmagic

gihrig: @samit4me Yes, unfortunately #webpackmagic seems to be a an anti-pattern for many attempting to add any sizeable library to RBP, based on my recollection of issue history. I have not studied the issue and claim no expertise or special understanding, but RBP's webpack configuration feels "over engineered" to me. That's likely to be due to my lack of understanding. Still, the situation as described above, seems odd.

It would be nice to have studied input from a qualified webpack wizard.

mxstbr:
Weird. I wish I knew what was up here, but if it works, it works™

LGTM, anybody else agree?

gihrig:
Certainly agree for now - if it works, it works™ LGTM

Dattaya:
LGTM

PR #1262, opened 2016-11-24
feat(webpack) Add circular dependency plugin for Webpack.

haikyuu:
This pull request adds [circular-dependency-plugin](https://github.com/aackerman/circular-dependency-plugin). 
It is configured to show warnings when it detects a circular dependency. It is **very handy**, and can save you a **lot of time**. 
This is a sample warning:
screenshot 2016-11-24 09 33 39


internals/webpack/webpack.dev.babel.js
```diff
@@ -18,6 +19,10 @@ const plugins = [
     inject: true, // Inject all files that are generated by webpack, e.g. bundle.js
     templateContent: templateContent(), // eslint-disable-line no-use-before-define
   }),
+  new CircularDependencyPlugin({
+    exclude: /a\.js|node_modules/, // exclude node_modules
+    failOnError: false, // show a warning when there is a circular dependency
+  })
```
samit4me: Build is failing due to the linter complaining about a missing comma here

haikyuu:
No, i didn't file an issue for it since it's a simple addition. Should i?

samit4me:
LGTM, but yeah typically an issue is required first, see [CONTRIBUTING](https://github.com/mxstbr/react-boilerplate/blob/master/.github/CONTRIBUTING.md) guidelines. Reason being, something like this may need a discussion to decide if it should be added prior to you putting in the effort of actually doing the work.

This plugin does sound useful, so it's a 👍 from me, but let see what others think, should this plugin be added???

gihrig:
Having not worked with it, I would nonetheless agree. It looks like a useful tool to avoid potentially nasty problems.

mxstbr:
LGTM, sounds useful 👍

PR #1315, opened 2016-12-07
Add ability to create daemon sagas

### LINKED ISSUE #1183: Run saga policy (prevent saga duplication in more flexible way)
**Feature request**

Hello guys. This topic is not new. I've checked issues #390, #637 and #888 and have a few ideas how to improve proposed solutions.

So, I like the idea and solution of preventing saga duplication proposed by @kraffslol at #390 (by changing saga injector), but this solution was rejected. The point is:
> Sagas themselves should manage their lifecycle 

and proposal solution is to cancel watchers on `LOCATION_CHANGE` action:
```javascript
yield take(LOCATION_CHANGE);
yield cancel(watcher);
```

But in some cases this way has problems. #888 is an example when new saga should not be run automatically, because we would like to handle this kind of location change action in previously run saga. To do that we have to tweak async injector (in other case we well have saga duplicates that will ruin our application).

Long story short, what about to have a combined solution? To do that we can add some kind of saga run policy. Injector should check this policy and run (or not run) each saga. This way has benefits of both solutions and can handle all possible saga-usage cases.

Code example to illustrate my idea:
```javascript
// some container's sagas.js

export function* daemonSaga() {
  // run only once on load
}

export function* disposableSaga() {
  // run on each page enter
}
disposableSaga.runAlways = true;

export default [
  daemonSaga,
  disposableSaga,
];

// store.js
// ... much code here
store.asyncReducers = {};
store.asyncSagas = {};
return store;

// asyncInjectors.js
export function injectAsyncSagas(store, isValid) {
  return function injectSagas(sagas) {
    // invariant and warning here...
  
    sagas.forEach((saga) => {
      if (!Reflect.has(store.asyncSagas, saga) || saga.runAlways === true) {
        store.runSaga(saga);
        store.asyncSagas[saga] = true;
      }
    });
  }
}
```

What do you think?

afitiskin:
This PR fixes #1183 with 100% backward compatibility. Now we are able to create daemon sagas. Daemon sagas run only once (when loaded), next time injector ignore them. So, we don't need to track `LOCATION_CHANGE` and manually stop simple sagas to prevent duplicates: we just need to set `isDaemon` attribute of saga to `true` and that's it!

Before:
```javascript
/**
 * Watches for LOAD_REPOS actions and calls getRepos when one comes in.
 * By using `takeLatest` only the result of the latest API call is applied.
 */
export function* getReposWatcher() {
  yield fork(takeLatest, LOAD_REPOS, getRepos);
}

/**
 * Root saga manages watcher lifecycle
 */
export function* githubData() {
  // Fork watcher so we can continue execution
  const watcher = yield fork(getReposWatcher);

  // Suspend execution until location changes
  yield take(LOCATION_CHANGE);
  yield cancel(watcher);
}

// Bootstrap sagas
export default [
  githubData,
];
```

After:
```javascript
/**
 * Watches for LOAD_REPOS actions and calls getRepos when one comes in.
 * By using `takeLatest` only the result of the latest API call is applied.
 */
export function* getReposWatcher() {
  yield fork(takeLatest, LOAD_REPOS, getRepos);
}
getReposWatcher.isDaemon = true;

// Bootstrap sagas
export default [
  getReposWatcher,
];
```

### DIFF
--- a/app/store.js
+++ b/app/store.js
@@ -41,6 +41,7 @@ export default function configureStore(initialState = {}, history) {
   // Extensions
   store.runSaga = sagaMiddleware.run;
   store.asyncReducers = {}; // Async reducer registry
+  store.asyncSagas = {}; // Async saga registry
 
   // Make reducers hot reloadable, see http://mxs.is/googmo
   /* istanbul ignore next */
--- a/app/utils/asyncInjectors.js
+++ b/app/utils/asyncInjectors.js
@@ -57,7 +57,12 @@ export function injectAsyncSagas(store, isValid) {
       '(app/utils...) injectAsyncSagas: Received an empty `sagas` array'
     );
 
-    sagas.map(store.runSaga);
+    sagas.forEach((saga) => {
+      if (!(saga.isDaemon === true && Reflect.has(store.asyncSagas, saga))) {
+        store.asyncSagas[saga] = true; // eslint-disable-line no-param-reassign
+        store.runSaga(saga);
+      }
+    });
   };
 }
 


mxstbr:
@justingreenberg what do you think about this change?

justingreenberg:
@afitiskin awesome work :)

@mxstbr this change will definitely achieve stated goal, but i'm still reluctant to move any additional logic into our internal utilities... inevitbably newcomers will mistake our `isDaemon` flag as part of `redux-saga` api (since this is such an elegant solution :)) which will only lead to more confustion. i know i sound like a broken record at this point but saga duplication in this case is not a bug, it just forces people to understand the redux-saga API... IMO people should be customizing their own utilities in exactly this way, but based on their use-case not our demo app

tl;dr - this is awesome work by @afitiskin, but i would probably wait for next major release and move helpers into external package as discusssed

mxstbr:
Sure, sounds good to me.

Thank you @afitiskin for taking the time to kick off this discussion and taking the time to submit a PR!

gihrig:
This looks awesome :tada: 

Between the switch to Jest, replacing Post-CSS with Styled-components and now Daemon sagas... With a little documentation for Daemon sgas, I'd say we're very ready for 3.4.0.

Thoughts?

mxstbr:
Did you see that I closed this PR? 😉

gihrig:
I did, but I'm suggesting that it be included in the same (next?) release that also contains Jest and Styled Components. Those three make a pretty big change, I think worthy of at least a minor version bump.

mxstbr:
Yeah need to publish 3.4 already! Will get to that soon.

gihrig:
@justingreenberg @mxstbr 
> inevitbably newcomers will mistake our isDaemon flag as part of redux-saga api (since this is such an elegant solution :)) which will only lead to more confustion.

I see the issue of appearing to 'pollute' redux-saga api, but this issue is a frequent source of confusion as it is. I seriously think the confusion caused by this change is the lesser evil.

I vote to reopen and merge this PR. Regardless, documentation is the answer to either problem.

@afitiskin On the redux-saga project there are (currently) 15 [issues](https://github.com/yelouafi/redux-saga/issues?utf8=%E2%9C%93&q=is%3Aissue%20duplicate) returned on a search for "duplicate". Consider opening a PR there to make `isDaemon` an official part of the redux-saga API ?

justingreenberg:
@gihrig i'm sorry this issue is still coming up so much... it's frustrating because i **so** want it to be a learning experience for users... it's just javascript 😜 anyway, this specific issue has come up so many times, it probably sensible to address it. with respect to this PR, i can say it is the most elegant of the proposed "solutions" i've seen.

*what bugs me is that the "problem" is described as a saga duplication issue, when it isn't.*

the *real issue* is that **users want to bind daemon sagas to component lifecycle**, in the case of our demo it's a route handler loading data. our current implementation (even after merging this PR) was written for the demo—there are better solutions for route-based data loading

#### https://github.com/jfairbank/redux-saga-router

if you know your route map/data needs in advance, this tool syncs directly with history/router

```js
// sagas.js
import { router } from 'redux-saga-router'

function * rootSaga () {
  yield * router(history, {
    * '/' () {
      yield fork(takeLatest, LOAD_REPOS, getRepos)
    }
  })
}
```

#### https://github.com/threepointone/react-redux-saga

this is super simple, component/react lifecycle based

```js
// containers/HomePage/index.js
export class HomePage extends React.PureComponent {
  render() {
    //...
    return (
      <article>
        <Saga saga={getReposWatcher} /> // cancelled on unmount
```

even basic lifecycle hooks combined with `store.runSaga` may be a more sensible solution. the real motivation behind `injectAsyncSagas` had nothing to do with managing saga lifecycle, it was to **create a closure for `System.import` so saga logic could be bundled with route module**

gihrig:
@justingreenberg As always, thanks for the deeper insight ✨ 

Do you think it would be a good idea to integrate a redux-saga-router based solution into the example app as a demonstration of best practice? Pros/cons?

Issue #1320, opened 2016-12-09
#1205 breaks cleaned project start


gihrig:
A cleaned project fails on `npm start`

See this [#1205 comment](https://github.com/mxstbr/react-boilerplate/pull/1205#issuecomment-265881061)

PR #1321, opened 2016-12-09
fix(templates): use correct name of a selector

Dattaya:
Fixes regression introduced in #1205, #1320 can be closed once we merge this one.


mxstbr:
AppVeyor failures not related, shipping!

PR #1324, opened 2016-12-09
Remove webpack DedupePlugin

Dattaya:
Shouldn't have removed previous branch was easier to force push new changes, so I had to make another PR.

### DIFF
--- a/internals/webpack/webpack.prod.babel.js
+++ b/internals/webpack/webpack.prod.babel.js
@@ -24,9 +24,6 @@ module.exports = require('./webpack.base.babel')({
       async: true,
     }),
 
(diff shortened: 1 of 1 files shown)


Dattaya:
I cherry picked @samit4me commit, are we going to have a merge conflict now, @mxstbr?

gihrig:
Why bother with a merge to `master`? 

If we're about to release 3.4, this will be there soon enough, no?

mxstbr:
You're right @gihrig, good call. Thanks so much for doing this anyway @Dattaya, but I think we'll just ship 3.4 at this point. Sorry for the troubles!

Issue #1286, opened 2016-11-29
Update templates be in sync with current app functionality


jeremyadavis:
## Issue Type

- Bug

## Description

The templates need updating to remove items from 3.2 that no longer exist and add new things from 3.3.

Here are a few of things I found after setting up a new app off of 3.3.3.
**Stuff Missing to be added**
Offline Plugin Dev Message #1095 
[global-styles reference](https://github.com/mxstbr/react-boilerplate/blob/master/app/app.js#L33)

**Stuff to be removed**
CSS Modules. Should be fixed with #1260 

**Other suggestions**
I modify my templates to include following:
- import styled-components 
- include propTypes
- import createStructuredSelector and update mapStateToProps to use it instead

It might be a good idea to include those by default.

These are just the items I've personally run into, there might be a few more things lingering.

## Steps to reproduce

`npm run clean`

mxstbr:
Good catches, please submit PRs, we'd very much appreciate it!

jeremyadavis:
k, I'll try to go back over the generator templates and clean up anything that seems amiss.
The only other item I've hit so far is that the saga tests need to have
``` /*eslint-disable redux-saga/yield-effects */ ```
added.

What's the most productive way to go about creating the PR? There are some template fixes staged for 3.4 release so I don't really want to work off master knowing merge conflicts are coming down the pike.

gihrig:
Definitely work off of and submit to `dev` and have a look at open PRs to see what others might be working on.

jeremyadavis:
Implemented with PR #1331

PR #1331, opened 2016-12-10
fix(generators, templates): add example app functionality into project templates and generators

jeremyadavis:
Implements recommendations from #1286 

The updates are slightly opinionated by including propTypes and styled-components. 
I considered replacing the <div> wrapping generated components with a ```styled.div``` Wrapper component, but was unsure if others would find it useful or annoying in the case they don't use them.

### DIFF
--- a/app/containers/HomePage/tests/sagas.test.js
+++ b/app/containers/HomePage/tests/sagas.test.js
@@ -13,6 +13,7 @@ import { getRepos, getReposWatcher, githubData } from '../sagas';
 
 const username = 'mxstbr';
 
+/* eslint-disable redux-saga/yield-effects */
 describe('getRepos Saga', () => {
   let getReposGenerator;
 
--- a/internals/generators/component/es6.js.hbs
+++ b/internals/generators/component/es6.js.hbs
@@ -5,6 +5,7 @@
 */
 
 import React from 'react';
+// import styled from 'styled-components';
 
 {{#if wantMessages}}
 import { FormattedMessage } from 'react-intl';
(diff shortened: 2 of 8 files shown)


gihrig:
I generated all three flavors of components and and both of containers. I get a `no-unused-vars' lint error on all three
image

After clearing those (and expected tests), all is good :smile:

jeremyadavis:
Yeah, that was another reason why I considered making the Wrapper component to use the styled variable. 

Would you rather I just ```eslint-disable-line no-unused-vars``` on the styled-components import or go ahead and make the Wrapper component?

gihrig:
That's a good question. In the example app:
8 of 15 components have no wrapper, styles are applied in index.js.
7 of 15 components have some type of wrapper file for styles, usually called "wrapper".

Given that:
1. The wrapper tag to be `styled.` can vary, `div`, `header`, `section`, `footer`, component, etc.
2. The balance between wrapped and un-wrapped is about even in the example app.
3. The wrapped vs un-wrapped balance is (mostly) "react components" vs styled html tags.
4. The components created are expected to be part of a finished app, not just an example.

My ideal would be to add a prompt to the generators "do you want styles?" and if yes, "do you want a style wrapper?" and build accordingly.

If that's more than you want to take on now, I would suggest to simply include the import, and comment it out.

```javascript
// import styled from 'styled-components';
```

I'm interested to hear what @mxstbr thinks about this.

jeremyadavis:
I went with just commenting the import out for now since there was just work done to remove the 'do you want styles?' prompt.

For my projects I prefer to use a styled Wrapper on my components instead of a plain div so that I can easily start styling the component and leverage props and global styles if need be.

gihrig:
Thanks @jeremyadavis! That's perfect 💯 

In the interest getting 3.4 shipped, I'm going to "Update branch"...

gihrig:
@mxstbr LGTM, Comments?

mxstbr:
Awesome, thank you!

PR #1334, opened 2016-12-11
Test coverage updates

samit4me:
- Exclude test files from test coverage as including them looks bad because of all the red and it actually, reduces overall stats.
- Added coverage threshold.


gihrig:
Yes totally awesome :tada:

I deleted a few tests and build failed 😄  ✨ 

image

Issue #1335, opened 2016-12-11
mozjpeg failing on OS X / macOS and possibly others


samit4me:
`mozjpeg` is a dependency of `image-webpack-loader@3.0.0`!

The problem is that `mozjpeg` appears to have some environmental dependencies that doen't exist by default on OS X / macOS, see [issue here](https://github.com/tcoopman/image-webpack-loader/issues/49).

The "solution" is to use `image-webpack-loader@2.0.0` as it uses `jpegtran` instead of `mozjpeg`.  `image-webpack-loader` made the switch in v3 as you can see in [this commit](https://github.com/tcoopman/image-webpack-loader/commit/0cad7983fdae77f5f532a2725a2d997e24949824#diff-b9cfc7f2cdf78a7f4b91a753d10865a2).

andykenward:
I found that i had to install `libpng` to solve this issue on macOS `10.12.1`

If you have brew
`brew install libpng`

gihrig:
As covered in [#1322 comment](https://github.com/mxstbr/react-boilerplate/pull/1322#issuecomment-266267339) this error is found on some OS X installations, it does not seem to be version specific.

As an interim solution I have added to the dependency update notes that RBP will remain at `image-webpack-loader@2.0.0` until the upstream issue is resolved.

samit4me:
Thanks @andykenward , I will try that tonight. Even if this does work, I'm not sure requiring people install stuff outside of (Yarn / npm) is a great solution!

samit4me:
I can confirm that `brew install libpng` has resolved the problem, thanks @andykenward!

Now to decide if that is an acceptable solution OR do we use v2 until something upstream changes?

gihrig:
> I'm not sure requiring people install stuff outside of (Yarn / npm) is a great solution!

I totally agree. To say (document) in `gotchas.md` Mac OS: This project _may_ require you to install libpng:

```bach
 brew install libpng
```

Seems rather specialized considering that RBP is not graphics focused.

I vote for holding image-webpack-loader@2.0.0 until the upstream issue is resolved.

samit4me:
@gihrig You seem to handle all the dependencies around here, how would you like to handle this?

I assume we go back to v2 and then check on the upstream progress each time the deps are updated?

mxstbr:
> I assume we go back to v2 and then check on the upstream progress each time the deps are updated?

[...]

PR #1342, opened 2016-12-13
Rollback image-webpack-loader

samit4me:
Use v2 of `image-webpack-loader` as v3 requires environmental dependencies on mac, see #1335 

I'm not sure if we keep note of which dependencies are pinned anywhere? Should we create a doc for this?


mxstbr:
> I'm not sure if we keep note of which dependencies are pinned anywhere? Should we create a doc for this?

We _could_ add a note to `CONTRIBUTING.md`, but I'm going to ship this for now to make sure we can get `v3.4.0` out!

Issue #893, opened 2016-08-21
[suggestion] Remove pagespeed


fisch0920:
Using `slow-deps`, it appears that `psi` is the largest and most time-consuming npm dependency by far.  Given that I'm guessing most people don't even really use `npm run pagespeed`, I'd like to start a discussion here regarding removing this feature from `react-boilerplate`.

mxstbr:
I did not know that, and that sounds reasonable!

fisch0920:
I'd suggest running `slow-deps` yourself, but for my project which builds on react-boilerplage, here are the most pertinent results:

```
-------------------------------------------------------------------------------
| Dependency                                     | Time     | Size   | # Deps |
-------------------------------------------------------------------------------
| psi                                            | 2m 44.2s | 18 MB  | 323    |
| webpack                                        | 1m 42.6s | 14 MB  | 192    |
| node-sass                                      | 1m 35.6s | 13 MB  | 180    |
| stylelint                                      | 1m 27.6s | 16 MB  | 164    |
| karma                                          | 1m 11.6s | 16 MB  | 153    |
| css-loader                                     | 58.9s    | 11 MB  | 115    |
| isparta                                        | 54.2s    | 12 MB  | 102    |
| enzyme                                         | 53.7s    | 9.6 MB | 121    |
| ngrok                                          | 53.7s    | 19 MB  | 98     |
| karma-coverage                                 | 50s      | 7.2 MB | 98     |
| isparta-loader                                 | 42.8s    | 12 MB  | 103    |
| babel-preset-es2015                            | 37.3s    | 6.6 MB | 93     |
| postcss-cssnext                                | 36.8s    | 9.7 MB | 78     |
| babel-preset-stage-0                           | 34.5s    | 4.4 MB | 56     |
| html-webpack-plugin                            | 34s      | 6.0 MB | 76     |
| babelify                                       | 32.3s    | 5.9 MB | 55     |
| babel-core                                     | 23.9s    | 5.9 MB | 53     |
| html-loader                                    | 23.5s    | 4.6 MB | 66     |
| express                                        | 22.8s    | 1.0 MB | 42     |
| request                                        | 22.7s    | 3.2 MB | 66     |
```

Note that some of these may differ from what's in react-boilerplate now, but it'll at least give this discussion some context.

mxstbr:
Let's get rid of pagespeed which should make everything a huge deal faster as a first step.

[...]

PR #1352, opened 2016-12-16
Remove psi and pagespeed

mkhazov:
Addresses #893

Resolved conflicts in #894 by @fisch0920


mxstbr:
@gihrig was that a "LGTM"? 😉

gihrig:
Yes! LGTM.

If one is interested in this kind of data it's still available without plugins on the `Audits` tab in Chrome Devtools

PR #1357, opened 2016-12-16
Trigger full page reload when hmr fails

kachkaev:
Fixes https://github.com/mxstbr/react-boilerplate/issues/1353


gihrig:
Yep, LGTM, much improved UX ✨

PR #1358, opened 2016-12-16
chore(deps): update webpack

Dattaya:
Fixes #1346
- [x] Documentation is updated (if necessary)
- [x] Internal code generators and templates are updated (if necessary)
- [x] Fixed all of the errors of this kind:

```js
ERROR in multi main
Module not found: Error: Can't resolve 'babel' in 'F:\react-boilerplate'
BREAKING CHANGE: It's no longer allowed to omit the '-loader' suffix when using loaders.
                 You need to specify 'babel-loader' instead of 'babel'.
```

- [x] Replaced `System.import` with `import()`, because `System.import()` is deprecated.
- [x] Disable performance warnings in development with `performance.hints = false`
- [x] Remove `json-loader`, since: "*.json files are now supported without the json-loader."
- [x] ❗️ Fix errors occuring during testing: `SyntaxError: Unexpected token import`
- [x] Fix eslint errors:
```bash
  22:17  error  Unable to resolve path to module 'containers/App'                                  import/no-unresolved
  22:17  error  Missing file extension for "containers/App"                                        import/extensions
  25:41  error  Unable to resolve path to module 'containers/App/selectors'                        import/no-unresolved
  25:41  error  Missing file extension for "containers/App/selectors"                              import/extensions
  28:30  error  Unable to resolve path to module 'containers/LanguageProvider'                     import/no-unresolved
  28:30  error  Missing file extension for "containers/LanguageProvider"                           import/extensions
  ...
```
- [x] Final cleanup

WebStorm is very confused:
image

And it's weird, because runtime `import` loader is a [level 3 proposal](https://github.com/tc39/proposal-dynamic-import), soon to be released (by soon I mean in one year's time 😛); had to file a new issue: https://youtrack.jetbrains.com/issue/WEB-24675.

PR #1361, opened 2016-12-16
codeStyle(server): remove console.log causing server screen clutter

gihrig:
cross-env NODE_ENV=development node server

{ _: [] } <--- removed
Server started ! ✓


gihrig:
Oh, now I see that you fixed this in #1358. 😃 

Thanks for approving ✨

PR #1362, opened 2016-12-16
Added nodemon module, server will now restart on every file change

IvanJov:
Regarding the issue https://github.com/mxstbr/react-boilerplate/issues/1290 , I have added nodemon module and updated `npm start` script to use nodemon. Server will restart on evert change.

Let me know if you want some documentation to be updated too, thanks!


mxstbr:
I'm not sure about this change, what do you think @collaborators?

IvanJov:
@mxstbr Hmm, why? 🙂

mxstbr:
Another dependency 😊

IvanJov:
Hmm, you're right, but it adds something useful 😃

gihrig:
It's such a simple thing to add, if you need it. Might be better as an addition to the docs explaining the conditions where it's needed and showing steps to install.

I use and definitely need a similar setup in production, but in development it has never really bugged me to occasionally restart the server with one mouse click and three keystrokes.

Feels a little like the PSI, PageSpeed feature we just removed in #1352.

IvanJov:
@gihrig You're right, it might be better to update the docs. I am sure that people use react-boilerplate to build backend too (I am one of them), so in development process they need some automatic way to restart server on each file update. We can update the docs to explain how that can be achieved!

I can update it, if you can tell me which doc file is best for this, thanks!

mxstbr:
I've actually come around to this a bit, but I really don't want to change anything server-side while #1236 is ongoing 😕

I'll close this for now, but I'd much appreciate a note how to do this somewhere in the docs! Thanks so much for kicking off this discussion!

IvanJov:
@mxstbr You're welcome! I will probably update the docs with a tutorial how to implement `nodemon`, might be useful 😄 

@justingreenberg I agree, I can start a new doc. Let me know you have more ideas what can `api-development.md` contain. Let's see what other contributors think!

gihrig:
>  a new doc docs/general/api-development.md

Love the idea!!! ❤️

bbar:
Is anyone currently working on updating the documentation to include `docs/general/api-development.md`, or did you guys decide to go another way? If not, I don't mind updating the documentation.

dwightgunning:
I just setup nodemon based on this PR and it was quite straight forward.

When adjusting the 'start' script I also used `-w` switch to ensure that only the changes in the `server` folder will trigger a restart. Changes to the React app in `./app` are already being HMRd so we don't want nodemon involved there.

```
"start": "cross-env NODE_ENV=development ./node_modules/nodemon/bin/nodemon.js -w ./server server",
```

Issue #1363, opened 2016-12-16
Default locale translation file is not updated


p10ns11y:
## Issue Type

- [x] Bug (https://github.com/mxstbr/react-boilerplate/blob/master/.github/CONTRIBUTING.md#bug-reports)

## Description
- The changes made  to `messages.js` is not updated in `app/translations/{defaultLocale}.json`

Now, I would like to have this as separate PR after discussion and mutual agreement 

-  **extract-intl:** `const i18n = require('../../app/i18n')` will break `npm run extract-intl` if `translations` folder is deleted, as inside `i18n.js` translation files are imported (`import enTranslationMessages from './translations/en.json`)
- **appLocales** could be moved to `constants` thus by remove `i18n` dependency in `extract-intl` and for compatibility do `export { appLocales } from 'App/path/to/constans/'  in `i18n`
- **118n**: `addLocaleData` will not be loaded actually as `app.js` does not import the whole module just `{ translateMessages}` that could be done more robustly in `app.js` itself

```
appLocales.forEach((locale) => {
  System.import(`react-intl/locale-data/${locale}`).then((data) => {
    addLocaleData(data);
  });
});
```

## Steps to reproduce
1. Translation file is not updated: just update any `defaultMessages` in some container's messages
2. Delete-proof: delete `translations` folder and run `npm run extract-intl`

# Versions

- Node/NPM:6/3
- Browser:N/A

blling:
Please upgrade to the latest version:)

PR #1364, opened 2016-12-16
Fix(translation) default locale file not updated

p10ns11y:
Addressing https://github.com/mxstbr/react-boilerplate/issues/1363


mxstbr:
Note: this breaks `npm run generate language` and possibly `npm run clean`.

I'm not a fan of moving locale related stuff to the `App` container. `i18n.js` exists so people who just want a running and working i18n setup don't have to worry about it and just have an extra file. Everybody using react-boilerplate will touch the `App` container eventually, so as little things as possible should be there.

In general I love where you're going, but can we do it while leaving the i18n stuff contained in `i18n.js` please?

p10ns11y:
@mxstbr : Done!

There is possibility that `npm run generate language` just add the locale to `appLocales`
then we can do the rest of things instead of pushing the code template. That I will do as separate PR later sometime

internals/scripts/extract-intl.js
```diff
@@ -14,14 +14,14 @@ const pkg = require('../../package.json');
 const presets = pkg.babel.presets;
 const plugins = pkg.babel.plugins || [];
 
-const i18n = require('../../app/i18n');
-import { DEFAULT_LOCALE } from '../../app/containers/App/constants';
-
 require('shelljs/global');
 
+import { DEFAULT_LOCALE } from '../../app/containers/App/constants';
+import { appLocales } from '../../app/i18n';
```
mxstbr: I don't think `import`s work here?! This is ran in Node, untranspiled, which doesn't have `import`?

p10ns11y: @mxstbr : `import { DEFAULT_LOCALE } from '../../app/containers/App/constants';` was already there in master branch hence I just followed.  

`npm run extract-intl` => ` babel-node --presets latest,stage-0 -- ./internals/scripts/extract-intl.js` 

Hence `import` very well works.  

However `node internals/scripts/extract-intl.js` will not work as you said. 
https://github.com/p10ns11y/react-boilerplate/blob/d109f617f57eaae5fc7a08f7c91ad40477ab29be/internals/scripts/extract-intl.js#L100

Using arrow `async` function which is not available in node as well. Then we need to rewrite the entire script native node.

p10ns11y:
@gihrig Do I miss anything ?

gihrig:
From #1363 

> Translation file is not updated: just update any defaultMessages in some container's messages

Run `npm run extract-intl` Works as expected ✨ 

> Delete-proof: delete translations folder and run npm run extract-intl

image

p10ns11y:
@gihrig : Well, the very first commit of this pull request solved that https://github.com/mxstbr/react-boilerplate/pull/1364/commits/8902bfc605600bc05de177293810ca6740c21258

By moving  `appLocales` to `app/containers/App/constants` I was able to get rid of `i18n` dependency as `i18n` importing translations files.

Later @mxstbr suggested to keep the `i18n` and `app` as it. Then I limited the issue and pull request solving only the problem of extraction failure upon new updates. 

=> `npm generate language` should add locales in different file that has to imported in `i18n`

Would you suggest of rewriting the `npm generate language` robust way that only requires the values of `appLocales` then we can iterate throw and dynamically import.

gihrig:
@p10ns11y 

>Would you suggest of rewriting the npm generate language robust way that only requires the values of appLocales then we can iterate throw and dynamically import.

Honestly I'm not the best person to ask about that as I don't have much interest in the i18n feature as architected. But if we have it should be reliable, IMO.

I'm not sure that recovery from a deleted `app/translations` folder is necessary, but the system probably should recover from missing `xx.json` files, since the purpose of `extract-intl` is to modify `xx.json` files. The error that is presented now, as above is ugly, but meaningful. So maybe it's OK. As I said, I'm not the best one to ask...

Issue #645, opened 2016-07-05
Problem with service workers and stale code


ncrawlins:
Sorry if there is something fundamental I'm missing in the docs, but I can't figure this out. When I build for production and push to the server, the service workers (or app cache maybe?) still serve old versions of the app when I navigate to the production site.  If hit CTRL-F5 in Chrome to force refresh, I see the new version of the app, but if I then just hit F5, it will revert to the old version (served by the service workers, I guess?)

Is there something I need to be doing when pushing a new version of the app to production to let the service workers know they need to update their cache?

gihrig:
I see a similar issue. I don't know of a real solution, but I work around the problem in my local environment before `npm run start:production` by:

1. Start react-boilerplate app with `npm start`
2. Open `localhost:3000` (in Chrome)
3. Open Chrome dev tools > Resources tab
4. Delete all cached content under Local Storage, Session Storage, Application Cache, Cache Storage and Service Workers.
5. Close browser tab and stop server
6. `npm run start:production`

This needs to be done for every production version you build, and is a local workaround. I'm concerned that clients in the field may not see new content when a new production version is deployed (though I have not tested that).

All browsers listed below run the app in off-line mode as expected. In my local environment, content updates and new `npm run start:production` does not update content reliably in all browsers.
I'm on Mac Os 10.11.5
Chrome 51.0.2704 - Content _not_ updated with Cmd-R
FireFox 47.0.1 - Content _not_ updated with Cmd-R
Safari 9.1.1 - Content updated after Cmd-R (twice)

ncrawlins:
It's not a problem in development, because service workers aren't running.  It's only a problem when a new version is deployed to production.  It is being served from IIS 8.5 running .net core.  

Restarting chrome in windows does not fix it, I have to clear all cache or unregister the service workers using  chrome://serviceworker-internals/

I have no problems clearing my cache, but right now we are getting active feedback from users that can't be expected to mess with browser settings, as basic as that may seem.  They are getting a new version about once a week, sometimes more.

I was hoping there was some way to maybe put a setting in the OfflinePlugin config or somewhere. So that when i pushed a new build it would know to grab the new index.html.

But if that's just the way it's meant to work, then I guess I can just disable the offline stuff for now because they don't really need it, and the app is changing a lot.

gihrig:
@mxstbr , anyone else, is there a production-ready solution to updating content in offline-first?

mxstbr:
[...]

PR #1381, opened 2016-12-24
docs(servers): update server conf files for offline-first per #645

gihrig:
Addresses [#645 comment](https://github.com/mxstbr/react-boilerplate/issues/645#issuecomment-263550516) re caching.

While reviewing an unrelated issue I realized a need for updated server configs to fully support offline-first operation. This PR implements those changes as I believe they should be.

**I have not tested these configurations**

It would be great if others could review, test and confirm these changes are correct.

Additional resources:

Apache
http://stackoverflow.com/a/33287352/522756

Nginx
https://www.nginx.com/blog/nginx-caching-guide/ (search pragma)


Dattaya:
@ajarbol, could you test this nginx config to see whether it works as expected or not?

gihrig:
@ajarbol (any others knowledgeable in Nginx) This PR is a foundation for #1393 on the same subject, but security best practice. That needs review also.

docs/general/server-configs.md
```diff
@@ -2,17 +2,19 @@
 
 ## Apache
 
-This boilerplate includes a `.htaccess` file that does two things:
+This boilerplate includes an `app/.htaccess` file that does three things:
 
 1. Redirect all traffic to HTTPS because ServiceWorker only works for encrypted
    traffic.
 1. Rewrite all pages (e.g. `yourdomain.com/subpage`) to `yourdomain.com/index.html`
    to let `react-router` take care of presenting the correct page.
```
mxstbr: This empty line needs to be removed, it is significant in markdown.

gihrig: I think you're referring to line 11? It's removed.

gihrig:
Apparently, I am able to approve my own PR. But this seems just a bit 'over the top'.

Still I think this change is better included in 3.4 than left out, even if untested. These server configs are recommendations only, too small really to just copy-paste and go. The boilerplate does have some specific web server config requirements. Isn't it better to give folks our best effort, even if not quite right, than nothing?

So @mxstbr or @Dattaya, I'm suggesting that lacking a tested approval, we merge as is for 3.4. Agree?

PR #1322, opened 2016-12-09
v3.4.0

mxstbr:
It's time for another release!

Thanks so much for contributing everybody, please test this branch thoroughly and let me know if everything works as expected!

@gihrig @Dattaya et al is there any PRs we need to land before shipping this?

# Changelog

## Main

- **Switch to Jest for testing**, massive thanks to @dmitriiabramov 
- **Update to webpack 2 RC** (includes switching from `System.import` to `import()`), thanks to @Dattaya 
- **Add progress bar while loading code splitted routes**, thanks to @KarandikarMihir 
- **Add the "Hitchhikers Guide to `react-boilerplate` to the docs**, massive thanks to @KarandikarMihir 
- Update all dependencies, as always huge thanks to @gihrig, our master of dependencies
- Enable `import/first` eslint rule and rewrite all imports, huge thanks to @KarandikarMihir 
- Add Node v7 support and deprecate v4 support, thanks to @samit4me @MariusRumpf
- Prevent i18n language duplication, thanks to @harijoe 
- Add the webpack circular dependency plugin to avoid hard to debug errors, thanks to @haikyuu
- Refactor all selectors and generators with new naming convention, thanks to @Dattaya 
- Update generator templates, thanks to @Virsaviya @jeremyadavis 
- Add support for the `--host` parameter when running `npm start`, thanks to @ifedotov 
- Showcase reselects `createStructuredSelector` in the example, thanks to @Dattaya 
- Exclude test files from coverage report, thanks to @samit4me 
- Lint the templates for the generators, thanks to @Dattaya 
- Huge improvement to DX testing a freshly cleaned project, thanks to @outdooricon 

## Other Updates

- **Remove the webpack DedupePlugin** as it leads to errors in production and is now included by default, thanks to @samit4me 
- Remove the `npm run pagespeed` command and all related dependencies and files, thanks to @mkhazov 
- Trigger a full page reload when HMR fails, thanks @kachkaev 
- Don't import all of lodash, thanks to @jwinn 
- Improve `.editconfig`, thanks to @avdeev
- Improve the example components, thanks to @tomazy
- Fix `npm run clean` not working as expected, thanks to @adjnor 
- Fix the i18n button not updating with new state, thanks to @adjnor
- Fix console error when changing language, thanks to @samit4me 
- Fix default polyfill language, thanks to @web2style
- Fix language generation, thanks to @chaintng 
- Switch to new webpack query syntax, thanks to @shrynx
[...]

PR #1388, opened 2016-12-27
chore(deps): updated redux-saga version

Andarist:
Just a simple version bump. 0.13 -> 0.14 does not include any breaking changes, but few depractions only. I think `redux-saga` is on its way to the 1.0 version, so the sooner ppl straighten those out the better for everybody.


gihrig:
@Andarist Hi, as you may have seen when searching all PRs for "chore" I'm the guy doing routine dependency updates.

In recent weeks we have had a few breaking changes resulting from subtle changes in tracked dependencies. Since we are working on releasing a new boilerplate version (#1322) I have not been updating dependencies while that process moves forward. I would rather have a solid release with stable dependencies than the very latest with possible unknown issues.

Soon after 3.4 goes out, I expect to revisit dependency updates, I'm sure there will be _many_ packages to be updated.

Was there a specific reason you chose redux-saga to update? 
Does this update solve any specific issue? (beyond the deprecations you mentioned)

So, while I don't think we'll be merging this PR, I do appreciate the effort!

Thanks for contributing!

Andarist:
> Was there a specific reason you chose redux-saga to update? Does this update solve any specific issue?

Well, I chose this library as Im a maintainer of it and just wanted to spread the word about new version being available. `react-boilerplate` is a great starting point for many people and it would be a shame if it uses outdated version especially as our docs are promoting preferred ways of doing things instead of deprecated ones (obviously ;) ), so one searching for documentation might stumble upon inconsistencies caused by the old version.

That being said the new version add new APIs (just new versions of the deprecated ones), fixes 1 minor scheduling problem and provides some of the missing typings, so it might be desired for people using TypeScript.

> I would rather have a solid release with stable dependencies than the very latest with possible unknown issues.

As mentioned above - we pay much attention lately to not break any APIs as we are reaching the stability point, so its completely safe to upgrade.

gihrig:
@Andarist Thanks for the additional info. 

> it would be a shame if it uses outdated version especially as our docs are promoting preferred ways of doing things instead of deprecated ones

In that case it seems that updating the `redux-saga` library is only the first small step. The larger, more significant issue is updating `react-boilerplate` to take advantage of the new API.

Is there a guide to deprecated API and updated replacements?

Andarist:
[0.14 Release notes](https://github.com/redux-saga/redux-saga/releases/tag/v0.14.0).

I've already updated the code on my local branch but needs to update the tests too, hopefully Ill have time to do it 2morrow.

Andarist:
I've pushed the updated sample code + tests and even managed to increase a coverage (+0.4% ftw o/)

mxstbr:
This LGTM, thanks!! @gihrig I feel like we should land this?

gihrig:
@mxstbr AppVeyor is still being difficult 😦 , but go ahead and merge this 🎉 

@Andarist Thank you for contributing, much appreciated! ✨

PR #1399, opened 2016-12-30
Use node servers default host when none specified

### LINKED ISSUE #1398: (release 3.4) ngrok LAN connection failing to connect
I'm seeing a regression in how ngrok is serving up the content on the LAN url for 3.4 .

When I perform `yarn start` on master, the output looks normal and when I navigate to the LAN url, the boilerplate page loads fine. When I then immediately switch to 3.4 branch, run `yarn` and then `yarn start`, the terminal output looks the same (normal) but when I navigate to the new LAN url, I cannot reach the site. Navigating to localhost:3000 works as expected.

I have tried bumping the ngrok version down to 2.2.3, but that doesn't seem to help.

I'm not really sure where to begin for debugging...

outdooricon:
Node's http.Server#listen does some smart choosing for default host '::' or '0.0.0.0' depending on IPv6 support. Use that instead of binding to localhost specifically. This mimics previous server functionality and allows ngrok access to work again.

This fixes #1398

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -20,7 +20,10 @@ setup(app, {
 });
 
 // get the intended host and port number, use localhost and port 3000 if not provided
-const host = argv.host || process.env.HOST || 'localhost';
+const customHost = argv.host || process.env.HOST;
+const host = customHost || null; // Let http.Server use its default IPv6/4 host
+const prettyHost = customHost || 'localhost';
+
 const port = argv.port || process.env.PORT || 3000;
 
 // Start your app.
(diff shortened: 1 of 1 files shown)


gihrig:
`npm run start:tunnel` does work as expected, but --host and --port don't

image

outdooricon:
I think you need to use `--`, so `npm start -- --host test --port 2500`. I have a lack of access to my computer right now to be able to tell you definitively, but that at least is what the new docs say to do.

gihrig:
@outdooricon D'oh! You're right, my bad 😦 

LGTM

mxstbr:
Awesome, thanks!

Dattaya:
@mxstbr, was this also merged into dev?

outdooricon:
@Dattaya fortunately only 3.4 had the issue.

Issue #865, opened 2016-08-14
Tests fail in minimal example app after npm run clean


gihrig:
Repro steps:
Note: wait for #862 to land. Other eslint errors will abort before these can run.
`npm run setup`
`npm run clean`
`npm run test`

```
ERROR in ./app/tests/store.test.js
Module not found: Error: Can't resolve './store' in '/Users/glen/Documents/JavaScript/Research/react-bp2/app/tests'
 @ ./app/tests/store.test.js 18:209-227
```

The error is a result of the templates dir structure not matching that of the boilerplate App. The immediate issue is addressed in #863, specifically [#863 (note)](https://github.com/mxstbr/react-boilerplate/pull/863/files#diff-46c47e6183968ac4496ee7740147bc2bL3)

*.test.js files are grouped under `/templates` directory. `npm run clean` copies these files to `/tests` directory in the minimal example project created by the clean command.

This creates an issue where the path to the tested component varies between `templates` and `tests`. 

This results in a bug where tests fail after `clean` because the tested component does not resolve.

uzarubin:
It looks like this has been resolved already. No?

gihrig:
@uzarubin Yes closed by #863.

Thanks for brining this up 👍

gihrig:
@uzarubin I was mistaken. This issue is _not_ fixed, only worked around.

The problem is as stated in the OP. The directory structure in templates does not match the generated structure.

My bad 😢

uzarubin:
@gihrig no worries. I was just looking over issues to see which ones were resolved. I'll try to look into this one sometime tomorrow.

mxstbr:
Is this still an issue?

gihrig:
Yes this is still an issue. 

 `internals/templates/*` files are moved to various locations by the clean script. `store.test.js` in particular has eslint disabled [here](https://github.com/mxstbr/react-boilerplate/blob/master/internals/templates/store.test.js#L6) becuase the templates dir structure does not match the generated dir structure. This is the root cause for this issue, the eslint disable is a work around.

It has been agreed [here](https://github.com/mxstbr/react-boilerplate/pull/863#issuecomment-239659433) that `internals/templates/` structure should match the destination structure. This requires reworking [`internals/scripts/clean.js`](https://github.com/mxstbr/react-boilerplate/blob/master/internals/scripts/clean.js#L24-L29) as well.

gihrig:
Thoroughly fixed by @outdooricon in #1414 🎉

PR #1414, opened 2017-01-06
Reorg some templates and add tests

outdooricon:
This adds enough testing to get the post `npm run clean` code test coverage from this:
screen shot 2017-01-06 at 1 04 57 pm

to this:
screen shot 2017-01-06 at 10 43 22 am

This is important because it sets people on the right track from the start to maintaining high code-coverage. Also, without this the `npm build` won't be able to run after a clean (that's what triggered my work on it).


gihrig:
I'm getting a test failure on your `moreCleanTests` branch before cleaning. This is due to #865 which was worked around by disabling ESLint. Now it's become a problem for testing.

image

It would fantastic if you wanted to fix #865 :smile:

outdooricon:
Hmm, I'm confused... Why can't I reproduce this error?
screen shot 2017-01-06 at 2 48 56 pm

gihrig:
After cleaning, all looks good ✨

outdooricon:
ok, addressing #865 right now...

gihrig:
@outdooricon 

> Hmm, I'm confused... Why can't I reproduce this error?

Good question! What I did.

`git clone https://github.com/outdooricon/react-boilerplate.git rbp-tests && rbp-tests`
checkout `moreCleanTests` as local branch (using GitUp)
`yarn`
`npm test`

Got that failure.

outdooricon:
well that's just bizarre :) Regardless, this should fix it now

gihrig:
OK, all works now 😄 

But why place `store.test.js` in `templates/appTests` as opposed to `templates/appContainer/tests` ?

internals/templates/store.test.js → internals/templates/_appTests_/store.test.js

outdooricon:
Because appContainer is containers/app, whereas appTests correlate to app/.

On Fri, Jan 6, 2017, 15:58 Glen Ihrig <notifications@github.com> wrote:

OK, all works now 😄

But why place store.test.js in templates/appTests as opposed to
templates/appContainer/tests ?

internals/templates/store.test.js → internals/templates/*appTests*
/store.test.js

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/mxstbr/react-boilerplate/pull/1414#issuecomment-271019734>,
or mute the thread

gihrig:
> Because appContainer is containers/app, whereas appTests correlate to app/.

Correction:

> Because appContainer is containers/App, whereas appTests correlate to app/.

A naming mess...

I realize this is getting farther away from the original PR, but how about changing `templates/appContainer` to simply `templates/App'? 

That more closely matches the structure of the working application. If there is to be confusion between `app` and `containers/App` then it seems to me that `app` should be renamed to something like `client` to better match `server`. But that is way too far beyond the scope of this PR.

outdooricon:
At this point I'd really like to just make a containers folder and then have App and the others inside of it. I feel like that results in the most obvious correlation. What are your thoughts on that?

gihrig:
> At this point I'd really like to just make a containers folder and then have App and the others inside of it. I feel like that results in the most obvious correlation. What are your thoughts on that?

Yes, I could go for that!

outdooricon:
Ok @gihrig , done!

gihrig:
Oh, almost forgot, bonus points for fixing #863! 🎉

outdooricon:
:+1: glad I could help!

outdooricon:
Hmm, this CI error came up after i updated my branch with latest from dev. The source seem to be the change to store.js now running straight "import" for the hot reload in #1358 for node 5 & 6. I'm not really sure what to do at this point, @Dattaya do you have any insight into this?

mxstbr:
Oh wow this is an awesome PR, can we land it for the 3.4 release? I _really_ want to get that out, but this is amazing!

outdooricon:
Thanks! I'd really like that. I just need some guidance on this new CI failure, or maybe someone can kick off CI again with the cache cleared and see if that helps?

Dattaya:
oh, that error reappeared again 😢, it can be solved by either disabling caching of `node_modules` in `.travis.yml` or merging #1415.

outdooricon:
Removing the caching seems like a good move, for consistent builds sake, IMO. I'll make the change.

outdooricon:
Ok, that's better. @gihrig @Dattaya @mxstbr Are we ok with the no cache solution?

mxstbr:
> Are we ok with the no cache solution?

I'm pretty sure this will make build times a lot longer since we have to wait for `npm install` to finish every single time. Can we somehow invalidate the cache and try again with a fresh install if the build fails?

outdooricon:
@mxstbr I believe so: https://docs.travis-ci.com/user/caching/#Clearing-Caches However, I did not have to permissions to do this. I will back out my travis change if you want to go that route.

Dattaya:
I can't see the cache for this PR, @outdooricon, please, revert back the last commit and I'll clear the cache if the error still persist.

outdooricon:
Ok, I pulled the change. Can someone force travis to start a new build please?

outdooricon:
Ok, I just made another minor commit to force the build and it does look like it's still dealing with the cache issue.

Dattaya:
Done—removed the cache, restarted the build and the problem is gone.

Dattaya:
I have one question, not directly related to this PR, because it can be done in another: should we test `global-styles.js` or just remove it from the coverage report: `"!app/global-styles.js",`?

internals/scripts/clean.js
```diff
@@ -13,68 +13,36 @@ if (!test('-e', 'internals/templates')) {
...
-cp('internals/templates/notFoundPage/messages.js', 'app/containers/NotFoundPage/messages.js');
-cp('internals/templates/homePage/homePage.js', 'app/containers/HomePage/index.js');
-cp('internals/templates/homePage/messages.js', 'app/containers/HomePage/messages.js');
-
-// Handle Translations
-rm('-rf', 'app/translations/*')
-mkdir('-p', 'app/translations');
-cp('internals/templates/translations/en.json',
-  'app/translations/en.json');
+mv('internals/templates/containers', 'app');
```
Dattaya: Doesn't work on Windows, but it will work if you remove `containers` folder completely: `rm('-rf', 'app/containers');`. Same true for `translations`.

gihrig:
Re caching:

1. We definitely should have it.
2. Can we merge #1415 now? - That should fix the caching issue, and if not we should know about it.

internals/scripts/clean.js
gihrig: > Doesn't work on Windows

Are you referring to  a CI build, or some other problem?

gihrig:
@ Dattaya Re: test global-styles.js, I would say that since it's JS it should be tested, but if it's not practical to get 98%+ coverage for some reason, skip it.

outdooricon:
> @ Dattaya Re: test global-styles.js, I would say that since it's JS it should be tested, but if it's not practical to get 98%+ coverage for some reason, skip it.

FWIW, I did take a decent stab at creating a test for it but couldn't for the life of me figure out how to test injectGlobal.

internals/scripts/clean.js
outdooricon: ah yes, thanks for noticing that

gihrig:
Tested `npm run clean` on OS X 10.11.6 all good 💯

Dattaya:
Thank you so much, @outdooricon 💯

PR #1319, opened 2016-12-08
hotfix(deps): lock react-addons-test-utils

Dattaya:
In case you're interested in a hotfix to the master branch, @mxstbr

PR #1415, opened 2017-01-07
RFC: chore(ci): switch to yarn

Dattaya:
## React Boilerplate

Thank you for contributing! Please take a moment to review our [**contributing guidelines**](https://github.com/mxstbr/react-boilerplate/blob/master/.github/CONTRIBUTING.md)

- [ ] ? Documentation is updated (if necessary)

As it was proposed by @gihrig in https://github.com/mxstbr/react-boilerplate/pull/1358#issuecomment-270221247
A theoretical upside of this PR: no problems with dependencies, it's sort of a downside at the same time, because we won't be able to detect issues with npm like recently that we had with `react-addons-test-utils` (https://github.com/mxstbr/react-boilerplate/pull/1319)


gihrig:
> it's sort of a downside at the same time, because we won't be able to detect issues with npm like recently that we had with react-addons-test-utils (#1319)

As the NPM world (?) move to Yarn this will become less of an issue. Perhaps we should add a warning message to setup.js when Yarn is not found to let the user know that `npm install` may produce dependency errors.

That could be considered beyond the scope of this PR, and since Travis and AppVeyor are happy wit it, i'm happy with it too 🎉 

I would still like to hear opinions and experience from others...

gihrig:
@collaborators The related issue is blocking new PRs. It would be nice to have additional thoughts on this PR before merging...

gihrig:
As this is an internal change, the only to really test it is to put it into use - merging...

Issue #1406, opened 2017-01-04
package.json run lint:css missing


jimmyheaddon:
Hello, the npm script 'lint:css' was removed on Nov 2nd 2016 in [this pull request](https://github.com/mxstbr/react-boilerplate/commit/33b473912addd03f4436d8915e1df3c5a80b90c5#diff-b9cfc7f2cdf78a7f4b91a753d10865a2).

However, the [commands documentation](https://github.com/mxstbr/react-boilerplate/blob/master/docs/general/commands.md) still refers to this command.  Should it still exist, or is the documentation out-of-date?  I can't see any way of linting the CSS anymore, as the 'lint' command simply redirects to 'lint:js' now?

gihrig:
@jimmyheaddon You are correct. Please consider a openning PR to update the docs.

There is experimental support for linting styled-components [here](https://github.com/styled-components/stylelint-processor-styled-components). It would be great to hear about your experience with that project.

jimmyheaddon:
@gihrig Sure, I can raise a PR to remove all references to lint:css from the documentation.  I'm guessing we'd want a separate PR in future for the styled-components support?

gihrig:
Yeah, since styled-components support for linting is experimental, I'd recommend a separate PR.

PR #1416, opened 2017-01-07
#1406 remove css lint documentation references

jimmyheaddon:
Removed referenced to lint:css, which is currently not supported.

### DIFF
--- a/docs/general/commands.md
+++ b/docs/general/commands.md
@@ -171,7 +171,7 @@ about the bundle size.
 npm run lint
 ```
 
-Lints your JavaScript and CSS.
+Lints your JavaScript.
 
 ### JavaScript
 
@@ -180,11 +180,3 @@ npm run lint:js
 ```
 
 Only lints your JavaScript.
-
-### CSS
-
-```Shell
-npm run lint:css
-```
-
-Only lints your CSS.
--- a/docs/general/commands.md
+++ b/docs/general/commands.md
@@ -171,7 +171,7 @@ about the bundle size.
 npm run lint
 ```
 
-Lints your JavaScript.
+Lints your JavaScript<!-- and CSS-->.
 
 ### JavaScript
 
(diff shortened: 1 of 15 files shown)


jimmyheaddon:
As requested in issue #1406

docs/general/commands.md
```diff
@@ -171,7 +171,7 @@ about the bundle size.
 npm run lint
 ```
 
-Lints your JavaScript and CSS.
+Lints your JavaScript.
 
 ### JavaScript
```
gihrig: Looks kind of weird/redundant. How about leaving both CSS and JS sections in the doc for future use, but commenting them out `<!-- ... --> ` until style linting get worked out?

image

Dattaya: ping @jimmyheaddon

jimmyheaddon:
@gihrig how's that look?  I'd avoided comments because they're notoriously tricky in markdown (most MD viewers struggle with the `<!-- -->` syntax), but GitHub handles it, so if that's OK then this should suffice.

If we're not happy that the comments show in other markdown viewers, then I think we're safer deleting the text as per the original commit?

gihrig:
@jimmyheaddon 
>If we're not happy that the comments show in other markdown viewers, then I think we're safer deleting the text as per the original commit?

Agreed, I didn't that, thanks for bringing it up. So just delete the CSS and second JS reference, since `npm run lint` does the same as `npm run lint:js`.

I hadn't noticed before, but this PR needs to be based on `dev` (will be merged into 3.4 before release. You can change this on the PR screen, no need for a new PR).

jimmyheaddon:
Thanks @gihrig, I've changed the base to dev, and have removed the commented out markdown, along with the redundant reference to lint:js.  Hopefully we can get styled-components linting supporting in soon!

gihrig:
@jimmyheaddon Yikes, that branch change brought a whole bunch of unrelated stuff.

Can you open a new PR against current `dev` branch and make these changes please. Sorry for the trouble. 😢

jimmyheaddon:
@gihrig Yeah I didn't think that had gone quite to plan, but I don't have the privileges to correct it inline, so I'll do as you've suggested!

PR #1420, opened 2017-01-09
Alternative saga lifecycle

adelsz:
This is a proof of concept demonstrating an alternative saga lifecycle management.

Lets suppose we have the following routing scheme:
```
/tree/branches/ -> BranchPage container
/tree/branches/:branchId 
/tree/about -> AboutPage container
```
And we require:
1. The `BranchPage` sagas should be injected and started when the user enters `/tree/branches/*` from `/tree/about` or any other path.
2. The `BranchPage` sagas should be canceled when the user leaves from `/tree/branches/*` to any other path like `/tree/about`.
3. The `BranchPage` container sagas keep running (without interruption/reinjection) while the user is on a `/tree/branches/*` path even when he jumps between different branch ids `/tree/branches/branch1`, `/tree/branches/branch2` etc.

This can be achieved by injecting the route `BranchPage` sagas on `onEnter` and canceling them on `onLeave` route events of the `/tree/branches/*` route.
For this we modify the `injectAsyncSagas` method to return our sagas Tasks. After that we can fully manage our sagas from `app/routes.js`.
Note that this also allows us to have global per app sagas for the root path `/*`.

To keep it simple this PR demonstrates this approach on the boilerplate default Example app.

The tests\generators\templates have been updated accordingly.

### DIFF
--- a/app/containers/HomePage/sagas.js
+++ b/app/containers/HomePage/sagas.js
@@ -2,8 +2,8 @@
  * Gets the repositories of the user from Github
  */
 
-import { take, call, put, select, cancel, takeLatest } from 'redux-saga/effects';
-import { LOCATION_CHANGE } from 'react-router-redux';
+import { fork, call, put, select } from 'redux-saga/effects';
+import { takeLatest } from 'redux-saga';
 import { LOAD_REPOS } from 'containers/App/constants';
 import { reposLoaded, repoLoadingError } from 'containers/App/actions';
 
@@ -27,18 +27,15 @@ export function* getRepos() {
   }
 }
 
+export function* repListen() {
+  yield* takeLatest(LOAD_REPOS, getRepos);
+}
 /**
  * Root saga manages watcher lifecycle
  */
 export function* githubData() {
   // Watches for LOAD_REPOS actions and calls getRepos when one comes in.
-  // By using `takeLatest` only the result of the latest API call is applied.
-  // It returns task descriptor (just like fork) so we can continue execution
-  const watcher = yield takeLatest(LOAD_REPOS, getRepos);
-
-  // Suspend execution until location changes
-  yield take(LOCATION_CHANGE);
-  yield cancel(watcher);
+  yield fork(repListen);
 }
 
 // Bootstrap sagas
--- a/app/containers/HomePage/sagas.js
+++ b/app/containers/HomePage/sagas.js
@@ -2,8 +2,7 @@
  * Gets the repositories of the user from Github
  */
 
-import { fork, call, put, select } from 'redux-saga/effects';
-import { takeLatest } from 'redux-saga';
+import { call, put, select, takeLatest } from 'redux-saga/effects';
 import { LOAD_REPOS } from 'containers/App/constants';
 import { reposLoaded, repoLoadingError } from 'containers/App/actions';
 
@@ -27,15 +26,12 @@ export function* getRepos() {
   }
 }
 
-export function* repListen() {
-  yield* takeLatest(LOAD_REPOS, getRepos);
-}
 /**
  * Root saga manages watcher lifecycle
  */
 export function* githubData() {
   // Watches for LOAD_REPOS actions and calls getRepos when one comes in.
-  yield fork(repListen);
+  yield takeLatest(LOAD_REPOS, getRepos);
 }
 
 // Bootstrap sagas
--- a/app/containers/HomePage/tests/sagas.test.js
+++ b/app/containers/HomePage/tests/sagas.test.js
@@ -2,10 +2,7 @@
  * Tests for HomePage sagas
  */
 
-import { cancel, take, put, takeLatest } from 'redux-saga/effects';
-import { createMockTask } from 'redux-saga/lib/utils';
-
-import { LOCATION_CHANGE } from 'react-router-redux';
+import { put, takeLatest } from 'redux-saga/effects';
 
 import { LOAD_REPOS } from 'containers/App/constants';
 import { reposLoaded, repoLoadingError } from 'containers/App/actions';
@@ -49,20 +46,9 @@ describe('getRepos Saga', () => {
 
 describe('githubDataSaga Saga', () => {
   const githubDataSaga = githubData();
-  const mockedTask = createMockTask();
 
   it('should start task to watch for LOAD_REPOS action', () => {
     const takeLatestDescriptor = githubDataSaga.next().value;
     expect(takeLatestDescriptor).toEqual(takeLatest(LOAD_REPOS, getRepos));
-  });
-
-  it('should yield until LOCATION_CHANGE action', () => {
-    const takeDescriptor = githubDataSaga.next(mockedTask).value;
-    expect(takeDescriptor).toEqual(take(LOCATION_CHANGE));
-  });
-
-  it('should cancel the forked task when LOCATION_CHANGE happens', () => {
-    const cancelDescriptor = githubDataSaga.next().value;
-    expect(cancelDescriptor).toEqual(cancel(mockedTask));
   });
 });
--- a/app/routes.js
+++ b/app/routes.js
@@ -19,19 +19,45 @@ export default function createRoutes(store) {
   return [
     {
       path: '/',
+      onEnter(nextState, replace, callback) {
+        // onEnter gets called when we visit a route
+        // childRoute changes do not trigger onEnter, which is a desired behavior
+
+
+        // Prevent saga reinjection if they are running
+        if (this.loadedSagas) {
+          callback();
+          return;
+        }
+
+        // Inject sagas as usual
+        const importModules = System.import('containers/HomePage/sagas');
+
+        importModules.then((sagas) => {
+          this.loadedSagas = injectSagas(sagas.default);
+          callback();
+        });
+
+        importModules.catch(errorLoading);
+      },
+      onLeave() {
+        // onLeave gets called when we leave the route
+        // Cancel the sagas if they are running
+        if (this.loadedSagas) {
+          this.loadedSagas.forEach((saga) => saga.cancel());
+          delete this.loadedSagas;
+        }
+      },
       getComponent(nextState, cb) {
         const importModules = Promise.all([
-          import('containers/HomePage/reducer'),
-          import('containers/HomePage/sagas'),
-          import('containers/HomePage'),
+          System.import('containers/HomePage'),
+          System.import('containers/HomePage/reducer'),
         ]);
 
         const renderRoute = loadModule(cb);
 
-        importModules.then(([reducer, sagas, component]) => {
+        importModules.then(([component, reducer]) => {
           injectReducer('home', reducer.default);
-          injectSagas(sagas.default);
-
           renderRoute(component);
         });
 
--- a/app/utils/asyncInjectors.js
+++ b/app/utils/asyncInjectors.js
@@ -62,7 +62,8 @@ export function injectAsyncSagas(store, isValid) {
       '(app/utils...) injectAsyncSagas: Received an empty `sagas` array'
     );
 
-    sagas.map(store.runSaga);
+    // return redux-saga Tasks, so they can be canceled on router events
+    return sagas.map(store.runSaga);
   };
 }
 
--- a/internals/generators/route/routeWithReducer.hbs
+++ b/internals/generators/route/routeWithReducer.hbs
@@ -2,23 +2,40 @@
       path: '{{ path }}',
       getComponent(nextState, cb) {
         const importModules = Promise.all([
-          import('containers/{{ properCase component }}/reducer'),
-          {{#if useSagas}}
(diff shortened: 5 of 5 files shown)


petemill:
@adelsz this is great. You're calling `.cancel()` on each saga. How is it you've come to have that function on your returned sagas?

adelsz:
@petemill Added a return statement, modifying `injectAsyncSagas`  [here](https://github.com/adelsz/react-boilerplate/blob/e9ce045f7cc64ea0ebdbc5f99de0de99c50bb664/app/utils/asyncInjectors.js#L66).

yantakus:
It doesn't work for me, it never gets inside
```javascript
if (this.loadedSagas) {
    callback();
    return;
}
```
So the saga gets injected twice. It's quite weird, because this code is executed:
```javascript
importModules.then((sagas) => {
   this.loadedSagas = injectSagas(sagas.default);
   callback();
});
```

Also, why do you think the problem is in sagas?

I tried to add `console.log` in different places and figured out every route gets fired twice. Isn't it the root problem?

gihrig:
Well, @yantakus experience to the contrary, I loaded it up and put it through its paces, no duplicate actions, everything seems to work as expected. 😄 

In other news... @andreipopovici [likes](https://github.com/mxstbr/react-boilerplate/issues/1434#issuecomment-271843095) it 🎉

andreipopovici:
It just makes sense to manage saga lifecycle within the route lifecycle, if sagas are paired with routes.

OTOH, `react-boilerplate` has no route generator, but does have a container generator which creates a `sagas.js` within a container's folder. So it *could* be argued that saga lifecycle would belong inside a React Component's mount/unmount lifecycle, too.

However, something like (react-redux-saga)[https://www.npmjs.com/package/react-redux-saga], which was suggested elsewhere, is pretty hairy to integrate into `react-boilerplate`, since it needs to wrap the root `<App/>` Component into a `<Sagas>` HOC.

Overall, this method worked better for me, so +1!

petemill:
@adelsz thanks, it works beautifully and perfectly. I'm working on a couple options for less code repitition in routes.js. I also allowed for a saga to define whether it is *not* cancelled, and whether it is re-run each time that route is invoked or just left as a single instance running. Just working out whether that's useful or not.

@yantakus is the problem definitely that `onEnter` for your routes are being called twice, before `onLeave`? That sounds mysterious. Perhaps also check that `this` is correct and not undefined in your `console.log`. Make sure you're not doing `onEnter: function() {....` but you're doing `onEnter() {...`

yantakus:
@petemill Yes, you are right, I'm doing `onEnter: function() {....` and this is the key here.

I have lots of routes, so I'm using a `createRoute` helper function for DRYing them:

```javascript
  const createRoute = (path, name, component, sagas) => {
    const route = {
      path,
      name,
      getComponent (nextState, cb) {
        // ...
      }
    }
    if (sagas) {
      route.onEnter = function (nextState, replace, callback) {
        // ...
      }
      route.onLeave = function () {
        // ...
      }
    }
    return route
  }
```

I can't just include `onEnter` and `onLeave` by default, as I have routes without sagas and for such routes my `getComponent` isn't executed if there's `onEnter` and `onLeave` even everything inside these methods is wrapped into `if (sagas) {}`. This is really weird.

I tried to use `route.onEnter = () => {}` but in this case `this` inside it is undefined.

Any ideas?

gihrig:
>I tried to use route.onEnter = () => {} but in this case this inside it is undefined.

@yantakus I presume you are aware of ES6 arrow functions [non-binding of this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#No_binding_of_this)?

petemill:
@yantakus in your version, your `onEnter` and `onLeave` are two separate functions, with two separate versions of `this`. You'll need to make, perhaps, a class, say `class SagaRoute { onEnter():..., onLeave():... getComponent():...` where each function's `this` will be the same object. Another way could be have the functions as separate reusable, but use `reuseableOnEnter.bind(this...`

I'm actually coming up with my own version sometime today of what you're trying to do, so I'll let you know. And sorry, all, if this thread is somewhat irrelevant to the main issue.

yantakus:
@gihrig, @petemill Thanks a lot for the clarifications!

petemill:
I wrapped all this functionality in a class, to make for a clean routes.js. Seems to work pretty well, and still allow for a custom onEnter and onLeave.

Usage:
```js
[
  new LazyLoadReduxRoute({
    getComponentModule: () => System.import('containers/HomePage'),
    path: '/',
    name: 'home',
  }),
  new LazyLoadReduxRoute({
    reducerName: 'loginPage',
    getComponentModule: () => System.import('containers/LoginPage'),
    getReducerModule: () => System.import('containers/LoginPage/reducer'),
    getSagaModule: () => System.import('containers/LoginPage/sagas'),
    path: 'login',
    name: 'Login',
  }),
]
```

Class (needs to be exported from a factory function because of reliance on store). Sorry for the flow typings:
```js
// @flow
declare var System:any;
import { getAsyncInjectors } from './asyncInjectors';


export default function LazyLoadReduxRouteFactory(store: any) {

  const { injectReducer, injectSagas, removeSagas } = getAsyncInjectors(store);



  return class LazyLoadReduxRoute {

    reducerName: string;
    getComponentModule: Function;
    getReducerModule: Function;
    getSagaModule: Function;
    loadedSagaTasks: [];
    getComponentCustom: ?Function;
    onEnterCustom: ?Function;
    onLeaveCustom: ?Function;

    constructor({ reducerName, getComponentModule, getReducerModule, getSagaModule, getComponent, onEnter, onLeave, ...routeProps }: Object) {

      //validate
      if (typeof getComponentModule !== 'function') {
        throw new Error('Must pass param getComponentModule as function to create a LazyLoadReduxRoute');
      }
      //save details to instance
      this.getComponentModule = getComponentModule;
      //process optional args
      if (typeof getReducerModule === 'function') {
        if (typeof reducerName !== 'string') {
          throw new Error('When providing a reducer, must also provide a reducerName');
        }
        this.reducerName = reducerName;
        this.getReducerModule = getReducerModule;
      }
      if (typeof getSagaModule === 'function') {
        this.getSagaModule = getSagaModule;
      }
      //allow events we handled to have custom logic
      if (typeof getComponent === 'function') {
        this.getComponentCustom = getComponent;
      }
      if (typeof onEnter === 'function') {
        this.onEnterCustom = onEnter;
      }
      if (typeof onLeave === 'function') {
        this.onLeaveCustom = onLeave;
      }
      //expose any other route properties caller provides
      Object.assign(this, routeProps);
    }


    async getComponent(nextState: any, cb: Function) {

      //get modules in paralell
      const ops = [ this.getComponentModule() ];
      //reducer is optional
      if (this.getReducerModule) {
        ops.push(
          this.getReducerModule()
          .then(reducerModule => injectReducer(this.reducerName, reducerModule.default))
        );
      }
      const [ componentModule ] = await Promise.all(ops);
      cb(null, componentModule.default);
    }


    async onEnter(nextState: any, replace: Function) {

      //should we be loading sagas
      if (this.getSagaModule) {
        //sanity check, should not have loaded tasks yet (or still from previous visit to page)
        if (!this.loadedSagaTasks) {
          //load saga module
          const sagas = await this.getSagaModule();
          //remember sagas and saga-tasks, for cancelling later
          this.loadedSagaTasks = injectSagas(sagas.default);
        }
        //handle sagas not previously unloaded
        else {
          console.error('sagas were already loaded', { tasks: this.loadedSagaTasks });
        }
      }
      //custom route onEnter
      if (this.onEnterCustom) {
        this.onEnterCustom(nextState, replace);
      }
    }


    onLeave(prevState: any) {

      //custom route onLeave
      if (this.onLeaveCustom) {
        this.onLeaveCustom(prevState);
      }
      //cancel running tasks, if any
      if (this.loadedSagaTasks) {
        this.loadedSagaTasks.forEach(saga => saga.cancel());
        //remove reference to cancelled tasks
        delete this.loadedSagaTasks;
      }
    }


  }



}

```

adelsz:
Hey guys, glad to see that many of you find this approach useful.
@petemill, interesting approach with the route factory, but I find it a bit too complicated and implicit.
The route generator is already a route factory in some sense.
I prefer to keep it simple and straightforward by adding the imports explicitly to `routes.js`, and to ask the user directly if he needs sagas for this route on route generation.

I guess I will now proceed to bring this PR to a better state by adding the needed generators and tests.

petemill:
@adelsz agreed, it's not good for inclusion in this boilerplate project, was an exercise to see how much code repetition could be removed, if the developer so desired (as I did), as somewhat unrelated to your PR, but bought about by comment from @yantakus. Good luck with the rest, I'm already using this in an app now.

gihrig:
@williamgranli [likes it](https://github.com/mxstbr/react-boilerplate/issues/1434#issuecomment-272395047) 😄

michelalbers:
Thank you! Works like a charm for me and it's so simple yet elegant! 👍

adelsz:
@michelalbers 
Thanks for mentioning it, I should have probably added a small section on this in the PR body.
Yes, child routes must be specified if you don't want their changes to trigger parent's route `onEnter` and `onLeave` events.

michelalbers:
After playing around with this and thinking about it some more, I came to the conclusion that the saga lifecycle in the example app is very very confusing and this should be the way its done. Otherwise sagas will be injected multiple times and a developer has to first start noticing that and then second find a solution. I use the method described above now and it's working flawlessly. I'm for merging this! :)

nndung179:
In `onEnter` event. The statement below is actually not working

```javascript
if (this.loadedSagas) {
    cb();
    return;
}
```

I suggest to move it into the block 
```javascript
importModules.then((sagas) => {
   if (this.loadedSagas) {
        cb();
        return;
     }
     this.loadedSagas = injectSagas(sagas.default);
    // @TODO - loading data - dispatch action to call api here
     cb();
});
```

Here is the full of code:
```javascript
onEnter(nextState, replace, cb) {
  const importModules = System.import('containers/VideoDetailsPage/sagas');
  importModules.then((sagas) => {
    if (this.loadedSagas) {
      cb();
      return;
    }
    this.loadedSagas = injectSagas(sagas.default);
    // @TODO - loading data - dispatch action to call api here
    cb();
  });
},
```

philipchmalts:
In the onLeave cancel event has anyone experienced the `uncaught at update Generator is already running` issue when cancelling sagas?

FoxxMD:
@petemill what does the import statement to use `LazyLoadReduxRoute` look like in your routes.js?

Pinesy:
@FoxxMD This is how i've got it working...

```js
import LazyLoadReduxRouteFactory from './utils/LazyLoadReduxRouteFactory';

  // create reusable async injectors using getAsyncInjectors factory
  const LazyLoadReduxRoute = LazyLoadReduxRouteFactory(store);

  const routes = [
    new LazyLoadReduxRoute({
      getComponentModule: () => System.import('containers/HomePage'),
      path: '/',
      name: 'home',
    }),

   return routes;
}

**Comment by michelalbers** [comment]
Please merge this :) The current approach is so confusing for new devs. As soon as you have routing implemented you will run into multiple injected sagas which will cause _slow_ errors.

**Comment by vicbell** [comment]
Any suggestions of use react-router v4 to do this?

**Comment by FoxxMD** [comment]
@petemill one more question for you:

Whenever I try to use a `LazyLoadReduxRoute` as a "passthrough component" IE it has a path but it just a shell for the index route and child routes contained in it the component never gets rendered. Using `LazyLoadReduxRoute` only seems to work when it is the final route to be rendered. Do you know why that might be?

**Comment by marsonmao** [comment]
I've also got the saga injection problem and found this PR quite convincing.
So that I implemented it into my react-boilerplate code but made a little modification.
The basic idea of utilizing onEnter and onLeave is the same, but I change the implementation of `this.loadedSagas`.

I created a `createRoute` helper function and move `loadedSaga` there, like this:
```
export function createRoute(store, path, name, onEnterPreProcess, importSagaPath, importReducerPath, importComponentPath) {
  let loadedSagas = null;
  return {
    path,
    name,
    onEnter(ns, rep, cb) {
      let willEnter = onEnterPreProcess(ns, rep);
      if (willEnter) {
        const ms = importSagaPath();
        ms.then((sagas) => { loadedSagas = injectSagas(sagas.default); });
        ms.catch(errorLoading);
      }
      cb();
    },
    onLeave() {
      if (loadedSagas) {
        loadedSagas.forEach((saga) => saga.cancel());
        loadedSagas = null;
      }
    },
    // ...omitting others
  };
}
```
My colleague suggest that using `this.loadedSagas` could be dangerous if the code that keeping the returned Route object did things like using the `onEnter` function on independent reference, then `this` would be undefined. Making `loadedSaga` a variable inside the helper function could prevent it.

Dattaya:
We switched to RR4 in 3.5 so this PR is no longer applicable, but thanks @adelsz and everyone who participated in the discussion.

PR #1424, opened 2017-01-09
chore(internals): do not check file sizes of main chunk and favicon

Dattaya:
Image in the `Header` component is a part of the example app and is wiped out on `npm run clean` so I didn't want to exclude it specifically in the webpack production config (that would also require changing somehow its name to start with some prefix, or just exclude all images). So we are left with only one warning after building the example app.


gihrig:
Over three days old without comment... - Merging.

Thanks @Dattaya :sparkles:

PR #1432, opened 2017-01-10
Move test

gihrig:
Now that we are focussing more on tests and requiring a high percentage of test coverage for CI builds to pass, there is more reason than ever to allow building when tests are incomplete. This is a pain for manual acceptance testing.

This PR moves toward a better single responsibility for build and test scripts. If you want to manually choose processes, select from `test`, `build` and `start:prod` If you know you want it all `start:production`.

This PR:
1. Removes the testing phase from `build`.
2. Adds a testing phase to `start:production.
3. Modifies `.travis.yaml` and `appveyor.yaml` to run tests in the build process after the above.
4. Clarifies appropriate docs.
5. Removes a minor (manual) acceptance testing annoyance from `selectors.test.js` generator.


internals/generators/container/selectors.test.js.hbs
```diff
@@ -5,6 +5,6 @@
 
 describe('makeSelect{{ properCase name }}Domain', () => {
   it('Expect to have unit tests specified', () => {
-    expect('Test case').toEqual(false);
+    expect('true).toEqual(false);
```
Dattaya: Found a redundant single quote in front of `true`, other than that, I support this PR.

gihrig:
Thanks @Dattaya ! Fixed that ✨

Dattaya:
Awesome 👍

gihrig:
@collaborators Can we get another approval and land this? It puts a finishing touch on @outdooricon's excellent work on tests in #1414. Which has been merged and will be included in 3.4.

Thanks ✨

Dattaya:
> Travis does not directly support yarn
> As described at https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn
> Travis looks for yarn.lock and maps npm calls to yarn, if present.

well, this line from the blog: `If your Node.js project has the yarn.lock file in the repository root, we proceed to install Yarn and execute the yarn command instead of the default npm install.` in my understanding means that Travis would automatically call `yarn install` if `.yarn.lock` is present no matter what's inside `.travis.yml` file, so that means no `npm` calls in the config file are going to be mapped to `yarn`.

We probably also need to correct a bit `.travis.yml` so we test in the `scripts` section:
```config
script:
  - node ./internals/scripts/generate-templates-for-linting
  - npm run test
  - npm run build
```

mxstbr:
Sorry for the late response, love this.

gihrig:
@Dattaya Thanks for looking into this 😄 

>that means no npm calls in the config file are going to be mapped to yarn.

- Not sure what to do with this... Replace all npm calls with yarn calls? I did that at first and it failed on my fork/Travis. So restored npm calls and it passed. 

> We probably also need to correct a bit .travis.yml so we test in the scripts section:

- Not sure why those were in `install` section. Moved to scripts.

- Travis builds succeed, now as before using `yarn` (on my fork). Not sure any of this has to do with travis.yaml...
image

Will open a new PR with move from install -> scripts. See #1455

Dattaya:
> Replace all npm calls with yarn calls?

For some reason I thought you wanted to restore npm completely, just running scripts with `npm` instead of `yarn`, I don't think it matters. So everything's fine.

PR #1437, opened 2017-01-10
Fix #1435 comment

### LINKED ISSUE #1435: Remove unknown reference in index.html
In `./app/index.html` there is this comment:

```html
<!-- A lot of magic happens in this file. HtmlWebpackPlugin automatically includes all assets 
(e.g. bundle.js, main.css) with the correct HTML tags, which is why they are missing in this HTML
 file. Don't add any assets here! (Check out webpackconfig.js if you want to know more) -->
```

Since there doesn't seem to be a `webpackconfig.js` file, do we want to replace it with a reference to 
[this file](https://github.com/mxstbr/react-boilerplate/blob/master/internals/webpack/webpack.dev.babel.js#L17) 
or should we remove `(Check out webpackconfig.js if you want to know more)`?

kelsonic:
Fixes #1435 comment issue.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -18,6 +18,6 @@
     <div id="app"></div>
     <!-- Open Sans Font -->
     <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,700" rel="stylesheet">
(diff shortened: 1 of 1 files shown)


kelsonic:
Sure thing! 🍺

PR #1454, opened 2017-01-13
Move test

gihrig:
Minor edits to `.travis.yaml` as suggested [here](https://github.com/mxstbr/react-boilerplate/pull/1432#issuecomment-272398963).

Should have no effect on operation.


gihrig:
See #1455

PR #1484, opened 2017-01-19
Docs Update: sass.md | New webpack version -loader suffix

vedatmahir:
New webpack version no longer allowed to omit the '-loader' suffix when using loaders.

### DIFF
--- a/docs/css/sass.md
+++ b/docs/css/sass.md
@@ -14,7 +14,7 @@ to the loaders section in `internals/webpack/webpack.base.babel.js` so it reads 
    {
       test: /\.scss$/,
       exclude: /node_modules/,
-      loaders: ['style-loader', 'css-loader', 'sass-loader']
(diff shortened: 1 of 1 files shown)


KarandikarMihir:
@vedatmahir Which issue # does this PR belong to?

vedatmahir:
@KarandikarMihir sorry! Issue #1485

KarandikarMihir:
Ah! Thanks for contributing. :+1: :innocent:

vedatmahir:
@KarandikarMihir :) thanks.

I've just changed `loaders` to `loaders/rules`. Because codebase still use 'loaders' instead of 'rules'.
Btw, should i open a new issue for webpack 2 migration ?

KarandikarMihir:
> Because codebase still use 'loaders' instead of 'rules'.

@vedatmahir Yeah I saw that, good call! @gihrig takes care of updating dependencies from time to time. I should leave that up to him.

PR #1494, opened 2017-01-20
Add Stateless Functions to Container Generator

outdooricon:
Since we can choose to generate PureComponents or Components, this adds the ability to choose Stateless Functions as well.


mxstbr:
Getting errors on CI here?

```
SyntaxError: unknown: Unexpected token (16:58) Error transforming file: app/containers/RbGeneratedContainerPureComponent/index.js
SyntaxError: unknown: Unexpected token (16:62)
```

outdooricon:
@mxstbr That should take care of it. The only failures now are about using const in node 5, which isn't anything new that I added

internals/generators/container/index.js
```diff
@@ -20,10 +20,10 @@ module.exports = {
     },
   }, {
     type: 'list',
-    name: 'component',
-    message: 'Select a base component:',
-    default: 'PureComponent',
-    choices: () => ['PureComponent', 'Component'],
+    name: 'type',
+    message: 'Select the base component type:',
+    default: 'Stateless Function',
```
samit4me: This will vary between project and developer (let's put it up for discussion) but I find that the 'Class Component' is by far the most popular way to define container components. Would love to hear what others do?

internals/generators/container/index.js
```diff
@@ -20,10 +20,10 @@ module.exports = {
...
+    choices: () => ['Stateless Function', 'PureComponent', 'Component'],
```
samit4me: Should we rename 'Component' to 'Class Component' just to be clear?

KarandikarMihir: @samit4me  [React](https://facebook.github.io/react/docs/react-api.html#react.purecomponent) calls them `React.PureComponent` and `React.Component`, because both of them are classes.

samit4me: Yeah, that makes sense, after reading again I tend to agree, thanks @KarandikarMihir

gihrig: 👍  Agree on 'React.Component', 'React.PureComponent' and 'Stateless Function'.

Reactland has a naming problem with 'Component', JS vs CSS has a problem with 'Class', plain JS functions as 'Functional Components' has a problem with 'Functional Programming', etc. 

It's important (and often hard) to name things correctly 😄

internals/generators/container/index.js
```diff
@@ -47,10 +47,22 @@ module.exports = {
   }],
   actions: (data) => {
     // Generate index.js and index.test.js
+    let componentTemplate;
```
samit4me: The build is failing using node 5 (we support the last 3 versions, so 5, 6, 7). Might be best to use `var` here instead

internals/generators/container/index.js
KarandikarMihir: God I love web development!

internals/generators/container/index.js
outdooricon: What I don't understand is that this code is [doing the same thing that the component generator](https://github.com/react-boilerplate/react-boilerplate/blob/master/internals/generators/component/index.js#L39) has been doing already. Why fail here and not there?

samit4me:
Spent 15 minutes trying to diagnose, but I am not familiar with this code, so I'm not much help.

If you run `node ./internals/scripts/generate-templates-for-linting` locally you should be able to replicate the problem.

It is possible to view the CI error log via the details link on the failed build or click this link https://travis-ci.org/react-boilerplate/react-boilerplate/jobs/204570975

outdooricon:
Really bizarre, ok I'll clean it up soon

outdooricon:
@samit4me lost track of this pr, sorry I took so long... it should be ready now

samit4me:
LGTM, thanks for getting it working 🎉

samit4me:
@gihrig @KarandikarMihir Are you happy for this to be merged?

gihrig:
I'm swamped with work so I haven't tested thoroughly, but overall it looks good. @samit4me if you have worked with the code and feel that it's ready merge, I won't object. 😄

KarandikarMihir:
@samit4me @gihrig @outdooricon I just tested this. And it works like a charm! :smile: 

Why are we not using arrow functions for stateless functions? Just curious.

samit4me:
We pretty much follow the Airbnb React/JSX Style Guide and it recommends a normal function over arrow function, see https://github.com/airbnb/javascript/tree/master/react#class-vs-reactcreateclass-vs-stateless, so if you're happy with this, let's get this merged.

KarandikarMihir:
@samit4me Thanks! Yes. Let's merge this.

Issue #1287, opened 2016-11-29
Remove name property from routes


Stalso:
Hello. All your routes in `routes.js` have `name` property. 

    {
      path: '/',
      name: 'home',
      getComponent(nextState, cb) {
        const importModules = Promise.all([
          System.import('containers/HomePage/reducer'),
          System.import('containers/HomePage/sagas'),
          System.import('containers/HomePage'),
        ]);

        const renderRoute = loadModule(cb);

        importModules.then(([reducer, sagas, component]) => {
          injectReducer('home', reducer.default);
          injectSagas(sagas.default);

          renderRoute(component);
        });

        importModules.catch(errorLoading);
      }

 But I can not understand why. This property is not mentioned in `react-router` docs. As I see, it was removed from `react-router`. Why do we need it?

mxstbr:
To be completely honest, I don't know! Can you try removing it from the example app and seeing if anything changes?

dispix:
Hi,

Was taking a look at some issues and stumbled upon this one. The `name` property is deprecated in `react-router` (see https://github.com/ReactTraining/react-router/issues/1840). So I think this can be safely removed without impacting anything.

mxstbr:
Would appreciate a PR!

beardedtim:
Working on a PR now. Should be done with it today. Is anyone else working on this before I make a PR?

gihrig:
Closed by #1509

PR #1509, opened 2017-01-22
removed route names from app/routes

beardedtim:
This commit will close issue #1287 

I have ran tests and it looks like this does not affect anything.


beardedtim:
@KarandikarMihir didn't even catch that! I suppose I should have dove deeper into the codebase before making the PR. I made that edits and pushed to the same branch. Let me know if I was silly and missed anything else!

KarandikarMihir:
@beardedtim It's alright! 😄  `templates` is hidden away from the core parts and it's only natural that you didn't notice it. Here's my LGTM to your latest commit. 👍 :wink: 

I've read every comment from the [original issue](https://github.com/ReactTraining/react-router/issues/1840) just to make sure that this change is indeed harmless. If anybody thinks otherwise, now is the time.

KarandikarMihir:
@beardedtim Hi. There's one more change that we should make. Please remove [this line](https://github.com/react-boilerplate/react-boilerplate/blob/master/internals/generators/route/route.hbs#L3) so that generator won't create `name` property.

beardedtim:
@KarandikarMihir I removed it from there and from `routeWithReducer.hbs` because I figured if we are removing it from all routes that the reducer doesn't need it either.

KarandikarMihir:
@beardedtim Good catch! Thanks very much 👍

KarandikarMihir:
@beardedtim Good catch! Thanks very much 👍 

@gihrig Can you please take a look at this? It's ready.

PR #1521, opened 2017-01-25
fix(eslint): Enable rule react/no-array-index-key

### LINKED ISSUE #1503: New eslint-plugin-react rules complain "error Do not use Array index in keys"
Updated eslint-plugin-react rules disallow use of array index as key.

Error: ~/react-boilerplate/app/components/List/index.js 13:31
error Do not use Array index in keys react/no-array-index-key

Disable "react/no-array-index-key" in package.json
See: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md

To reproduce:
confirm #1501 has been merged
checkout current `dev` branch
remove  "react/no-array-index-key" from package.json
`npm test`

Boilerplate 3.4+ `dev` branch (ad871d7a1b70cec5b365fdc30eb86ee7e189158f)
Node 7.4.0
npm 4.1.1
yarn 0.19.1
Mac OS 10.11.6
Chrome 55.0.2883.95 (64-bit))

carloscuatin:
Fixes #1503

### DIFF
--- a/app/components/List/index.js
+++ b/app/components/List/index.js
@@ -9,8 +9,8 @@ function List(props) {
 
   // If we have items, render them
   if (props.items) {
-    content = props.items.map((item, index) => (
-      <ComponentToRender key={`item-${index}`} item={item} />
+    content = props.items.map((item) => (
+      <ComponentToRender key={`item-${item.id}`} item={item} />
     ));
   } else {
     // Otherwise render a single component
--- a/app/components/List/tests/index.test.js
+++ b/app/components/List/tests/index.test.js
@@ -14,11 +14,12 @@ describe('<List />', () => {
 
   it('should render the items', () => {
     const items = [
-      'Hello',
-      'World',
+      { id: 1, name: 'Hello' },
+      { id: 2, name: 'World' },
     ];
+    const item = ({ name }) => <div>{name}</div>; // eslint-disable-line react/prop-types
     const renderedComponent = render(
-      <List items={items} component={ListItem} />
+      <List items={items} component={item} />
     );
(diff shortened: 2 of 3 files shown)


carloscuatin:
@gihrig the component `ListItem` don'ts accept a `object` example:

```js
const items = [
  { id: 1, name: 'Hello' },
  { id: 2, name: 'World' },
];
```

Which would be the best option to pass another component that accepts the property established by the fake state ?

gihrig:
@carloscuatin Off hand, I don't have an answer for that...

Anyone else?

carloscuatin:
```js
const items = [
  { id: 1, name: 'Hello' },
  { id: 2, name: 'World' },
];
const Item = ({ name }) => <div>{name}</div>;
const renderedComponent = render(
  <List items={items} component={Item} />
);
```

possible solution @gihrig

gihrig:
@carloscuatin That looks good, and it works ✨ 

A few issues though:
1. `const Item` Item is not an object or Component, it should be lowercase.
2. Eslint has an issue with 
```javascript
const item = ({ name }) => <div>{name}</div>;
```
image

Suggest adding ` // eslint-disable-line react/prop-types`

KarandikarMihir:
@carloscuatin I tried the exact same thing, but it didn't work 😕 Maybe I missed something. Let's go with this if it works 👍

gihrig:
@KarandikarMihir 

>  I tried the exact same thing, but it didn't work

What happened?

Current iteration works for me...

@carloscuatin Another win for code quality ✨  
Thanks for hanging in there through all the back and forth!

Issue #1462, opened 2017-01-15
Show reducer warning after generating a component


Necmttn:
- [x] Bug 

## Description

![image](https://cloud.githubusercontent.com/assets/5212808/21961039/fc5b5a8c-db39-11e6-82ca-478715bec8f6.png)

## Steps to reproduce

git clone https://github.com/mxstbr/react-boilerplate.git
yarn
npm run start
npm run clean
npm run generate

? What should it be called? FormTwo
? Select a base component: PureComponent or Component  (* both have same Issue)
? Do you want headers? Yes
? Do you want an actions/constants/selectors/reducer tuple for this container? Yes
? Do you want sagas for asynchronous flows? (e.g. fetching data) Yes
? Do you want i18n messages (i.e. will this component use text)? Yes

than import FormTwo to homepage and render it
it will throw error.

# Versions

- React-Boilerplate (see `package.json`): 3.4.0
- Node/NPM: 7.1.0
- Browser: Chrome 55

Necmttn:
I just understood where i was doing wrong. 
seemslike we need to add our reducer to app/reducers.js. 
it would be nice after npm run generate finish it warn the user for add reducer to reducers.js

KarandikarMihir:
Yeah, that is because reducers/sagas won't be injected asynchronously in the root saga/root reducer when you import a container inside another container. 

Ideally, a container should only be imported async in `routes.js` when you go to its corresponding route. But for the sake of clarity, showing a warning after plop would be a nice-to-have.

kylesavant:
See: https://github.com/mxstbr/react-boilerplate/issues/1407

Necmttn:
@KarandikarMihir  on weekend I will try to prepare warnings for it.

KarandikarMihir:
@Necmttn Hey! Any progress on this one?

Necmttn:
Well there's several options for error. 
what we found so far is 
1)
![image](https://cloud.githubusercontent.com/assets/5212808/22304272/b2d9a7fc-e371-11e6-8f18-3930f448ad00.png)

```
      // Reducer
      actions.push({
        type: 'add',
        path: '../../app/containers/{{properCase name}}/reducer.js',
        templateFile: './container/reducer.js.hbs',
        abortOnFail: true,
      });
      actions.push({
        type: 'add',
        path: '../../app/containers/{{properCase name}}/tests/reducer.test.js',
        templateFile: './container/reducer.test.js.hbs',
[...]

Issue #1407, opened 2017-01-04
Receive error "Cannot read property 'toJS' of undefined" when importing a new container


kylesavant:
Steps to reproduce:
```
1.) Clone project: git clone --depth=1 https://github.com/mxstbr/react-boilerplate.git

2.) Setup project: npm run setup

3.) Create a new container:
npm run generate container
? [CONTAINER] What should it be called? SideNav
? [CONTAINER] Select a base component: PureComponent
? [CONTAINER] Do you want headers? No
? [CONTAINER] Do you want an actions/constants/selectors/reducer tupel for this container? Yes
? [CONTAINER] Do you want sagas for asynchronous flows? (e.g. fetching data) Yes
? [CONTAINER] Do you want i18n messages (i.e. will this component use text)? No

4.) Import new container in 'app/containers/App/index.js':
Add line: import new container from 'containers/SideNav';
Below: import Footer from 'components/Footer';

5.) Call container in return statement for app:
Add line: <NewContainer />
Below: <AppWrapper>

6.) Run app: npm start

7.) Open app in browser: http://localhost:3000
```

Expected results:
.Expect to see the new container when I run the app, with no errors in console

Actual results:
.The page does not render
.I get this error in chrome inspector console:
```
Uncaught (in promise) TypeError: Cannot read property 'toJS' of undefined
    at eval (eval at ./app/containers/SideNav/selectors.js (main.js:1009), <anonymous>:24:20)
    at eval (webpack:///./~/reselect/es/index.js?:62)
    at eval (webpack:///./~/reselect/es/index.js?:24)
    at selector (webpack:///./~/reselect/es/index.js?:73)
    at Connect.configureFinalMapState (webpack:///./~/react-redux/lib/components/connect.js?:154)
    at Connect.computeStateProps (webpack:///./~/react-redux/lib/components/connect.js?:141)
    at Connect.updateStatePropsIfNeeded (webpack:///./~/react-redux/lib/components/connect.js?:203)
    at Connect.render (webpack:///./~/react-redux/lib/components/connect.js?:338)
    at eval (webpack:///./~/react/lib/ReactCompositeComponent.js?:793)
    at measureLifeCyclePerf (webpack:///./~/react/lib/ReactCompositeComponent.js?:74)
```

Am I doing something wrong here?

kylesavant:
I see that if I inject my new container reducers and sagas into routes.js for a specific route the error goes away. But how do I inject these reducers/sagas into all routes?

[...]

PR #1523, opened 2017-01-26
feat(internals): Improve generator warnings  (#1462)

Necmttn:
fixes the conflict at (#1407, #1462) @gihrig  please check. 

Documented in [gotchas.md](https://github.com/react-boilerplate/react-boilerplate/blob/dev/docs/general/gotchas.md#load-reducers-optimistically).


gihrig:
@Necmttn There are lint errors when run against `dev` branch. Please run `npm test` on your fork and correct those.

KarandikarMihir:
The message could be improved. We need something concise but descriptive at the same time 😅

> ⚠️ Please pass this reducer explicitly to combineReducers in reducers.js if you don't want to inject it asynchronously.

Or

> ⚠️ Please pass this reducer explicitly to combineReducers in reducers.js if you want to use it out of the box.

Or something on similar lines. Perhaps someone with prowess in English can suggest better wording?
@mxstbr We could make use of your inputs on this one.

gihrig:
I would suggest a concise one-liner, since this is going on an already busy screen.

> Remember to add this reducer to combineReducers in app/reducers.js

KarandikarMihir:
@gihrig Yup! This one's better.

>Remember to add this reducer to combineReducers in app/reducers.js

gihrig:
@Necmttn 

I vote for:

1. Message in terminal when running generator:

`⚠️ Remember to add this reducer to combineReducers in app/reducers.js`

2. Add the same text as a code comment at the top of the generated `reducer.js`.

KarandikarMihir:
I second @gihrig.

KarandikarMihir:
Umm. 

>⚠️ Remember to add this reducer to combineReducers in app/reducers.js


But it doesn't quite say that you don't _HAVE TO_ add this if you're going to go for asynchronous injection. I'm sorry I'm being a pest here.

gihrig:
>  I'm sorry I'm being a pest here.

Yes, you are 😄 

⚠️ Not loading in a route?, Add to app/reducers.js: combineReducers. Ref #1407.

Better still, make this [#1407Comment](https://github.com/react-boilerplate/react-boilerplate/issues/1407#issuecomment-271066646) a gotchas.md entry and reference that.

KarandikarMihir:
😅 😄 This better!! It shouldn't create any confusion for the users. Thank you, @gihrig 

> Better still, make this #1407Comment a gotchas.md entry and reference that.

I'll do that.

OsmanMesutOzcan:
I have a custom prompt that seamlessly warns without the need for another confirmation. You can check it out. https://github.com/OsmanMesutOzcan/inquirer-confirm-warning

KarandikarMihir:
@OsmanMesutOzcan Hi! You've got a cool tool there. But using that would mean adding one more dependency in the project. We generally tend to avoid that.

Furthermore, we just want to display one line, without needing any confirmation from the user. So there's no point in adding interactivity here. Thanks for the suggestion, though :)

OsmanMesutOzcan:
@KarandikarMihir The current solution actually needs an extra confirmation. Thats why i suggested an option. You can just show an extra line using it.

Necmttn:
@gihrig 
I agree with you. 
I just made changes accordingly. any other suggestions  or good to go ?

KarandikarMihir:
@Necmttn Hi. Isn't it 

> :warning: Not loading in a route?, Add to app/reducers.js: combineReducers. Ref #1407.

As per [this comment](https://github.com/react-boilerplate/react-boilerplate/pull/1523#issuecomment-275830051)

gihrig:
@Necmttn I agree with @KarandikarMihir regarding warning text.

@KarandikarMihir Will need to complete that [gotchas.md entry](https://github.com/react-boilerplate/react-boilerplate/pull/1534#issuecomment-276139672) so it can be referenced in a code comment.

gihrig:
@Necmttn Re: the conflict, it's OK to remove `yarn.lock` from the PR. It has no bearing on the proposed changes.

Necmttn:
@gihrig 
for remove file from PR i follow, http://stackoverflow.com/a/9498304/6942552
i hope is the correct way to do it.

gihrig:
>for remove file from PR i follow, http://stackoverflow.com/a/9498304/6942552
>i hope is the correct way to do it.

Seems to have worked 😄 

@Necmttn I closed/opened the PR (to force a rebuild) in hopes the AppVeyor failure was a transient issue.

Question though, what's your dev platform? Mac, Windows, Linux?

gihrig:
@KarandikarMihir I'm really tied up with work ATM, can you check this out?

gihrig:
Working with this I find the prompt to be a likely source of confusion:
image

How are 'Y' and 'N' answers different? Of course there's no difference.

I think I would prefer to eliminate the prompt, thinking the yellow warning and the code comment are sufficient to draw attention without adding an extra keystroke to the generate process.

Alternatively the prompt could be changed to simply 'OK'.

Or, more complex (probably best), ask if the developer wants to have the reducer wired up and do so if requested.

Thoughts?

KarandikarMihir:
@gihrig Same here. I had tested this but couldn't find time to point it out here. I agree with you. It should say 'OK' or nothing really, if that's possible.

@Necmttn Can you please make these changes?

Necmttn:
the ``(y/N)`` is default value coming from [Inquirer]( https://github.com/SBoudrias/Inquirer.js/#confirm---type-confirm)   and this is the most simple prompt they have in the library.

I don't know any other way around.

gihrig:
> the (y/N) is default value coming from Inquirer

@Necmttn Can we just not use Inquirer and use Chalk to color the message?

Necmttn:
@gihrig , now it's only show's the message without ``(y/N)`` but still you need to press enter for next prompt.

gihrig:
@Necmttn Can we remove the need for the "mysterious" enter?

The page content is short enough that it won't be missed.

Necmttn:
Alternative-)
image
if we do like this.  the location of (y/N) is changing. but we don't need to press ``"mysterious" enter`` but i think this will cause miss direction the user.  because it will be shown even user doesn't want reducer. 

```
{
    type: 'confirm',
    name: 'wantActionsAndReducer',
    default: true,
    message: `Do you want an actions/constants/selectors/reducer tuple for this container?
  ${chalk.yellow('⚠️ Not loading in a route?, Add to app/reducers.js: combineReducers. Ref #1407.')}`,
  },
```
With Current Version we only show warning when user want to add reducer.  
a fixes can be we can add ``[ENTER] `` end of string so UX could improve.

image
something like this.

gihrig:
@Necmttn 

 - I don't understand the color change (compared to [here](https://github.com/react-boilerplate/react-boilerplate/pull/1523#issuecomment-279830518)) shown in your last post. Is this a proposed change, or just the way it appears on your terminal?

- > With Current Version we only show warning when user want to add reducer.
a fixes can be we can add [ENTER] or [Okay ?] end of string so UX could improve.

Is there no way to simply display the message only if the user wants a reducer without making it an additional prompt?

If that's true a suitable prompt might be "OK" or "Got It!" But from a UX perspective, it's just warning requiring on user action.

Necmttn:
@gihrig  it's just how it's appears on my new terminal settings 😄 
there's no way unless, inquirer.js supports [Warn] prompt ( so far I know )

`` Got It! `` sound cool to me.

KarandikarMihir:
@Necmttn This is one of the long-pending PRs. Do you think we can make it to the finish line? :smile: Let me know.

KarandikarMihir:
Closing this due to inactivity. Feel free to reopen.

Issue #1213, opened 2016-11-08
Discussion: Style-guides


spen:
Hi! 👋

A big issue that we're facing on a project is the lack of style-guides to refer to during code reviews - leading to a lot of inconsistencies and anti-patterns.

I noticed that theres a tracking issue (#14) that's been closed but with intent to carry on adding docs on as-they-come basis. I was also happy to see the AirBnB guide referenced 👏

I'm very keen to kick off a PR to get something added but wanted to get a discussion going first to see what peoples thoughts are right now, since (#14) is now quite old!

Some thoughts:
- AirBnB's JS style guide is awesome so I +1 using them as default.
- Many existing projects may throw lint errors/warnings if we enforce all of AirBnBs rules
- As per the above, it could be worth encouraging users to only lint & update files as they're touched & provide details on that work flow.
- Tweaking or Removing the guidelines/rules RBP provide should be hassle free & clearly instructed.
- Do we copy in the whole file? Reference it? Agree on and create a subset?
- Do we add RBP specific rules?

samit4me:
Completely agree that a JS style guide that automatically enforced is invaluable. I'm not sure if I understand your question completely tho, as this project is currently using both:
- Airbnb JavaScript Style Guide
- Airbnb React/JSX Style Guide

See [package.json](https://github.com/mxstbr/react-boilerplate/blob/master/package.json) [here](https://github.com/mxstbr/react-boilerplate/blob/master/package.json#L85-L88) and [here](https://github.com/mxstbr/react-boilerplate/blob/master/package.json#L236-L242)! This will detect errors during the build step and it should also highlight problems in your editor if you have the appropriate plugin install. I feel as tho I am missing something here tho???

spen:
Ahh, I should have been more clear - sorry!
I mean in terms of docs, a .md file that explains the guidelines in a wordy way, wether that be a copy of the Airbnb GLs, a short write up that justifies it and references it or something along those lines.
The eslint additions are great but I feel there's room for some docs that support them - could be wrong though, having the eslint rules ship out with the boiler plate might just be enough!

samit4me:
That would be great, maybe just a quick summary with a link to the Airbnb Style Guide, followed by a brief description of the overrides! Is that more like what you think should happen?

spen:
Yeah! That sounds pretty decent! :D
I'll put together a PR this evening to get things rolling!

KarandikarMihir:
[...]

PR #1533, opened 2017-01-29
Docs: Add JS guidelines doc with first few rules explained

spen:
As talked about in #1213, here is a 'first draft' of a JS style guide doc.
I've only added a few explanations of deviations from AirBnB for now - more could be added over time, whether that's before or after a merge.

I realise that I might not be in the best position to really see this through on my own, mostly because my time has been spent working on non RBP projects for some months now 😕
Also, there are eslint rules that I'm not so sure about and so can't write accurate justifications for them.

Happy to hand this over, let it go (if it's not deemed important) or just work with others who know better than I do - what ever works best! 🙂

### DIFF
--- a/docs/js/guidlines.md
+++ b/docs/js/guidlines.md
@@ -1,0 +1,60 @@
+# JavaScript Style Guidelines
+
+React Boilerplate subscribes to AirBnB's well established [JavaScript](https://github.com/airbnb/javascript) and [React](https://github.com/airbnb/javascript/tree/master/react) style guides with a few tweaks
+
+These tweaks are defined in the [`package.json`](../../tree/master/package.json) and some are explained throughout this document.
+
+While it's good practice to follow a common code style, the most important thing is that you and your team agree on a consistant coding style. The default style guides included here in React Boilerplate are highly recommended, but if your team would like to change some rules - feel free to do so!
+
+## Overrides
+
+### arrow-parens 
+
+eslint: [`arrow-parens`](http://eslint.org/docs/rules/arrow-parens)
+
+> Why? Not sure
+
+```js
+// Bad
+a => {}
+
+// Good
+(a) => {}
+```
+
+### arrow-body-style 
+
+eslint: [`arrow-body-style `](http://eslint.org/docs/rules/arrow-body-style)
+
+> Why? In cases where a function returns a single expression the function block and explicit return statement can be ommited, leaving just the expression.
+
+```js
+// Bad
+const foo = () => {
+    return 'something';
+};
+
+// Good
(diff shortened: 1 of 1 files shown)


KarandikarMihir:
This is looking great, @spen! 😄 👍  Thanks for taking out time for this, we really do appreciate it. You can update the doc one rule at a time, at your own pace. No rush! In the meantime, I'll try to follow your style and add some more rules.

docs/js/guidlines.md
```diff
@@ -0,0 +1,60 @@
...
+
+While it's good practice to follow a common code style, the most important thing is that you and your team agree on a consistant coding style. The default style guides included here in React Boilerplate are highly recommended, but if your team would like to change some rules - feel free to do so!
+
+## Overrides
+
+### arrow-parens 
+
+eslint: [`arrow-parens`](http://eslint.org/docs/rules/arrow-parens)
+
+> Why? Not sure
```
mxstbr: Consistency 😉

mxstbr:
This is awesome, feel free to take your time! (though I can't wait to land this)

docs/js/guidlines.md
spen: :P I thought that might be the case but wasn't sure!
More specifically, is it for consistency with 2+ arg functions? `( one ) =>` & `( one, two ) =>`?
I think that's a fair comment so I'll update this bit in the morning (it's getting late in Sydney!)

spen:
Thanks for your initial thoughts guys :)
I'm sure it won't take much longer than a week or two anyway - I might just have to throw plenty of Q's about individual rule choices :)

Also, what do you think to addressing guidelines by eslint rule? I was going for that originally but AirBnBs guide tends to talk about a subject, then list related rules... this might be a nicer approach for readability... I'll give that a go in my next batch!

gihrig:
> what do you think to addressing guidelines by eslint rule?

Great idea! Considering that RBP style is AirBnB with exceptions, as listed in `package.json`, documentation and reasoning by rule seems the logical choice. That's where/how RBP devs are likely to encounter the subject.

An overall philosophy regarding these choices would be good as an intro, we already have that from docs/general/README.md [Introduction](https://github.com/react-boilerplate/react-boilerplate/blob/master/docs/general/README.md#introduction):

>Using React Boilerplate, you get to start your app with our community's current ideas on what represents optimal developer experience, best practice, most efficient tooling and cleanest project structure.

spen:
Sorry folks, I've not got enough time to follow up on this especially as it's likely a bit dated by now.
Feel free to commandeer <3

PR #1538, opened 2017-01-30
Move root route configuration to routes.js, add documentation for creating a global saga

RobertSheaO:
The route configuration was previously split between `app.js` and `routes.js`. I've moved the `App` (rootRoute) configuration to the `routes.js` file, so it is all in one place. I also changed loading the root route to use `getComponent`, instead of `component`, to make it easier for people to implement a global saga if they want to do that. There is also new documentation on how to add a global saga. I've also updated the spacing for the route generators and tested them.

Issue #1447, opened 2017-01-13
Discussion: Use redux-logic instead of redux-saga?


andreawyss:
Please take a look at redux-logic. 
It has all the features of redux-saga and more with a much simpler declarative syntax. 
What do you think about switching to it?

gihrig:
> What do you think about switching to it?

Well, it's still early days for `redux-logic`. We'll have to wait for a while and see where it goes. I would not consider switching until it has proven itself more. 

If this discussion generates enough interest and someone submits a PR, which members of the community then find to be a better solution than `redux-saga`. Then it would be considered (that's my opinion, what _I_ think).

The concepts from the [project page](https://github.com/jeffbski/redux-logic) look interesting, but the proof is always in the implementation.

It seems the developer of `redux-logic` is working on integrating his project into a [fork](https://github.com/jeffbski/react-boilerplate/tree/redux-logic) of `react-boilerplate`. I'd be interested to hear about people's experience with that project, particularly if they have previous experience with `react-boilerplate` using `readux-saga`.

Ref: #1315 "Daemon Sagas"
Ref: #1420 "Alternative Saga Lifecycle".

jeffbski:
While searching for the best way to create business logic with Redux, I ended up not being completely happy with any of the common choices (fat action creators, thunks, sagas, epics (redux-observable)) which I [chronicled here](http://codewinds.com/blog/2016-08-16-business-logic-redux.html). So I ended up building redux-logic about 5 months ago and have been very happy with the results. 

It is a young project but I'm getting really good feedback from people that have tried it. Most people report that their code is greatly simplified and more understandable than their previous versions. Other than the ongoing exercise of adding more docs and examples and adding TypeScript support, there's nothing major that is outstanding, things have been stable.

My main goal was to make it simpler to implement business logic while still having some great built-in features like cancellation, latest filtering, debouncing, and throttling. So I made those declarable and then exposed an API where you can write your business logic for both intercepting (validating/transforming/augmenting) actions as well as async processing. 

The nice thing is that you can use the style of JS that you like for your team, you don't have to be proficient in generators or observables. You can use a style you are used to whether that is: callbacks, promises, async/await, or observables. Under the covers I implement the declarable features using observables but those are simply enabled by setting a properties on the logic, the redux-logic developer writes their business logic code in whatever JS style their team is most comfortable with. 

[...]

Issue #1384, opened 2016-12-26
RFC: Remove per-route sagas


mxstbr:
While our current per-route, codesplit saga setup is awesome (thanks @justingreenberg!) it's sadly let to quite a bit of confusion. (see issues #1277, #1274, #1183, #1137 and many more for examples)

What do we think about removing per-route sagas, adding a global `app/saga/` folder and letting people handle the saga lifecycles themselves however they want to?

This would then allow us to get rid of `react-router-redux`, as the only reason we currently have it is `redux-saga`. That, in turn, would then allow us to remove ImmutableJS by default since the only reason we have it is to show the integration with `react-router-redux`, as it's not that easy. Without that, integrating ImmutableJS is super easy, so there's no need to ship with it by default. (ref #441)

Between getting rid of all the support requests and the much simpler runtime setup, I feel like this would be a very beneficial change. What do you think @justingreenberg @gihrig et al.?

ipselon:
> Between getting rid of all the support requests and the much simpler runtime setup, I feel like this would be a very beneficial change.

I totally agree.

gihrig:
I agree, Anything to streamline and simplify is welcome.

justingreenberg:
@mxstbr i **totally support** this decision :)

outdooricon:
I don't know if other users feel this way but I've always looked up to boilerplate as the "right way" to have my react projects set up. It's not so much of an example for me as it is more of the de facto project and I feel like others look up to it in the same way. If ImmutableJS is removed, then I wonder if others will no longer think it is necessary to use since boilerplate isn't using it., which would a shame considering its optimizations when integrated with reselect and react-redux.

AustinMCrane:
yes, i was just about to ask how to add sagas and reducers to a container that doesnt have a separate route (ie: Navbar)

Dattaya:
Hard to know if it's going to reduce the number of new issues or increase it.

> This would then allow us to get rid of react-router-redux

We can remove `react-router-redux` right now and replace it with a simple history listener:

```js
  browserHistory.listen(location => store.dispatch({ type: 'NAVIGATE', location }));
```

[...]

Issue #1277, opened 2016-11-28
How is loaded the global state and is there a global sagas ?


kopax:
I'm having a look at the selectors of App containers and I see : 

    const selectGlobal = () => (state) => state.get('global');

But I can't find where this is loaded into react the 'global' state of the application.

When I was looking on how to add a global states a month ago, I did had to inject my global reducers and global sagas inside the `createRoutes` function.

Is it global just because our landing is App ? Or did you directly load it somewhere ?

I have a logout action and I want it to be stored under a App/sagas.js, is there an equivalent for injecting a globalSagas ?

samit4me:
Is this what your looking for:

```
/**
 * Creates the main reducer with the asynchronously loaded ones
 */
export default function createReducer(asyncReducers) {
  return combineReducers({
    route: routeReducer,
    global: globalReducer,
    language: languageProviderReducer,
    ...asyncReducers,
  });
}
```

https://github.com/mxstbr/react-boilerplate/blob/master/app/reducers.js#L41-L51

kopax:
Thanks! 
Sorry for asking. I don't know why I've looked for `globalReducer` and sublim didn't find this reference.

kopax:
Is there an equivalent for the sagas ?

zamrq:
simplify your life

```javascript
// selector.js
const selectGlobal = () => (state) => state.toJS();

const selectFormContainer = () => createSelector(
  selectFormContainerDomain(),
  selectInitialValues(),
  selectGlobal(),
  (substate, initialValues, global) => Object.assign(substate.toJS(), { initialValues, global })
);
```

const { global } = this.props;
console.log(global) // global state

samit4me:
You are correct, routes.js is where all the action happens! In short, there is no such equivalent. Sagas are different to the reducers.

There can only be a single store and to know the shape of that store you need all the reducers up front, so on route load, if there is a reducer, it is combined with the other reducers and then the entirely new store is generated including existing state and then this new store replaces the existing store.

[...]

PR #1545, opened 2017-01-30
Fix #1518: Move root route configuration to routes.js, add documentation for creating a global saga

### LINKED ISSUE #1518: How to configure App level saga?
At the `rootRoute` configuration in `app.js`, App is passed in as the base component for `react-router`, and the `childRoutes` inside `routes.js` are the JavaScript representation of the `react-router` route configuration where sagas are injected. How can I inject a saga at the `rootRoute` level?

What I'm trying to accomplish is the ability to load information from a server when the user enters the app, no matter which route they enter. I'm trying to avoid adding a `componentWillMount` to every container.

I've modified my `containers/App/index.js` to connect to react-redux, and I've added a `global` reducer in in `reducers.js`

**containers/App/index.js**
```
componentWillMount() {
  this.props.onLoadAccountDetails();
}
...
export function mapDispatchToProps(dispatch) {
  return {
    onLoadAccountDetails: () => {
      dispatch(loadAccountDetails());
    },
    dispatch,
  };
}
export default connect(mapStateToProps, mapDispatchToProps)(App);
```

**reducers.js**
```
...
import globalReducer from 'upside-rewards-common/lib/containers/App/reducer';
...
export default function createReducer(asyncReducers) {
  return combineReducers({
    route: routeReducer,
    global: globalReducer,
    language: languageProviderReducer,
    ...asyncReducers,
  });
}
```

RobertSheaO:
Fix for issue #1518. The route configuration was previously split between `app.js` and `routes.js`. I've moved the `App` (rootRoute) configuration to the `routes.js` file, so it is all in one place. I also changed loading the root route to use `getComponent`, instead of `component`, to make it easier for people to implement a global saga if they want to do that. There is also new documentation on how to add a global saga. I've also updated the spacing for the route generators and tested them.

Original Pull Request Here: https://github.com/react-boilerplate/react-boilerplate/pull/1538


gihrig:
@RobertSheaO Thanks for cleaning up the PR! ✨ 

I'll be watching the discussion in #1518.

KarandikarMihir:
This is awesome. Quick question- why is there no saga in `App/`? I had to remove `sagas` and `injectSagas(sagas.default);` from the route config. Precisely, [from here](https://github.com/RobertSheaO/react-boilerplate/blob/bb69de2e18fd690560f613bef001b1d606f8df37/app/routes.js#L65) to get it to work properly. 

I love the overall idea. Although, there seems to be a [working solution](https://github.com/react-boilerplate/react-boilerplate/issues/1277#issuecomment-275651735) for this particular problem, this PR solves two problems:
1. Creates a good, if not ideal, example for developers of implementing app-level sagas. 
2. Unifies route configuration in a single file.

That being said, I haven't studied other implementations. There's a [nice comment](https://github.com/justingreenberg) from @justingreenberg but that involves installing `redux-saga-router`. 

I'd love to know what others think of this.

RobertSheaO:
Ah, good catch. I'll update the PR with that change. As for the missing
app/sagas.js, I didn't want to put it in the base app, as not everyone will
need it.

KarandikarMihir:
@RobertSheaO No problem. Great job!

RobertSheaO:
@KarandikarMihir I agree that a built-in usage example would be great. Not sure about repo-loading, since that only happens on submit; however, I could perhaps add an example of preloading @mxstbr 's repos on app load. Thoughts?

gihrig:
> There's a nice comment from @justingreenberg

@KarandikarMihir The "nice comment" link got changed to JK's profile page ???

>preloading @mxstbr 's repos on app load. Thoughts?

@RobertSheaO How about pulling the current GH star count for react-boilerplate?

RobertSheaO:
Sounds good, I'll add that and update my PR

marcandrews:
Nice work!

KarandikarMihir:
@gihrig Sorry! Here's the [correct link](https://github.com/react-boilerplate/react-boilerplate/pull/1315#issuecomment-267888341)

gihrig:
There are a number of open issues surrounding difficulties with sagas and the way they are handled.

Ref: #1384, #1420, #1315, #1518, #1447 and this PR #1545.

I'm concerned about the lack of real-world experience behind these various proposals. The only two I know of having been applied are #1420 used in production by @adelsz and and #1447 used in production by @jeffbski, but not with react-boilerplate. @kopax has experience with [#1447Comment](https://github.com/react-boilerplate/react-boilerplate/issues/1447#issuecomment-273010133) and react-boilerplate. Perhaps he can update us with his experience so far.

I'm personally leaning toward #1447 (redux-logic) and possibly some combination of other proposals, but have not gotten far enough into this to make firm stand. If anyone is interested, I have updated @jeffbski's [fork](https://github.com/jeffbski/react-boilerplate-logic/tree/dev) to the 3.4 react-boilerplate release.

What I would really like to see, for is basic requirements of a simple addition to the example app to be drawn up, such as implementation of a simple login facility for example, then implemented in each proposal. I think the experience of those implementing the same set of requirements in the several proposed changes to RBP async handling would be enlightening.

RobertSheaO:
I've been busy these past few days; however, I'll whip up an example to put into the example app. I'll also read over the above linked issues to see how they impact what I'm doing. Personally, I like the simplicity of this solution, since it uses sagas in the same exact way as all the other sagas. For my particular use case, this is how I am using them:

1) Every route, except my login route, has an `onEnter` property that checks for a user token.
######routes.js
```JS
...
const { injectReducer, injectSagas, redirectToLogin } = getAsyncInjectors(store);
...
{
  path: '/',
  name: 'home',
  onEnter: redirectToLogin,
  getComponent(nextState, cb) {
  ...
  },
}
```

######asyncInjectors.js
```JS
function redirectToLogin(store) {
  return (nextState, replaceState) => {
    // Look into store for token, if invalid, replaceState with the `/login` path.
}
```

2) Inside my `App/index.js` file that wraps the entire app, I check the token validity again, and if it's valid and I don't already have the data, I send the request to retrieve data using the global saga.

######App/index.js
```JS
componentWillMount() {
  if (accessTokenIsValid) {
    if (!this.props.accountDetails) this.props.onLoadAccountDetails();
  }
}
```

RobertSheaO:
@marcandrews See my above comment about using `onEnter` (react-router) for authentication. This happens at the earliest point in the app, the route config, which could potentially prevent executing unnecessary code for users who are not authenticated.

marcandrews:
In the hopes of not detracting further away from @RobertSheaO and the awesome work he's done in this PR, I've created a separate issue #1554 in response to @gihrig's comment.

jackywhlee:
I've implemented this PR on my application because I needed the ability to logout via sagas, and I did not want to import 1 saga onto every component onto my app, and this solution works nicely for me.

docs/js/global-saga.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+Next we'll import the `sagas.js` file into our `App` route configuration by updating our `getComponent` function to asynchronosly load our files. You can just replace it with the one below.
+######app/routes.js
+```JS
+export default function createRoutes(store) {
+  // create reusable async injectors using getAsyncInjectors factory
+  const { injectReducer, injectSagas } = getAsyncInjectors(store);
+
+  childRoutes : [
```
avdeev: `childRoutes = [`

docs/js/global-saga.md
```diff
@@ -0,0 +1,63 @@
...
+## Usage
+
+We'll be doing the following:
+
+1. Adding `sagas.js` to `app/containers/App` directory
+2. importing `sagas.js` into the App container route configuration
+
+Like other sagas, your global saga will live within a container, the `App` container. Go ahead and add a file and use the following starter code; it's the same starter code for other sagas.
+
+######app/containers/App/sagas.js
```
avdeev: `###### app/containers/App/sagas.js`

docs/js/global-saga.md
```diff
@@ -0,0 +1,63 @@
...
+];
+
+// Individual exports for testing
+export function* sagaName() {
+
+}
+```
+
+Next we'll import the `sagas.js` file into our `App` route configuration by updating our `getComponent` function to asynchronosly load our files. You can just replace it with the one below.
+######app/routes.js
```
avdeev: `###### app/routes.js`

magiccookie:
please merge

mxstbr:
Deploy preview failed.

Built with commit 2e577930bf285a99b2eb997fa884d665315eb9d1

https://app.netlify.com/sites/react-boilerplate/deploys/59690735a700c4681563d721

justingreenberg:
@RobertSheaO thank you so much for your contribution :) as we have just upgraded to RR v4 and refactored `asyncInjectors`, this should be resolved (please see #1746) — i would love your feedback if you have the time to pull down `dev` branch and take it for a spin! thanks again

RobertSheaO:
I’ll take a look. Thank you.

circuitry2:
What happened to this? docs/js/global-saga..md no longer exists??? What is the recommended approach now?

PR #1553, opened 2017-02-03
(fix) ProgressBar state error

### LINKED ISSUE #1505: Warning: setState(...): Can only update a mounted or mounting component.
This appears in Chrome console in dev mode when the "Features" page is opened. I have not determined how to reproduce it reliably, but I have seen it several times while updating deps.

warning.js:36 Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the AppWithProgressBar component.

PR at #1553

To reproduce:
Confirm #1501 has merged <--- It has
Checkout latest dev branch
`npm start`
Open in bowser
Click "Features" link

Boilerplate 3.4+ `dev` branch (ad871d7a1b70cec5b365fdc30eb86ee7e189158f)
Node 7.4.0
npm 4.1.1
yarn 0.19.1
Mac OS 10.11.6
Chrome 55.0.2883.95 (64-bit))

KarandikarMihir:
"Expected" to fix #1505 

Navigating between routes would throw `Warning: setState(...): Can only update a mounted or mounting component.` It could not be reproduced reliably. After applying several solutions, the current solution "seems" to have solved the problem because after navigating for a gazillion times, I couldn't reproduce the error.

Please test and let me know if it works. Also, if possible, record your activity so that it'll help me track down the real cause.


gihrig:
LGTM! 💯 

Clone
Checkout `bfix/progress-state-error`
`yarn`
`npm run build`
`npm start`

Progress bar works as expected on first visit to `Features`, also on revisit after language change.

No errors! ✨ 

Node 7.4.0
npm 4.1.1
yarn 0.19.1
Mac OS 10.11.6
Chrome 55.0.2883.95 (64-bit))

smashercosmo:
@KarandikarMihir Could you please explain your solution? Cause I really don't get it. "bind" just returns bound function. How does it help in this case?

KarandikarMihir:
@smashercosmo Yeah. in my case, the only way to get that error was to change the language from the dropdown in `<Footer />` and then navigate away to a different route. Changing the language would not only re-render the `<App />`, it would re-mount it. (Sometimes I saw that `<App />` would re-mount sporadically while navigating without changing the language, which made it even harder for me to understand the problem). But I knew that this is where I should be looking. 

After googling for a while, I came to know that `withProgressBar` was trying to use a listener which was registered with an _unmounted_ component to update the state of _newly mounted component_.

According to MDN- 

> # Function.prototype.bind()
> The **bind()** method creates a new function that, when called, has its this keyword set to the provided value, with a given sequence of arguments preceding any provided when the new function is called.
> ## Return Value
> A copy of the given function with the specified this value and initial arguments.

So basically, we'll get a new copy of `router.listenBefore` every time the component gets re-mounted. So I went for this solution.

**BUT**
I can't be 100% sure because of the unpredictable behavior of `LanguageProvider` or whatever that's causing the app to re-mount. What I need is as many people as we can have to test this. If you could, please test and let me know if it works at all. It is working for me.

smashercosmo:
@KarandikarMihir O, I see you changed code, so now there is an actual assignment of the bound function) My question was basically about that)

gihrig:
Still LGTM 💯 

Progress bar works as expected on first visit to Features, also on revisit to Features and Home after language change.

No errors! ✨

Node 7.5.0
npm 4.1.2
yarn 0.20.0
Mac OS 10.11.6
Chrome 56.0.2924.87 (64-bit)

smashercosmo:
By the way, can't we just set ```this.mounted``` flag to true in componentDidMount and unset it in componentWillUnmount and then just use it in updateProgress function:
```javascript
if (this.mounted) {
  this.setState({ progress })
}
```

samit4me:
I've been testing this and it was looking really good, but then after several attempts, I did manage to get the following error (hint: you got to be quick):

`Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the Router component.`

### Testing environment
Windows 10
Chrome 56
node 6.9.1
npm 3.10.8
yarn 0.19.1

### Steps taken
1.  `git clone https://github.com/KarandikarMihir/react-boilerplate.git`
1. `git co bfix/progress-state-error`
1. `yarn`
1. `npm start`
1. Open http://localhost:3000/ in Chrome
1. Open devtools
1. Open Network tab in devtools
1. Select GPRS in the throttling dropdown.
1. On the webpage, click the Features button.
1. Then quickly select "de" from the language dropdown in the footer
1. 💥 the error

I don't have time right now, but will do my best over the next few day to try and diagnose the issue further.

samit4me:
@KarandikarMihir @gihrig Did we want to merge these changes? LGTM, but maybe we should leave the original issue open and look into resolving it is a separate PR???

smashercosmo:
@samit4me Why would you want to merge these changes if they are not fixing the issue?

samit4me:
The change was already approved, so I thought it might be best to ask those involved!

gihrig:
> Why would you want to merge these changes if they are not fixing the issue?

Because It improves the issue to the point where it was not observed with the original testing.

I think it makes sense to merge as is (because it improves the issue) with the understanding there remains an issue that will appear on combination of a specific action and a slow connection. An "edge case"?

So, I would merge, but keep #1505 open for a better fix.

KarandikarMihir:
@samit4me Thanks for digging in! That was spot on! I removed `appWithRouter`, followed your steps and saw the error. I knew [something was fishy](https://github.com/react-boilerplate/react-boilerplate/pull/1553#issuecomment-277600665) with `<LanguageProvider />`.

But that being said, this PR provides some value because it improves the code a little bit- if anything, it's harmless. I've observed that the frequency of `setState` error has decreased after this fix but I don't have a definitive proof. So I say let's merge this in. Basically what @gihrig said. 

Sorry for the delay. I've been busy with work. Let's continue discussion in #1505 and hopefully we'll solve the problem by deductive reasoning.

samit4me:
Awesome, let's get this merged!

Issue #907, opened 2016-08-23
Large image in README


jt3k:
github web-interface is supported only 888 for width

 image in header have filesize 1.2mb and measures 2326 x 676px

I prepare two images:
1. original size 2326x676px but  filesize is 274KB ![banner-metal](https://cloud.githubusercontent.com/assets/1261240/17897541/371e9084-695d-11e6-98f1-b7925bf89e28.jpg)
2. down-scaled size 888x258 53KB ![banner-metal_](https://cloud.githubusercontent.com/assets/1261240/17897533/338c0406-695d-11e6-9476-cde12fa84987.jpg)

let's choose the size and replace the image in the header!

SimonSelg:
Good point. 

> github web-interface is supported only 888 for width

While that's true there are devices with a HighDPI screen (such as the Macbook from which I'm writing this) - from these devices the high res version makes sense.

See this screenshot (you might need to zoom in to see the difference if you're not on a highdpi device):
![image](https://cloud.githubusercontent.com/assets/9430680/18123498/6f3c9d78-6f6e-11e6-9688-38176b7425a8.png)

But compressing the image while keeping the resolution is a good idea 👍

lucidlemon:
![banner-metal](https://cloud.githubusercontent.com/assets/2458762/18167631/f7eb5fc2-7051-11e6-830f-4b0e7b4b601e.jpg)

I'd have a version at exactly @2x for retina screens prepared and optimized right here :) 206kb in size.

mxstbr:
Awesome! Feel free to submit a PR to change this to a smaller image!

KarandikarMihir:
We are getting the brand image from [another](https://github.com/mxstbr/react-boilerplate-brand) repository. If we are to change the image, how would you want to go about this? @mxstbr 

**README.md**
`<img src="https://raw.githubusercontent.com/mxstbr/react-boilerplate-brand/master/assets/banner-metal.jpg" alt="react boilerplate banner" align="center" />`

justingreenberg:
maybe take a look at https://github.com/FormidableLabs/react-progressive-image...

```js
import mainImage from 'main-image.jpg'
import compressedImage from 'compressed-image.jpg'

<ProgressiveImage src={mainImage} placeholder={compressedImage}>
  {(src) => <img src={src} alt="react boilerplate banner" />}
</ProgressiveImage>
```

_side note: the `alt` tag string should be internationalized :)_

[...]

PR #1555, opened 2017-02-04
Resized the banner to 768px width (1.2MB -> 70KB)

tomazy:
That's the maximum width we display it with in the header.

It also makes the app a bit lighter - I'm currently on 3G so I care :)

Processed using Gimp on macOS with 90% of the original quality.


KarandikarMihir:
Hi @tomazy Please follow the guidelines listed under [contributing.md](https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md). Create an issue for this PR, have a go from one of the members of RBP and then you can send a PR. This helps us streamline everything. 

Thanks for contributing! :sparkles:

gihrig:
@tomazy @KarandikarMihir If I'm mistaken please forgive the implication, but I think @tomazy may have been offended, or put off by @KarandikarMihir's response.

@tomazy, should actually be added as a member of the organization, see #1236 Server Side Rendering. This is a huge contribution, and one significant reason (IMO) for RBP being in the #three position (rather than #two) in terms of GH stars. 

GH stars are important in achieving the project's goal of providing the "_best developer experience and a focus on performance and best practices_" to as many as possible. See Andrew Farmer's [React Project Starter List]([url](http://andrewhfarmer.com/starter-project/)).

Oh, and about that image, it's part of a "throw away project" true, but also one that is looked upon as providing a best practice example. Serving a 1.2 Meg image is _not_ a best practice!

So if there's a misunderstanding and upset here, let's patch that up, please and add @tomazy to the Member's list. The project will need him around if it is to support SSR!

tomazy:
@gihrig  I'm not offended at all.  @KarandikarMihir comment actually made to go and look for a similar issue and I've found #907 It didn't know that you guys have a separate repo for the brand stuff and this image should probably be taken from there. I'll open a separate issue for it.

KarandikarMihir:
@tomazy @gihrig I'm sorry, I had no intention of being rude or hurting anybody. We get many uninvited PRs (which is actually an indicator of how people love RBP and want to be a part of it), most of them are from newcomers. My reading is that [CONTRIBUTING.md](https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md) being not very visible to everybody (because it's hidden away in `.github` folder) is one of the reasons why people don't follow the guidelines. So as a maintainer, I let people know that certain rules exist and following them will make it easier to maintain order.. My comment may have been blunt because I'm not a native English speaker and sometimes I lack the charm. I'm still learning that part. I apologize!

I admire and appreciate @tomazy's work on #1236 because I've been watching his work. Had I remembered that it's the same person, I'd not have written the comment because I know that _he's definitely not a newcomer._ **Furthermore, I realize that @tomazy deserves to be a member of RBP more than I do.** Welcome onboard! :smiley:  I'm always excited to work with people of such a great caliber and experience!

About the image- I totally agree with @gihrig. The larger image needs to be replaced by a smaller one.

> It didn't know that you guys have a separate repo for the brand stuff and this image should probably be taken from there.

I think only `README.md` has some assets that are being borrowed from brand repo. All of the other assets are locally hosted. See [Header component](https://github.com/react-boilerplate/react-boilerplate/tree/master/app/components/Header). So we'll just have to place the smaller image there.

gihrig:
@tomazy @KarandikarMihir Thank you both for openly addressing my concerns - I _really_ wanted to be sure!

Glad to hear there's no rift brewing between two of this awesome community's equally awesome members!!! 💖

Issue #1543, opened 2017-01-30
Help in debugging react-boilerplate code


vishalvijay:
Hi

From past fews days I'm trying react boilerplate. Great project guys \m/

I use to debug my JS application using `debugger` keyword or putting breakpoint in browser console. But with this approach I'm having real pain in debugging react boilerplate code.

- What is the recommended way of debugging this app?
- How do I do it?

samit4me:
@Virsaviya 
Thank you for raising an issue, I too get similar problems.

Could you please try changing [Line 55 in internals/webpack/webpack.dev.babel.js](https://github.com/react-boilerplate/react-boilerplate/blob/master/internals/webpack/webpack.dev.babel.js#L55)
from: `devtool: 'cheap-module-eval-source-map',` 
to: `devtool: 'eval-source-map',`. 

Then make sure you stop any currently running npm scripts and then run `npm start`.

I have not yet tried this with webpack v2, but it seems to work with webpack v1. Keen to hear how that goes!

vishalvijay:
Thanks @samit4me 

Debugging is working with that change.

Should I raise a PR for the same?

samit4me:
Yeah that would be great. Before you do tho, please take a look at the [Webpack Devtool Docs](https://webpack.js.org/configuration/devtool/) to get a feel for your options and also the [CONTRIBUTING guide](https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md) for this repo to get familiar with the process. Thank you for you help, it certainly makes life easier when you can debug hey :smile:

gihrig:
Please don't open a PR changing `devtool: 'cheap-module-eval-source-map',` This is used by design to save dev build time see #1087.

However: a PR documenting this would be most welcome. An addition to [`docs/general/faq.md`](https://github.com/react-boilerplate/react-boilerplate/blob/master/docs/general/faq.md) would be most appreciated. 😄 

Here's a list of [build tool options](https://webpack.github.io/docs/configuration.html#devtool) for refernence.

samit4me:
It looks like #1087 was mostly referring to `inline-source-map` and above was referring to `eval-source-map`. According to the [docs](https://webpack.js.org/configuration/devtool/#for-development) it seems `cheap-module-eval-source-map` is the same as `eval-source-map`, but
> It is "cheap" because it doesn't have column mappings, it only maps line numbers.

[...]

PR #1559, opened 2017-02-05
Improve debugging

samit4me:
It seems `cheap-module-eval-source-map` is used due to it's faster initial build times, but unfortunately debugging doesn't really work in Chrome. In my experimenting, `eval-source-map` takes about 600ms longer on initial build, but debugging works. Please see #1543 for further details.


gihrig:
@samit4me 600ms is no issue, but honestly, I have never been able to get sourcemaps to work in Chrome debugger :-/

Could you provide simple steps to demonstrate the value of this PR in Chrome latest?

samit4me:
Hi @gihrig, I'm not sure I understand the question? Does it not work for you?

By using 'eval-source-map' Chrome would pause execution on breakpoints set in Chrome devtools and / or `debugger` statements in source code.

samit4me:
@gihrig Would it be possible for you to test this PR???

If you are unable to place a breakpoint or use a debugger statement there is little point adding this.

Issue #1476, opened 2017-01-17
[Feature] Implement Redux-Actions 


jeremyadavis:
## Issue Type
- [x] Feature

## Description
In a few weeks I'm about to start another project off of this boilerplate and the first thing I usually do after cleaning it is implement [redux-actions](https://github.com/acdlite/redux-actions).

I haven't seen any discussion around the topic and was curious if you'd be interested in a PR for it.

Pros
- Removes action type constants file.
- Cleans up reducer file syntax.
- Standardizes actions to [Flux Standard Actions](https://github.com/acdlite/flux-standard-action).

Cons
- Adds new dependency
- Adds new reducer syntax style that differs from how most tutorials explain reducers.

I've been very happy with it in my apps. I have less boilerplate syntax to maintain and I can confidently get my params from actions without needing to look back at the action creator definitions.

gihrig:
@jeremyadavis Always in favor of reduced boilerplate, I would like to see examples of implementation (I didn't get a clear idea from the r-a project README). Need not be a full PR, unless that's easiest for you. Just a couple of snippets showing how action and reducer code is handled would be a great start. 😀

jeremyadavis:
K, I did a basic 1:1 conversion for the [App files](https://github.com/jeremyadavis/react-boilerplate/commit/556c461064ddd20f49fd256958aca6a47964ef41)

Redux-actions has a few different flavors of syntax that can be implemented. While this might not be the most terse style, I feel that the code is still readable and maintainable whereas some of the others ways were more confusing and/or depend too much on knowing how the redux-actions methods work.

If I did a full conversion I'd probably tweak a few other things like have the saga catch blocks send Error objects so that the [FSA](https://github.com/acdlite/flux-standard-action) can set the error flag.

jsvg:
With [redux-saga 1.4.3](https://github.com/redux-saga/redux-saga/releases/tag/v0.14.3), you can drop the `.toString()` method in your saga watcher to: `yield takeLatest(loadRepos.toString(), getRepos)`. That aside, I really like the reduction in boilerplate this would bring.

jeremyadavis:
@jsvg  Thanks, good to know.

mxstbr:
I'd much rather get rid of things than add even more abstractions at this point. `redux-actions` is great, but I know a lot of apps that are fine without it so I don't see a good reason to include it by default. Adding it is quite easy anyway!

Closing for now, thanks for kicking off this discussion.

[...]

Issue #92, opened 2016-01-15
Server side rendering (SSR)


thomaschaaf:
I would love to see server side rendering in this boilerplate. This would help SEO and initial page load speed could be increased.

This pretty recent article shows how this could be implemented.
https://medium.com/@MoBinni/a-modern-isomorphic-stack-6609c7c9d057#.xz4r6dzd5

mxstbr:
This is something I'm definitely interested in, but can't at the moment pursue because I'm hard at work with v3. I'll close this issue, but it's on my roadmap to check it out, thank you very much for reminding me!

----

If somebody wants to submit a PR I'd love to have a look at what a react-boilerplate + SSR might look like!

> Note: Please submit the PR against the `v3.0.0` branch if you do so!

kachkaev:
Further discussion: https://github.com/mxstbr/react-boilerplate/pull/625 (for those who landed here from search).

PR #1591, opened 2017-02-14
Using `combineReducers` to define shape of the state

tomazy:
This is another PR extracted from #1236 

Please help me to review it.


tomazy:
@samit4me good point! I've updated the code.

gihrig:
A generated route shows the Not Found Page in production mode, OK in dev mode.

I tested according to the following script. 

```
-- Test script --
yarn && npm start
Test Dev, Dev tunnel, Features, lang = de, NotFound, Production, off-line
npm run clean
npm run generate component TestComp /w defaults
npm run generate container TestPage /w defaults
npm run generate route to TestPage - /test
use TestComp on TestPage -> bypass all tests in TestComp and TestPage (set true = true)
npm start > localhost:3000/test
npm test (expect test failure due to incomplete test coverage)
npm run build

**Problem:** npm run start:prod  > localhost:3000/test <--- _shows Not Found Page_

Kill server  > localhost:3000 _OK_ ~/test _ site can’t be reached_ - see docs > server config
```

gihrig:
> This looks really good, I find this to be much cleaner and easier to read.

Yes but, It's a break from the "standard" redux reducer pattern most people are familiar with.

To expect that a new immutable state object will be returned when the reducer returns a boolean, just make no "sense" (there must be some "magic" happening somewhere).

I'm not arguing against this PR, but another, similar proposal was made to use [redux-actions](https://github.com/acdlite/redux-actions) to greatly simplify the creation of actions in #1476 and that idea was [shot down](https://github.com/react-boilerplate/react-boilerplate/issues/1476#issuecomment-274264189) for this reason.

If we are to go forward with the change, I would suggest:

1. The change needs to be documented. In code comments for sure, existing docs need to be reviewed as well.
2. The rationale for making the change needs to be spelled out, perhaps just in the opening comment of this PR.
3. If redux-immutable is a good idea, redux-actions is probably a good idea (for actions, not reduceres) for the same reasons.
4. While not a breaking change, IMO, this is sufficiently different enough that that it's inclusion should trigger a minor version bump to 3.5.0 and 3.4.1 should be released first.

Overall I like the change, boilerplate reduction is (usually) a good thing. 😄

tomazy:
@gihrig thanks for your input. I've updated the templates.

> A generated route shows the Not Found Page in production mode, OK in dev mode.

That's weird - the changes in this PR don't introduce anything that might cause that. The reducers should work **exactly** the same as previously. I've just tested your scenario and I can't reproduce it.

>> This looks really good, I find this to be much cleaner and easier to read.
>
> Yes but, It's a break from the "standard" redux reducer pattern most people are familiar with.

I'm not sure about that. Combining reducers is [mentioned](http://redux.js.org/docs/api/combineReducers.html) in the redux docs and it's a quite standard thing IMO. 
 
> If we are to go forward with the change, I would suggest: (...)

I'm afraid I will not find time to address your suggestions. I've been quite busy with other stuff and it becomes more difficult for to keep the #92 branch alive (that's the main reason for this PR). If you guys are not happy with this PR then we can close it and I'd remove these changes from #92 (they are not actually critical for it) to keep it leaner.

gihrig:
#92 Yikes, I had no idea this project (SSR) went back that far. It's been a long haul!

> If you guys are not happy with this PR then we can close it and I'd remove these changes from #92

@tomazy It was absolutely _not_ my intention to discourage. But I am concerned about the approachability of the project. It's got plenty of difficult stuff as is.

> Combining reducers is mentioned in the redux docs and it's a quite standard thing IMO.

I totally agree about combining reducers, no issue there. It's the change from returning an obvious state object to returning a primitive value. 

e.g. changing from:
```javascript
      // Delete prefixed '@' from the github username
      return state
        .set('username', action.name.replace(/@/gi, ''));
```
to:
```javascript
      // Delete prefixed '@' from the github username
      return action.name.replace(/@/gi, '');
```

It looks to me like a simple string is being returned rather than an immutable state object. Obviously I don't grok what's really happening here, that's all.

>I'm afraid I will not find time to address your suggestions. I've been quite busy with other stuff and it becomes more difficult for to keep the #92 branch alive.

I totally get it about difficulty finding time! 😬  

What do you, and others think about creating an SSR branch in this project? 

Kept up as a super-set of dev, that would allow these related PRs to have place to live as an evolving sub-project. This PR for example would be right at home in a branch dedicated to SSR and the issues I raised could be addresses separately by the community. I'm hopeful that would invite wider contribution to the SSR feature and still facilitate release as a part, or all?, of RBP 4.0.

With huge props to @tomazy for all the work put in on this long journey, I'd like to hear input from others. I think SSR is too big a feature to let slip away.

tomazy:
@gihrig 

> @tomazy It was absolutely not my intention to discourage.

I'm sure about that.

> But I am concerned about the approachability of the project. It's got plenty of difficult stuff as is.

I understand it's hard to find balance between the approachability vs. best practices. I know it's all subjective but I'd put more emphasis on the latter.

> What do you, and others think about creating an SSR branch in this project?

Sounds good to me

gihrig:
Cloned a new copy of your branch tested as before, all pass ✨ 

So this PR has my approval, but I'll let others have a say as to whether to merge into dev or create an SSR branch.

blling:
@tomazy ping! 
Conflicts need resolve:)

tomazy:
@blling here you go

gretzky:
Nice work

Issue #1589, opened 2017-02-13
styled-components style get overriden


eric-burel:
# React Boilerplate

## Issue Type

- [X] Bug (https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md#bug-reports)

## Description

This bug is similar to #620 but with `styled-components`. I am loading Bootstrap from the `app.js` file : 
```
// Import bootstrap style
import 'bootstrap/dist/css/bootstrap.min.css'

// Import CSS reset and Global Styles
import './global-styles';

// Font awesome font icons
import 'font-awesome/css/font-awesome.min.css'
```
 Whatever the order, the `<style>` tag for Bootstrap is always loaded *after* the `<style>` tag for the styled components. As `global-styles` are concatenated in the same tag, it is loaded before Bootstrap as well.

Thus I can't manage to override Bootstrap (specifities are of course equal) since it is always inserted after.

# Versions

- React-Boilerplate (see `package.json`): 3.4.0
- Node/NPM: 6.3

samit4me:
I've had similar issues @eric-burel and to summarize, the imported styles are injected into the `<head>` via webpack + plugins and the ordering cannot be guaranteed, so specificity is going to be your friend :bear: 

Please take a look at [this issue](https://github.com/styled-components/styled-components/issues/85) and let me know if it resolves the issue?
https://github.com/styled-components/styled-components/issues/85

eric-burel:
Thanks for the link. I think they closed the issue a bit quickly in the `styled-components` repo, increasing specificity is a quick (somehow dirty) fix, not a solution. 
The issue is that I deliver code to a client: will the next developper know that he must add `&& {}` to every styled components?

samit4me:
Another option would be to import your bootstrap styles in the index.html `<head>` specifically:

`<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/latest/css/bootstrap.min.css">`

This should help as the styles imported via JS will be placed after the above. Does that work?

GGAlanSmithee:
Might be related to https://github.com/webpack/webpack/issues/215? Seems impossible to make webpack include css in a specific order. Had the same problem in a project @ work, and ended up "fixing" it by inlcuding reset / bootstrap etc styles directly in the `<head>` as @samit4me suggested.

eric-burel:
@samit4me Yes it does :) , that's quite cleaner than the specificity override.

[...]

PR #1592, opened 2017-02-14
Update FAQ with Styles getting overridden

samit4me:
An attempt to shine a light on potential style conflicts between classic CSS and `styled-components`. See #1589.


eric-burel:
Hi, I think this is clear enough, however not everybody know the tricks to improve specificity in CSS. For example I did not know about the `&&` (my reflex would be to use `!important`).
Maybe adding an example could be good:

```javascript
const MyBlueDiv = styled.div`
    && {
        background-color: blue;
    }
`
```
```html
<!-- I will still be blue, whatever the class specified -->
<MyBlueDiv className="red" />
```

samit4me:
Thanks for the feedback @eric-burel it's appreciated 👍 I've taken your suggestions and added examples, let me know what you think.

eric-burel:
That's great! 

To make it perfect, I'd simply put the "1) Increase the specificity of the CSS you want to win" and "2) Import the CSS in the <head> of your index.html manually " in bold.

Also you could add a sentence before each of first solution example, like:
"I want the imported css to win:"
"Or, I want the styled-component css to win:"
because it's a bit unclear to have those two contradictory examples one after the other

samit4me:
Thanks again @eric-burel both good points, updated as suggested.

gihrig:
@samit4me It's looking really good ✨ 

The only thing I would add, is to include language specifiers in your code fencing for syntax highlighting.

e.g.
\```css

or 

\```javascript

See other examples higher up in the doc.

samit4me:
💯 @gihrig, thank you

Issue #1595, opened 2017-02-16
Module parse failed 


kajm:
Hi guys!

When I add a styles.css file in one of my components I get this "module parse failed" - error and its the message redirect me to the new file that I have made. I have tried googled it a couple of times, but I just can't figure it out what the problem is. 

I have currently this versions installed:
React-boilerplate: 3.4.0
NPM: 4.1.2
Node: 7.5.0
Google Chrome

and here is the error:
<img width="1180" alt="skarmavbild 2017-02-16 kl 20 03 44" src="https://cloud.githubusercontent.com/assets/9610681/23036474/40ca1d2c-f483-11e6-955d-f9bd4c5edeb1.png">

What could be wrong?

best regards,
Karl

gggdomi:
Hey @kajm,
I had the exact same problem, and I think I finally figured out the issue. It lies in `webpack.base.babel.js` when configuring CSS loaders:
```javascript
...
{
      // Do not transform vendor's CSS with CSS-modules
      // The point is that they remain in global scope.
      // Since we require these CSS files in our JS or CSS files,
      // they will be a part of our compilation either way.
      // So, no need for ExtractTextPlugin here.
      test: /\.css$/,
      include: /node_modules/,
      loaders: ['style-loader', 'css-loader'],
}
...
```

Due to  `include: /node_modules/`, only the .css files in `node_modules` will be processed by the loaders.
This new behaviour has been introduced by #1100 when switching to styled-components.
To import my local CSS file, I added the old code just before the previous rule:

```javascript
...

    }, {
      // Transform our own .css files with PostCSS and CSS-modules
      test: /\.css$/,
      exclude: /node_modules/,
      loaders: ['style-loader', 'css-loader'],
      // BTW, if you want to enable CSS modules you can use this instead:
      // loaders: 'style-loader!css-loader?modules=true&localIdentName=[name]__[local]___[hash:base64:5]',
    }, {
      // Do not transform vendor's CSS with CSS-modules
      // The point is that they remain in global scope.
...
```

Not sure if it's a good idea (I'm new to React and this boilerplate), but at least the error is gone :)

samit4me:
This is definitely an issue. Not sure how this was missed but this boilerplate has been using webpack 2 for quite a long time and it looks like the loaders need updating as @kajm mentioned. See https://webpack.js.org/guides/migrating/#module-loaders-is-now-module-rules

kajm:
[...]

PR #1599, opened 2017-02-18
Fix imported styles

samit4me:
Fixes imported styles (e.g. `import './styles.css'`) as identified in #1595.

### Reason for change
The source of the problem was `include: /node_modules/`. Essentially this would only allow CSS to be imported from the `node_modules` directory.

This rule was added [in this commit](https://github.com/react-boilerplate/react-boilerplate/commit/9d620d01e52d0e4983457ee7579f700ae55dbc17)  so that CSS in 3rd party code didn't get piped through CSS Modules, ONLY local CSS would go through CSS Modules.

When `styled-components` was added the CSS Modules loader was removed from webpack, which broke any imported CSS that did NOT live in `node_modules`. 

---

This PR also updates the webpack loader syntax based on the webpack 2 migration guide https://webpack.js.org/guides/migrating/#module-loaders-is-now-module-rules.

It would be fantastic if you could help me review and test this change :heart_eyes:


gggdomi:
Maybe we should keep two rules to differentiate third-party and local files, like it was before ?

```javascript
{
  // Transform our own .css files with PostCSS and CSS-modules
  test: /\.css$/,
  exclude: /node_modules/,
  use: ['style-loader', 'css-loader'],
}, {
  // Do not transform vendor's CSS with CSS-modules
  // The point is that they remain in global scope.
  // Since we require these CSS files in our JS or CSS files,
  // they will be a part of our compilation either way.
  // So, no need for ExtractTextPlugin here.
  test: /\.css$/,
  include: /node_modules/,
  use: ['style-loader', 'css-loader'],
}
```
instead of

```javascript
{
  test: /\.css$/,
  use: ['style-loader', 'css-loader'],
}
```

This is more verbose, but it would probably prevent mistakes from dev trying to enable CSS modules.

samit4me:
Thank you for the feedback @gggdomi 👍 

I agree, having two loaders will certainly make it more obvious that the local and 3rd party CSS are preprocessed separately, although in the same way.

If we do add two loaders, the benefits will only be seen if/when a user wants to enable CSS Modules. I can certainly understand this and this may be somewhat narrow-minded, but I struggle to find a reason to use CSS Modules when you have `styled-components` ([Comparing different ways to style components](https://github.com/styled-components/comparison)).

I'm happy to go either way, let's get this problem fixed.

gggdomi:
Thank you for all you've done @samit4me !

Then maybe just keep one loader with an explanatory comment for the users still interested in CSS modules ? It may avoid some issue reporting on GitHub, and a lot of headaches for them (I spent hours trying to figure it out yesterday. And finally switched to `styled-components` today 😅)

samit4me:
Yes, that is a good point @gggdomi. I've just added a small comment to help clarify the potential issues associated with modifying the CSS loader. Hopefully, this will save time for anyone else doing similar things in the future.

KarandikarMihir:
@samit4me Switching to new webpack syntax was in order :smile: Great that you've addressed it here.

I'd like to express one concern regarding the comment that you've left. Can we add that extra rule (the one that processes own css) and just comment it out? So that devs can just enable it by un-commenting it. People using sass/less can append respective loader and they'll be ready to roll. Does that make sense? Because our webpack config is already overwhelming, devs would be wary of making changes on their own.

One more thing, in the comment you've mentioned `then it's a good idea to have two CSS loaders`. Did you mean two separate _rules_? It can be confusing at times..

I meant to say something like this. It would be less verbose but easier to manage.

```js
/* Uncomment this to enable processing your own css.
{
  test: /\.css$/,
  exclude: /node_modules/,
  use: ['style-loader', 'css-loader'],
},
*/
{
  test: /\.css$/,
  use: ['style-loader', 'css-loader'],
}

```

> It may avoid some issue reporting on GitHub, and a lot of headaches for them 

@gggdomi I agree one hundred percent.

samit4me:
That is definitely possible @KarandikarMihir and thank you for the suggestion. I don't think uncommenting the top rule will work as you would also need to add `include: /node_modules/` to the bottom rule. What if we simply add two rules as suggested by @gggdomi? Something like the following:

```javascript
{
  // Preprocess our own .css files
  // This is the place to add your own loaders (e.g. sass/less etc.)
  // for a list of loaders, see https://webpack.js.org/loaders/#styling
  test: /\.css$/,
  exclude: /node_modules/,
  use: ['style-loader', 'css-loader'],
},
{
  // Preprocess 3rd party .css files located in node_modules
  test: /\.css$/,
  include: /node_modules/,
  use: ['style-loader', 'css-loader'],
}
```

KarandikarMihir:
@samit4me Ah, you're right. It won't be that straightforward. 

> What if we simply add two rules as suggested by @gggdomi? 

I feel like we're duplicating code here, but there's no other way. I'd personally prefer styles-components over css files hands down, but we can't force everybody :cry:  I vote for adding two separate rules :+1: Thanks for reconsidering it! Let's see what others have to say..

samit4me:
Sounds good, thank you @KarandikarMihir and @gggdomi for the suggestion. I will push the changes when I get home tonight!

KarandikarMihir:
Awesome! :smile:

KarandikarMihir:
Looking great! I tested it real quick

**Using third party css framework.**
1. `npm install semantic-ui-css`
2. import `semantic-ui-css/semantic.min.css` directly
2. Use Semantic UI css classes in one of the pages.

_Result: Worked as expected_

**Using our own css**
1. Create `styles.css`
2. import `styles.css` directly
2. Use classes from `styles.css`.

_Result: Worked as expected_

**Using our own less**
1. Install and add `less-loader` to the webpack config.
2. Modify filename test in the rule `test: /\.css$/` -> `test: /\.less$/`
3. Create `styles.less`
4. import `styles.less` directly
5. Use classes from `styles.less` in the markup.

_Result: Worked as expected_

Awesome! :100:

KarandikarMihir:
@gihrig Can you please take a look at this before we merge this?

gihrig:
Well, failed for me 😢 

Here's what I did:
`FeaturesPage/styles.css`
```css
.redBack {
  background-color: red;
}
```
`FeaturesPage/index.js`
```javascript
import Styles from './styles.css';
// ...
<H1 className={Styles.redBack}>
```

`http://localhost:3000/features`
```html
<style type="text/css">.redBack {
  background-color: red;
}
</style>
...
<h1 class="islPJm"><span>Features</span></h1>
...
.islPJm {
    font-size: 2em;
    margin-bottom: 0.25em;
}
```

Did I do something wrong...

KarandikarMihir:
@gihrig i think you're using CSS modules. Can you try appending `modules=true` as query param to the CSS loader?

KarandikarMihir:
For reference- https://github.com/webpack-contrib/css-loader/blob/master/README.md#options

gihrig:
> I think you're using CSS modules. Can you try appending modules=true as query param to the CSS loader?

OK that worked. ✨ 

>Using our own css
>
>Create styles.css
>import styles.css directly
>Use classes from styles.css.
>Result: Worked as expected

How did your test differ from mine?

Regardless, there needs to be some docs on this IMO.

KarandikarMihir:
It's the way I imported CSS. `import './styles.css';` instead of `import styles from './styles.css';` 

If we just leave `modules=true`, both ways will work out of the box. I agree, this needs to be documented.

gihrig:
I see...

It would appear that existing docs have an error.

When I followed the documented style:
```javascript
// ContrivedExample.js
import MyStyledComponent from './MyStyledComponent';
import './styles.css';

const ContrivedExample = (props) => (
  <MyStyledComponent className={styles.alert}>
```

I get an error `styles is not defined`

The syntax required to eliminate the error for this impart style is
```javascript
import './styles.css';
// ...
  <MyStyledComponent className="styles.alert">
```

But when I do that, my style is not applied, even with '&&' or '!important'.

KarandikarMihir:
Umm.. can you try just `className="alert"`?

samit4me:
Nice work @gihrig and @KarandikarMihir :tada: Really appreciate the help testing this.

CSS vs CSS Modules definitely looks to be the issue.

**RE: docs error**
:100: correct @gihrig, that was my bad, I'll fix that and push in with this PR if you are both okay with that?

KarandikarMihir:
@samit4me Yup, let's get that right in this PR only! 

Just to be clear 

**Using CSS modules (with `modules=true`)**
```javascript
import styles from './styles.css';

...
render() {
  return (
    <div className={styles.container}>
    </div>
  )
}
...
```

**Plain old CSS classes**
```javascript
import './styles.css';

...
render() {
  return (
    <div className="container">
    </div>
  )
}
...
```

gihrig:
>Umm.. can you try just className="alert"?

Closer, still no styles

```html
<h1 class="redBack islPJm"><span>Features</span></h1>
```

The `redBack` rule is written to a `<style>` tag, but the name is 'munged' so there is no match:
```html
<style type="text/css">._1cQjTiLjaI7eHyqRdskFw {
  background-color: red;
}
</style>
```

gihrig:
>RE: docs error
>💯 correct @gihrig, that was my bad, I'll fix that and push in with this PR if you are both okay with that?

Yes Please! 😄

KarandikarMihir:
@gihrig 

> `<h1 class="redBack islPJm"><span>Features</span></h1>`

It's `className` and not `class` :sweat_smile:

gihrig:
@KarandikarMihir 
>`<h1 class="redBack islPJm"><span>Features</span></h1>`
>It's className and not class 😅

This from the html, where it _is_ class, not className. In the JSX, I have:
```javascript
<H1 className="redBack">
```

KarandikarMihir:
Oh! Sorry! Wonder, what am I missing. Give me some time. I'll re-test.

KarandikarMihir:
@gihrig @samit4me  The part where I was wrong is you cannot use css modules and normal css classes at the same time. We need to decide how we want to apply css. Using css modules > css classes because it obfuscates the classnames, and avoiding conflict.. I'd say, let's leave `modules=true`.

If we want to enforce css modules, there's a new webpack syntax for that.

```javascript
{
  // Preprocess our own .css files
  // This is the place to add your own loaders (e.g. sass/less etc.)
  // for a list of loaders, see https://webpack.js.org/loaders/#styling
  test: /\.css$/,
  exclude: /node_modules/,
  use: [
      'style-loader',
      {
          loader: 'css-loader',
          query: {
              modules: true
          }
      }
  ],
}
```

**With CSS modules-**
screenshot from 2017-02-21 11 01 30

screenshot from 2017-02-21 11 01 41

screenshot from 2017-02-21 11 01 53

**With CSS classes-**
screenshot from 2017-02-21 11 00 56

screenshot from 2017-02-21 11 00 59

samit4me:
@gihrig 
> `<H1 className="redBack">`

 If the class name is "munged" it sounds like CSS Modules are in use, do you still have the following in your webpack config?
```
options: {
  modules: true
}
```

---

**RE: supporting CSS vs CSS Modules**

I vote that we support standard CSS but NOT CSS Modules. CSS Module support was dropped in favor of `styled-components` back when `styled-components` was added.

I believe the main reason for supporting normal CSS is so people can add styling in the "classic way", therefore provides developers with a smoother pathway towards modern CSS tooling. Having the `css-loader` rule present also makes it easier to enable sass / less, which I assume is common practice for migrating existing JS apps.

gihrig:
> you cannot use css modules and normal css classes at the same time. We need to decide how we want to apply css. Using css modules > css classes because it obfuscates the classnames and avoids conflicts.. I'd say, let's leave modules=true.

I agree. The syntax feels more natural to me and since there is no change to the `.css` file I think cssModules is an 'opinion' we can afford.

On the contrary, I expect some would prefer not to have class names obfuscated in the html. That's likely a more significant trade-off than internal syntax.

It would be nice if we could hear from some avid 'Bootstrap, Foundation', etc. users on this.

gihrig:
@samit4me 
>I believe the main reason for supporting normal CSS is so people can add styling in the "classic way", therefore provides developers with a smoother pathway towards modern CSS tooling.

The more I think of it the more I expect this would be a big issue for those who use a css library/framework like bootstrap. Where CSS debugging must be done in the browser. Not being able to understand the class names provided by the library will make debugging a _lot_ harder.

Based on that, I could change my vote to normal css.

KarandikarMihir:
We're extensively using a custom framework built on top of bootstrap. One of the most annoying things we've had to face is classname collision. When developers feel less creative, they start using generic names for classes which leads to horrible collisions across pages.

So, for "handwritten" css, turning on `modules` would be a good practice. We avoid collision. But the trade-off is that debugging would be a lot harder. And for installed frameworks, not enforcing anything makes debugging easier.

But now I lean more towards standard CSS because it's developers responsibility to write meaningful and unique classnames. Let developers decide what they want. We'll just enable handwritten css without modules. Anyways we're not using css files in the example app.

_I change my vote for standard css._

That's why I love `styled-components`!

gihrig:
So we agree on standard css. 😄 

Since it's a trivial change, can we document how to use css modules in the FAQ and in the code comments?

KarandikarMihir:
> can we document how to use css modules in the FAQ and in the code comments?

Yas! @samit4me Are you including this in the same PR?

samit4me:
Thanks again @KarandikarMihir @gihrig 😄 Standard CSS sounds great. I've pushed fixes for the FAQ and SASS docs.

I agree about pushing further docs to explain the usage of CSS Modules etc. I don't have time tonight, but I am thinking about renaming the **"Using Sass"** doc linked on the [CSS README](https://github.com/react-boilerplate/react-boilerplate/tree/dev/docs/css) to something like **"What about CSS Modules, SASS or LESS?"**. Then re-purposing the [existing docs](https://github.com/react-boilerplate/react-boilerplate/blob/dev/docs/css/sass.md) to say we support `styled-components` and why, but also provide example of how to setup webpack for each (CSS Modules, SASS and LESS) and also how to use each in your application.

What do you guys think? I should have time tomorrow night to draft up something.

KarandikarMihir:
@samit4me 
> also provide example of how to setup webpack for each (CSS Modules, SASS and LESS)

Sounds great!! Can't wait to see the first draft :+1:

gihrig:
Agreed 👍  I think will be much better organization of the data.

samit4me:
@gihrig @KarandikarMihir I've just pushed up my first draft of the updated Styling (CSS) docs. If you have any spare time, it would be greatly appreciated if you could have a read and try out the examples (I actually haven't tested them yet). Hopefully, this should be enough to get the ball rolling.

You can view it via https://github.com/samit4me/react-boilerplate/blob/b38658d6126cee4ee8fef8526efc8e1f2105c39f/docs/css/README.md

docs/css/README.md
```diff
@@ -1,14 +1,201 @@
...
+  background-color: red;
+}
+```
+
+**`Button.js`**
+```js
+import React from 'react';
+import './Button.css'; // Tell Webpack that Button.js uses these styles
+
+class Button extends Component {
```
KarandikarMihir: It's not important, but it should be `React.Component`. Or we can just import `Component` explicitly up at the top...

docs/css/README.md
```diff
@@ -1,14 +1,201 @@
...
+
+## CSS Support
+
+We support and recommend the use of [`styled-components`](#components).
+We also support the use of CSS [stylesheets](#stylesheet).
+
+There are many ways to style web applications, unfortunately, we cannot support them all.
+However, you can integrate the following styling methods:
+- [CSS Modules](#css-modules)
+- [Sass](#sass)
```
KarandikarMihir: Can we add `less` too? After reading this, it sounded like we don't support less. But we do! It's very similar to how we set up sass..

docs/css/README.md
```diff
@@ -1,14 +1,201 @@
...
++   },
++ ]
+}
+```
+
+### Usage
+
+The syntax is very similar to using a [Stylesheet](#stylesheet)
+and this often catches people out.
+The key difference in CSS Modules is that you import to a variable.
```
KarandikarMihir: should read as: ...import (_it_ or _styles_) to a variable.

docs/css/README.md
```diff
@@ -1,14 +1,201 @@
...
+  background-color: red;
+}
+```
+
+**`Button.js`**
+```js
+import React from 'react';
+import styles from './Button.css'; // different import compared to stylesheets
+
+class Button extends Component {
```
KarandikarMihir: Same here. `React.Component`

docs/css/README.md
```diff
@@ -1,14 +1,201 @@
...
+```
+npm i -D sass-loader node-sass
+```
+
+Modify [`webpack.base.babel.js`](../../internals/webpack/webpack.base.babel.js#L30)
+to look like the following:
+
+```diff
+{
+  test: /\.css$/,
```
KarandikarMihir: Should be `test: /\.scss$/`

docs/css/README.md
```diff
@@ -1,14 +1,201 @@
...
+  background-color: $error-color;
+}
+```
+
+**`Button.js`**
+```js
+import React from 'react';
+import './Button.scss';
+
+class Button extends Component {
```
KarandikarMihir: Same here. `React.Component`

docs/css/README.md
gihrig: Form my perspective "import to a variable" is fine, otherwise "import _styles_ to a variable". 😄

docs/css/README.md
gihrig: 👍

docs/css/README.md
gihrig: 👍

docs/css/README.md
gihrig: > Should be test: /\.scss$/

This is not the SASS section...

docs/css/remove.md
```diff
@@ -1,4 +1,4 @@
 ## Removing `sanitize.css`
 
-Delete [lines 31 and 32 in `app.js`](../../app/app.js#L31-L32) and remove it
+Delete [line 19 in `app.js`](../../app/app.js#L19) and remove it
 from the `dependencies` in [`package.json`](../../package.json)!
```
gihrig: Would be easier to maintain going forward if we gave example of code to be removed and let the dev find it themselves. As demonstrated here, line numbers change...

gihrig:
@samit4me Like @KarandikarMihir I have given it a quick skim, with a few comments. Will have to wait on testing example code... 😢

docs/css/README.md
samit4me: I can certainly add it in, it might be a learning experience as I've never used it before.

docs/css/README.md
samit4me: Most definitely, good pickup. I will go with `React.Component` as it is used heavily in the react docs.

docs/css/README.md
samit4me: 👍

docs/css/README.md
samit4me: 👍

docs/css/README.md
samit4me: Thanks guys, copy past fail at work right there 😉

docs/css/README.md
samit4me: 👍

docs/css/remove.md
samit4me: Yes, this is nightmare updating specific line numbers 👍

samit4me:
@gihrig @KarandikarMihir You guys are awesome, thank you for reviewing the changes so quickly. The issues raised have been addressed and I have also stepped through each guide to test if they actually work 😟. There were a few issues and they have now been resolved, so it's ready for round 2, thanks again 👍

docs/css/README.md
```diff
@@ -32,11 +33,12 @@ We support and recommend the use of [`styled-components`](#components).
...
 
 There are many ways to style web applications, unfortunately, we cannot support them all.
-However, you can integrate the following styling methods:
+However, you can integrate the following by the guides below:
 - [CSS Modules](#css-modules)
 - [Sass](#sass)
+- [LESS](#less)
 
-## `styled-components`
+## [`styled-components`](https://github.com/styled-components/styled-components)
```
jaredmorgs: When you apply backticks in headers, you tend to lose consistency in how the TOC is presented. 

screen shot 2017-02-23 at 7 53 50 am

I'd suggest leaving the Headings without any inline markup. Some readers might not see the Heading is actually a link. Promoting the link to a "See <link> for more information" will assist users of the doc.

If you leave the link naked you'll be able to very easily detect and fix URL changes in the rendered file, which may help other contributors fix link errors quickly.

docs/css/README.md
```diff
@@ -77,7 +81,7 @@ function Header(props) {
 For further examples see the
```
jaredmorgs: Good example of a "See also" call to action for readers. 👍

docs/css/README.md
```diff
@@ -77,7 +81,7 @@ function Header(props) {
...
 [official documentation](https://github.com/styled-components/styled-components)
 
-## Stylesheet
+## [Stylesheet](https://github.com/webpack-contrib/css-loader)
```
jaredmorgs: Promote link to a "See Also" as previously suggested.

docs/css/README.md
```diff
@@ -156,7 +160,7 @@ class Button extends Component {
 **IMPORTANT: if you enable this rule, [stylesheets](#stylesheet) will no longer work,
 it's one or the other unless you include or exclude specific directories.**
 
-## [Sass](sass.md)
+## [Sass](https://github.com/webpack-contrib/sass-loader)
```
jaredmorgs: Promote link to a "See Also" as previously suggested.

docs/css/README.md
```diff
@@ -193,7 +197,60 @@ to look like the following:
...
 
-class Button extends Component {
+class Button extends React.Component {
+  render() {
+    return <button className="danger">Click me</button>;
+  }
+}
+```
+
+## [LESS](https://github.com/webpack-contrib/less-loader)
```
jaredmorgs: Promote link to a "See Also" as previously suggested.

docs/css/remove.md
```diff
@@ -1,4 +1,34 @@
 ## Removing `sanitize.css`
 
-Delete [line 19 in `app.js`](../../app/app.js#L19) and remove it
-from the `dependencies` in [`package.json`](../../package.json)!
+To remove `sanitize.css` you will need to remove it from both:
+- [`app.js`](../../app/app.js)
```
jaredmorgs: Update to the Canonical URL, to avoid future link rot.

docs/css/remove.md
```diff
@@ -1,4 +1,34 @@
...
+```diff
+import FontFaceObserver from 'fontfaceobserver';
+import { useScroll } from 'react-router-scroll';
+-import 'sanitize.css/sanitize.css';
+
+// Import root app
+import App from 'containers/App';
+```
+
+- [`package.json`](../../package.json)!
```
jaredmorgs: Consider Canonical URL use here.

samit4me: The term "Canonical" still isn't in my vocabulary, but does this mean following the instructions on [this page](https://help.github.com/articles/getting-permanent-links-to-files/) to get a permanent link???

This does sound useful in examples that link to a specific line, but does this mean the link will need updating with each release that changes the linked file?

docs/css/README.md
samit4me: This sounds great, will get it changed, thanks 👍

samit4me:
Awesome, thanks a bunch @jaredmorgs. I've just pushed some updates:
- Removed backticks on headings
- Added "See Also" sections using a quote (limited options)

I'm not 100% sure on Canonical URLs as mentioned in a comment response above.

KarandikarMihir:
@samit4me Awesome. Docs are looking really great :smile:  Are we good to go? Or do you want to use canonical URLs?

samit4me:
I think we are good go @KarandikarMihir 👍

KarandikarMihir:
Let's merge this in! Thank you @samit4me for all your work :100: :smiley: @gihrig and @jaredmorgs for your valuable inputs! :)

PR #1609, opened 2017-02-23
refactor(react-router-scroll): Import only useScroll

sorin-davidoi:
Follow [the guideline](https://github.com/taion/react-router-scroll#minimizing-bundle-size) and only import what we need. Reduces main bundle size by 2 kB (709 kB -> 707 kB).


samit4me:
LGTM, nice work :sparkles:

Issue #1250, opened 2016-11-18
Accessing environment variables in client-side code and creating builds for multiple environments


karloespiritu:
I'm having issues accessing environment variables (i.e API_URL) after I deploy and build my app to multiple environments (staging, production, etc).

I'm using the boilerplate as a client application that consumes data from a REST API. My application is deployed in Heroku and I use Heroku Pipelines for my deployment and build  process. 

What's the best way to access environment variables after I build the boilerplate (`npm run build`)? I deploy the boilerplate to multiple environments (staging, production, etc.) with different environment variables, and I can't seem to find a preferred of doing this.

Thank you very much.

samit4me:
I'm not sure I fully understand your requirements, but if you are wanting a GLOBAL variable available in all environments, maybe the webpack [DefinePlugin](http://webpack.github.io/docs/list-of-plugins.html#defineplugin) could help!

If you add your own variable to the the [webpack.base.babel.js](https://github.com/mxstbr/react-boilerplate/blob/master/internals/webpack/webpack.base.babel.js#L58) it will be available on all platforms.

In the past I've added the API URL to `package.json` and then accessed it in the webpack config, for example:

```
var package = require('<pathto>/package.json');
...
new webpack.DefinePlugin({
  'process.env': {
    NODE_ENV: JSON.stringify(process.env.NODE_ENV),
  },
  API_URL: JSON.stringify(package.apiUrl)
}),
...
```

If you want to keep the linter happy you can add a globals sections to the eslint config, which is in `package.json` for this repo:
```
"eslintConfig": {
  ...
  "globals": {
    "API_URL": false
  }
  ...
},
...
```

Please let us know if that helps or if I've completely missed your question :)

karloespiritu:
Thanks, but it's not exactly what I'm looking for. I've actually tried that approach, but it requires defining the API_URL as a fixed string constant, and not as an environment variable. Let me illustrate what I'm trying to achieve:

1. In my development environment, when I run `npm run build`, the build script should read my `.env` file contains the ff key-value pair:

  ```
  # .env
  API_URL=http://localhost:8000
  ```

2. When I deploy it to my `staging` server using `git push staging`, the `npm run build` script should use the the environment variable defined in my *staging* server which contains the ff key-value pair:

  ```
  # .env
  API_URL=http://myapi-staging.herokuapp.com
  ```

[...]

PR #1616, opened 2017-02-27
add environment variables to webpack

xavierfuentes:
This PR will enable the use of .env files for development.

### DIFF
--- a/internals/webpack/webpack.base.babel.js
+++ b/internals/webpack/webpack.base.babel.js
@@ -1,10 +1,10 @@
 /**
  * COMMON WEBPACK CONFIGURATION
  */
+require('dotenv').config();
 
 const path = require('path');
 const webpack = require('webpack');
-const dotenv = require('dotenv').config(); // eslint-disable-line no-unused-vars
 
 const rawEnvVars = Object
   .keys(process.env)


gihrig:
I agree with @samit4me, documenting both approaches (.env files and adding env vars to `package.json`) would be preferable.

GGAlanSmithee:
@xavifuefer 
> for development.

@samit4me 
> I'm not sure .env files are suitable for all environments

Out of curiosity, would you mind sharing under what circumstances it would not be appropriate to supply run-time variables via a `.env` file? Especially for sensetive data such as db user info, session secrets etc?

samit4me:
@GGAlanSmithee I don't know for sure, but it was raised in issue #1250 
> The `.env` seems to be for local dev and not production https://devcenter.heroku.com/articles/heroku-local#set-up-your-local-environment-variables.

If anyone knows a method that will work in all environments, let's add it to the docs. I'm only aware of `package.json` + webpack + definePlugin option, but would love to know any other options.

xavifuefer:
The problem with the `package.json` is that it gets versioned and you don't want your prod vars in a repo. Credentials should only be accessible on the machines that need access to them. As an alternative we could commit the .env file with development-only settings so other developers can get started on the boilerplate without compromising credentials for other environments.

> In a twelve-factor app, env vars are granular controls, each fully orthogonal to other env vars. They are never grouped together as “environments”, but instead are independently managed for each deploy. This is a model that scales up smoothly as the app naturally expands into more deploys over its lifetime.
>
> – [The Twelve-Factor App](http://12factor.net/config)

GGAlanSmithee:
@samit4me 

interesting, never had any real problems with it. As long as the env is provided somhow (be it via `.env` or, heruko / dokku config) the application would function. For example `require('dotenv').config()` injects env variables only if it finds a `.env` file, othervise it "fails" silently.
 
@xavifuefer 

> My preference is to use a .env file for development and when the project gets deployed let's say on Heroku, you set the env variables there...

This is exactly what I'm doing, using dokku, never had any problems.

> As an alternative we could commit the .env file with development-only settings so other developers can get started on the boilerplate without compromising credentials for other environments.

It is generally adviced against comitting .env files, but in this case it makes total sense to provide it as a base.

samit4me:
This is great, would you be keen to write up some documentation describing how to set this up? Maybe in the **[Deployment](https://github.com/react-boilerplate/react-boilerplate/blob/8407cf05c4f1d43ca6c78625b7b8fee7627fdedb/docs/general/deployment.md)** section?

internals/webpack/webpack.base.babel.js
```diff
@@ -4,6 +4,30 @@
 
 const path = require('path');
 const webpack = require('webpack');
+const dotenv = require('dotenv').config(); // eslint-disable-line no-unused-vars
```
smashercosmo: Why create a variable if it is not used?

xavifuefer: the config method is the one responsible of loading the vars into the `process.env` object.

xavifuefer:
@samit4me do you mean documentation about the `.env` file and the env vars or how @GGAlanSmithee deals with them using dokku?

internals/webpack/webpack.base.babel.js
smashercosmo: I got it) But why does it need to be assigned to a variable?)

xavifuefer: what's your suggestion?
```
require('dotenv').config(); // without assigning it
```

or 

```
const dotenv = require('dotenv');
dotenv.config();
```

smashercosmo: I would prefer the first option, while requiring at the top of the file (like ```import 'babel-polyfill') ```). But that's just my personal preference) And I'm not a maintainer)

fugufish:
after adding these changes this im getting `Getting Module not found: Error: Can't resolve 'fs' in '...'` when using `yarn add`.

webpack starts up fine and everything runs as expected, however the error is distracting.

mparker11:
Has anyone figured out how to use package.json for defining environments? I'm trying to add development/production variable for things such as using dev/prod APIs but I keep getting this error: 

```
Unknown option: C:\folders\blah\project\package.json.development.VARIABLE.
A common cause of this error is the presence of a configuration options 
object without the corresponding preset name. Example:

Invalid: { presets: [{option: value}] }
Valid: { presets: [['presetName', {option: value}]] }
```

when I add

```
"development": {
    "VARIABLE": "value"
}
```

Any ideas?

internals/webpack/webpack.base.babel.js
```diff
@@ -1,10 +1,34 @@
...
+    // inside your code for any environment checks; UglifyJS will automatically
+    // drop any unreachable code.
+    NODE_ENV: JSON.stringify(process.env.NODE_ENV),
+  })
+;
+
+const StringifiedEnvVars = {
+  'process.env': Object
+    .keys(rawEnvVars)
+    .reduce((env, key) => {
```
thebergamo: To avoid disabling eslint you can use this approach:

```javascript
...
.reduce((env, key) => ({ ...env, [key]: JSON.stringify(rawEnvVars[key] }), {})
... 

```

gretzky:
Closing because this is incredibly out of date.

PR #1633, opened 2017-03-08
syntatical improvements to progress bar

GGAlanSmithee:
Just some minor syntatical improvements to the progress bar component;

* use class properties / static class properties where possible
* assign styled components to a variable before exporting it to allow for better dev experience (human readable name in chrome dev tools)


KarandikarMihir:
@GGAlanSmithee I just realized that `static propTypes` are not used in other files. I'd like to have uniform coding style. Let's have the `propTypes` outside of the class for now. Use of anonymous functions for class methods is acceptable because we haven't used those anywhere else. Does that make sense?

GGAlanSmithee:
@KarandikarMihir 

Updated it to not use the `static` keyword. (FYI it used it even before my PR).

KarandikarMihir:
@GGAlanSmithee No problem! Thanks a lot for your work! Merging when checks pass.

PR #1635, opened 2017-03-09
feat(core) add react performance profiling by default

jeremyadavis:
A small addition to give the boilerplate access the the React Performance Profiling tools by default.

As someone who recently finished  profiling an app after it was complete and regretting it, I am trying to keep a better handle on performance issues while working my next one.

This is the result of profiling with the example app by just doing the github name search:
rbp-reactperf

Read the [official docs](https://facebook.github.io/react/docs/perf.html) to see the other features.


justingreenberg:
i like this idea—will try to get reviewed and shipped with next release

jeremyadavis:
k, thanks

jeremyadavis:
FYI, looks like this is going away in React 16. [See Addons Sections](https://github.com/facebook/react/issues/10294)

Not sure the pull request is worth it if upgrading to 16 is going to be a priority.

justingreenberg:
closing since react 16 is coming up on roadmap, thank you for contribution :)

Issue #760, opened 2016-07-20
Feature - Rework setup UX to avoid new user confusion


gihrig:
As react boilerplate increases in popularity 😃 , new users become confused by system configuration errors (npn < 3 and failure to rerun setup after the first run, even though this feature is clearly documented).

Many people generally don't read the details (hard to imagine for software devs), and while this is their shortcoming, it places an unnecessary support burden on the react-boilerplate team. just search `is:issue setup`...

I'll open a PR as soon as I have something visible, might be a few days...

mxstbr:
Awesome, can't wait!

gihrig:
@richb-hanover [#754 (comment)](https://github.com/mxstbr/react-boilerplate/issues/754#issuecomment-234308419)
> Proposals:
> 
> 1. Don't worry about this. (I simply re-cloned the repo, and sat with the setup process running the second time.)
> 2. Detect the missing setup.js file and have the setup script mention that this could happen if it has been run twice.
> 3. Do something fancy with error recovery, etc. :-)
> 

4\. Leave setup in place, but make it very clear what is going to happen and require non-default action to run setup on an existing project. The idea was to facilitate experimentation by making it easy to return the project to original setup condition without having to got through a new clone/setup cycle.

mxstbr:
_ping_ any news here? Would love to land this for 3.3!

gihrig:
Yeah, I've been really busy, got started, distracted, sorry...

Since that time I've had a few thoughts:

Basically all that setup really accomplishes that's not already done in`npm install` is delete and replace`.git`.

What do you think about eliminating `setup` and adding a prompt to remove `.git` to `postinstall`?

This would clean up `package.json` and align the initial install setup process with 'industry standard' `npm install`.

Thoughts?

gihrig:
Well I've had some thoughts...

With all the testing I've been doing, I would like to see these changes.

1. deprecate `npm run setup` I install RBP at least 3-4 time per week sometimes more. I never use `npm run setup`, just run `yarn`.
2. I don't like auto deleting things like `.git` and `clean` plus `templates`. I say "give me the power to shoot myself in the foot, but inform me of what's going to happen and let me decide". To accomplish these goals I would:
- Change doc to recommend `yarn`, `yarn install`, or `npm install`.
- Change setup to show a deprecation message then run `yarn` or `npm install`
- Prompt to remove and replace `.git` at the end of the install process
[...]

PR #1653, opened 2017-03-20
Improve self destructive commands

gaoqiankun:
I think it is better to remove setup and clean command from package.json after executing.


gihrig:
@gaoqiankun I would rather see work on #760 

Also please comment on why this PR in that issue.

justingreenberg:
pushing to land #760 in next release, closing due to inactivity...

@gaoqiankun thank you for contributing, please feel free to join the discussion on #760

PR #1668, opened 2017-03-27
feat(internals): Only match files in extract-intl script (#1668)

zonr:
The bug is caught because we include chart.js[1] in our repository.

[1]: https://github.com/chartjs/Chart.js


justingreenberg:
looks good thank you @zonr!

PR #1697, opened 2017-04-06
Rename store.js to configureStore.js to prevent conflict

howardya:
There is an npm moduel called [storeJS](https://github.com/marcuswestin/store.js/)

If this module is installed, it will create conflict with the store.js in the app folder due to the alias.

```jsx
import store from 'store';
```

### DIFF
--- a/app/components/ProgressBar/Percent.js
+++ b/app/components/ProgressBar/Percent.js
@@ -1,7 +1,9 @@
 import styled from 'styled-components';
 
-export default styled.div`
+const ProgressBarPercent = styled.div`
   height: 2px;
   background: #29D;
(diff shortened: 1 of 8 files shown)


Dattaya:
@howardya, thanks! If you're still in the mood of contributing, please, also rename `store` to `configureStore` in every file that requires it, e.g. `app/containers/LanguageProvider/tests/index.test.js` and in `template` folder (`internals/templates/store.js`)

howardya:
@KarandikarMihir Please see PR #1904. Hope it is fine.

KarandikarMihir:
Closing in favor of #1904

PR #1698, opened 2017-04-06
Add Debugging section in Docs + VS Code instructions

auchenberg:
This PR moves the current WebStorm documentations from the FAQ to a new Debugging section, and adds VS Code instructions

Checks:


Dattaya:
I've added it to the milestone. @samit4me, can you review this PR?

docs/general/debugging.md
```diff
@@ -0,0 +1,74 @@
...
+```json
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Chrome",
+            "type": "chrome",
+            "request": "launch",
+            "url": "http://localhost:3000",
+            "webRoot": "${workspaceRoot}/src",
```
samit4me: Should this be `/app`?

docs/general/debugging.md
```diff
@@ -0,0 +1,74 @@
...
+            "userDataDir": "${workspaceRoot}/.chrome",
+            "sourceMapPathOverrides": {
+                "webpack:///src/*": "${webRoot}/*"
```
samit4me: Should this also be `app`?

auchenberg:
@samit4me You are right, hereby fixed. I also simplified the configuration a bit, as we have shipped a few improvements to our debugger.

samit4me:
Thanks for updating @auchenberg. I tried again but was unable to get it working. These are the steps I followed:

1. `git clone https://github.com/react-boilerplate/react-boilerplate.git`.
1. `cd react-boilerplate/`.
1. `git checkout dev`.
1. `yarn setup`.
1. Answered "n" to this `Do you want to clear old repository? [Y/n] n`.
1. Clicked "Debug" (menu bar) > "Open Configurations" (opens launch.json).
1. Replaced contents with what is listed in the updated docs and saved.
1. Opened terminal and ran `yarn start`.
1. Opened `/app/containers/HomePage/index.js` and placed a breakpoint on the first line of the render method.
1. Then clicked "Debug" > "Start Debugging".
1. A new Chrome window appeared and the site was running as expected, but in VSCode the breakpoints got disabled and were never ran and I got the following in the "DEBUG CONSOLE":

```
 Download the React DevTools for a better development experience: https://fb.me/react-devtools
SourceMaps.loadSourceMapContents: Could not download sourcemap from webpack:///~/redux-immutable/dist/index.js.map
SourceMaps.loadSourceMapContents: Could not download sourcemap from webpack:///./~/redux-immutable/dist/index.js?.map
SourceMaps.loadSourceMapContents: Could not download sourcemap from webpack:///~/redux-immutable/dist/combineReducers.js.map
SourceMaps.loadSourceMapContents: Could not download sourcemap from webpack:///./~/redux-immutable/dist/combineReducers.js?.map
SourceMaps.loadSourceMapContents: Could not download sourcemap from webpack:///~/redux-immutable/dist/utilities/index.js.map
SourceMaps.loadSourceMapContents: Could not download sourcemap from webpack:///./~/redux-immutable/dist/utilities/index.js?.map
SourceMaps.loadSourceMapContents: Could not download sourcemap from webpack:///~/redux-immutable/dist/utilities/getStateName.js.map
SourceMaps.loadSourceMapContents: Could not download sourcemap from webpack:///./~/redux-immutable/dist/utilities/getStateName.js?.map
SourceMaps.loadSourceMapContents: Could not download sourcemap from webpack:///~/redux-immutable/dist/utilities/getUnexpectedInvocationParameterMessage.js.map
SourceMaps.loadSourceMapContents: Could not download sourcemap from webpack:///./~/redux-immutable/dist/utilities/getUnexpectedInvocationParameterMessage.js?.map
SourceMaps.loadSourceMapContents: Could not download sourcemap from webpack:///~/redux-immutable/dist/utilities/validateNextState.js.map
SourceMaps.loadSourceMapContents: Could not download sourcemap from webpack:///./~/redux-immutable/dist/utilities/validateNextState.js?.map
[HMR] connected
```

Did you get these problems? I'm sure I've missed something! Are you able to list the steps you followed to get this working?

Also, it looks like the `dev` branch needs to be merged in.

Really keen to see this working, it would be amazing.

auchenberg:
I didn't checkout the dev branch, but master. Is Dev considered stable or
tip of tree?

auchenberg:
@samit4me Should be working now with Dev.

samit4me:
Thanks again @auchenberg. Yes, the `dev` branch is typically what we work on and once stable it's merged into master with each new release.

I followed the steps I outlined above and I still have the same problems, so I'm starting to feel quite stupid now. Are you able to follow them steps @auchenberg or @Dattaya and does it work for you? If not what did you do differently to get it working?

Out of interest, I also repeated all the above steps on the `master` branch and I no longer got all the errors in the "DEBUG CONSOLE" but the breakpoints still got disabled and did not work. If I hover of the disabled breakpoint I get a message `Breakpoint ignored because generated code not found (source map problem?).`.

Thinking it was just me or something environmental, I then created a new repo using create-react-app and followed the steps on the medium post and it worked as expected and I was able to successfully debug.

If we can get this working, could you please also merge or rebase in the latest changes from the `dev` branch as GitHub is now reporting it is "out-of-date".

Another topic up for discussion is the checking in of project files into the repo. I know in the [`.gitignore` file](https://github.com/react-boilerplate/react-boilerplate/blob/f1a167c47bfa78e4662c3193b874ce8d426d98d5/.gitignore#L10) we ignore intellij project files, but I've noticed this PR has added `.code` project files. I would suspect that `.code` should also be ignored, so we can either remove the `.code` from this PR and/or update the `.gitignore` file.

auchenberg:
@samit4me I think we are testing two different things. 

Setting a breakpoint in `/app/containers/HomePage/index.js` doesn't break in VS Code or Chrome DevTools for that matter (seems to be a webpack source mapping issue). I've been testing setting breakpoints in `app/components/ReposList/index.js`, and it works. (See https://youtu.be/p9uLt9uqceU)

The `.code/launch.json` is a predefined configuration file for VS Code, which contains the same configuration as the docs. Having this included will enable VS Code to work out of the box.  It's your call whether you want this or not.

There's a source mapping difference between `dev` and `master` which is the reason for another configuration is needed. I've updated the configuration to cover both scenarios.

Dattaya:
@samit4me the default `eval-source-map` didn't work for some reason, maybe because of issues with source maps Webpack has in Chrome [as described in their docs](https://webpack.js.org/configuration/devtool/#devtool). but `inline-source-map` does work, just press Restart button (Ctrl+Shift+F5) after you started debugging, and then:

image

Those warnings are still there though.

samit4me:
Thank you for all the info @auchenberg and @Dattaya, much appreciated ✨ 

> Setting a breakpoint in /app/containers/HomePage/index.js doesn't break in VS Code or Chrome DevTools for that matter (seems to be a webpack source mapping issue). I've been testing setting breakpoints in app/components/ReposList/index.js, and it works. (See https://youtu.be/p9uLt9uqceU)

That's awesome, thanks for taking the time to produce a video 👍, so good! I tried following the steps in the video by placing a breakpoint in the `<RepoList>` but still did not work 😞 

If what you say is true and it only works on some files and not others, we probably have to fix that issue before landing this otherwise there will be a flood of issues reporting that it does not work.

After getting it working (see down below) I also noticed that the debugger would not work if the devtools in Chrome were open. This is might be normal, but it's probably worth mentioning this in the docs if this is indeed true.

> The .code/launch.json is a predefined configuration file for VS Code, which contains the same configuration as the docs. Having this included will enable VS Code to work out of the box. It's your call whether you want this or not.

I personally prefer to have project config files checked into the repo but I also know a lot of developers that do not like it. In a small team it's easy to make a decision on this, but as an open source boilerplate project, it's probably a safer choice to not included editor config and add some docs describing how to setup stuff (e.g. debugger in this case).

> inline-source-map does work

💯 Just tested and it certainly works, no doubt about it and the source code is still super readable which is good. Will have to look further into this to see if there are any other implications to switching over to this. I looked into create-react-app as that was working for me and it is using `cheap-module-source-map`, but that does not work in this boilerplate. 

Thanks for all the effort, I am pretty short on time at the moment but still super keen to get this merged in. Not sure I'll have another chance to look before the weekend.

auchenberg:
@samit4me 

> If what you say is true and it only works on some files and not others, we probably have to fix that issue before landing this otherwise there will be a flood of issues reporting that it does not work

I see this as issue that's unrelated to landing this documentation change, and something that should be reported to the vscode-chrome-debugger repo.

> After getting it working (see down below) I also noticed that the debugger would not work if the devtools in Chrome were open. This is might be normal, but it's probably worth mentioning this in the docs if this is indeed true.

This is fully documented in https://github.com/Microsoft/vscode-chrome-debug, and expected functionality. I would expect developers to read the documentation for the products they use, instead of expecting these docs to include everything.

> I personally prefer to have project config files checked into the repo but I also know a lot of developers that do not like it. In a small team it's easy to make a decision on this, but as an open source boilerplate project, it's probably a safer choice to not included editor config and add some docs describing how to setup stuff (e.g. debugger in this case).

Config file has been removed.

Dattaya:
@samit4me, IMO this one is ready, what do you think?
@auchenberg, thank you so much for not giving up on this PR it has been left unanswered for such a long time, sorry.

samit4me:
Not sure @Dattaya, I've never used the VSCode debugger before, so I found it a little difficult to get setup. I believe this was amplified by the fact I couldn't actually get it working. I had to change the source map type to `inline-source-map` as you suggested.

Most of the issues are proabably either, related to this codebase or upstream issues. I understand these are outside the scope of this PR, but I believe we should identify and resolve these other issue before we land this PR.

Dattaya:
@samit4me, I can create a follow up PR that recommends to use `inline-source-map` for debugging (alternatively—replace our default `eval-source-map` with `inline-source-map` or `cheap-module-source-map` if it's good enough), otherwise in my experience debugging doesn't work very well in Chrome and every Chrome based debugger. It's a problem with webpack in Chrome.

auchenberg:
This source mapping problem is present in Chrome DevTools too, and is not related to VS Code.

samit4me:
Creating another issue to investigate changing to `inline-source-map` would be good :) If we could update the docs to mention that there are potential issues and suggest changing to `inline-source-map` and add a link to the [Chrome debug page](https://github.com/Microsoft/vscode-chrome-debug) for more info, that would probably be more than sufficient. What do you think?

auchenberg:
@samit4me What's holding this documentation change back? It's been 2 months now.

Dattaya:
I've noticed that @samit4me is no longer part of the team; merging with one approval as it's allowed for contributions to the documentation. Thanks again, @auchenberg 🌹.
@samit4me, thanks for the thorough review!

PR #1711, opened 2017-04-17
feat(internals): Add husky git hooks with Prettier integration

simobasso:
use [`husky`](https://github.com/typicode/husky) instead pre-commit
this is useful if we want to use other git hooks than pre-commit

run test before push
in order to avoid broken pr (and waste of time), we can use a git
hook for run tests before push.

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,8 @@
     "plugins": [
       "redux-saga",
       "react",
-      "jsx-a11y"
+      "jsx-a11y",
+      "prettier"
     ],
     "parserOptions": {
       "ecmaVersion": 6,
@@ -156,7 +157,8 @@
       "react/self-closing-comp": 0,
       "redux-saga/no-yield-in-race": 2,
       "redux-saga/yield-effects": 2,
-      "require-yield": 0
+      "require-yield": 0,
+      "prettier/prettier": "warn"
     },
     "settings": {
       "import/resolver": {
@@ -271,6 +273,7 @@
     "eslint-import-resolver-webpack": "0.8.3",
     "eslint-plugin-import": "2.7.0",
     "eslint-plugin-jsx-a11y": "5.0.3",
+    "eslint-plugin-prettier": "^2.3.1",
     "eslint-plugin-react": "7.0.1",
     "eslint-plugin-redux-saga": "0.3.0",
     "eventsource-polyfill": "0.9.6",
@@ -288,6 +291,7 @@
     "null-loader": "0.1.1",
     "offline-plugin": "4.8.1",
     "plop": "1.8.0",
+    "prettier": "1.7.4",
     "react-test-renderer": "15.6.1",
     "rimraf": "2.6.1",
     "shelljs": "0.7.7",
--- a/package.json
+++ b/package.json
@@ -43,7 +43,8 @@
     "test": "cross-env NODE_ENV=test jest --coverage",
     "test:watch": "cross-env NODE_ENV=test jest --watchAll",
     "coveralls": "cat ./coverage/lcov.info | coveralls",
-    "precommit": "npm run lint:staged"
+    "precommit": "npm run lint:staged",
+    "prepush": "npm run test"
   },
   "lint-staged": {
     "*.js": "lint:eslint"
--- a/package.json
+++ b/package.json
@@ -42,12 +42,12 @@
     "test:clean": "rimraf ./coverage",
     "test": "cross-env NODE_ENV=test jest --coverage",
     "test:watch": "cross-env NODE_ENV=test jest --watchAll",
-    "coveralls": "cat ./coverage/lcov.info | coveralls"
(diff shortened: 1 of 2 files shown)


KarandikarMihir:
@simobasso I agree with the concept of running tests before pushing changes. Let's hear from other maintainers and community members before we merge this in.

Aftabnack:
Maybe this could be opt in. It's a huge mood killer for me if I'm in the zone and it takes forever (3-5 mins is forever.:P) for me to push my code. 

We have built a setup in which we have testcases running in parallel with your development and it sort of notifies you and blocks you from developing until you fix it.

IMO this seems more proactive and productive for me than running it on push mainly because of two things. **a)** It runs in parallel and not before I push **b)** It fail fasts, every line that I chime in should be something that I don't break already existing test cases.

Thoughts? @KarandikarMihir @justingreenberg @gihrig @Dattaya

justingreenberg:
@simobasso @Aftabnack i would like to see this land!

husky is nice for creating integrated commit workflow, ie we can use hooks with https://github.com/okonet/lint-staged eg. to format staged files with Prettier (see RFC #1945)

peter-mouland:
I use husky for both, pre-commit and pre-push. Pre-commit for linting and fast unit tests.  I then use pre-push for functional/e2e tests.

@Aftabnack, I also like to commit and push often when dev'ing.  so i use `-n` or `--no-verify` when on feature branches.  for me, the hooks don't get in the way at all

simobasso:
Hi! If this pr it's ok for you, please ping me and I'll fix conflicts, rebase and update this one. As @peter-mouland we use pre-commit and pre-push every day.

So please, let me know. :)

Aftabnack:
Since I can override this when needed by `-n`. LGTM.

simobasso:
@Aftabnack @justingreenberg 🎉  
please check 2° commit and use that for https://hacktoberfest.digitalocean.com/ 🎃 

let me know what configuration we want 🌮 !

for problems example check travis log.

package.json
```diff
@@ -42,12 +42,13 @@
     "test:clean": "rimraf ./coverage",
     "test": "cross-env NODE_ENV=test jest --coverage",
     "test:watch": "cross-env NODE_ENV=test jest --watchAll",
-    "coveralls": "cat ./coverage/lcov.info | coveralls"
+    "coveralls": "cat ./coverage/lcov.info | coveralls",
+    "precommit": "npm run lint:staged",
+    "prepush": "npm run test"
   },
   "lint-staged": {
     "*.js": "lint:eslint"
```
justingreenberg: i think we need to `git add` the formatted results...

```
...
"lint-staged": {
  "*.js": [
    "prettier --write",
    "git add"
  ]
}
```

justingreenberg:
note: once this is approved and ready to merge, we will want a **_separate PR_** that runs Prettier on the entire codebase using repo config, so we can merge subsequently.

we should also include an npm script to format project eg `"prettier:all": "prettier --write ."` so that we can easily apply new versions of Prettier

simobasso:
hi @justingreenberg thanks for the fast reply

> 1. do we still need to run eslint since it would be expected to match prettier output anyway?

IMHO change 2300+ LOC in one commit is a bad thing 😅 

> 2. wouldn't eslint just print warnings to console? i think we need to git add the formatted results...

yes... and no 😄 

Yes for `precommit` (my fault) but no for `prepush` because now the codebase have 2k+ of "problems" with this config, so I prefer to do a gracefully introduction of `prettier` and `lint staged` (git add) works only on staged and `prepush` is after the stage and get all codebase.

Maybe we can split this one in two pull request, one for husky and one for prettier and in the last one we can add the `npm` command and the full clean on the codebase.

please let me know what you prefer!

gihrig:
> 1. do we still need to run eslint since it would be expected to match prettier output anyway?

In my experience, yes. prettier and eslint are different beasts, expecting compatible results from both has proven elusive.

That's why there is [prettier-eslint](https://github.com/prettier/prettier-eslint)

> Formats your JavaScript using prettier followed by eslint --fix

There's also a companion [prettier-eslint-cli](https://github.com/prettier/prettier-eslint-cli)

This combined with "Prettier-Javascript formatter" VC Code extension gives me formatting from within the editor, that passes cli format operation.

justingreenberg:
>Maybe we can split this one in two pull request, one for husky and one for prettier and in the last one we can add the npm command and the full clean on the codebase.

@simobasso this sounds like a solid plan :) thank you!!

@gihrig thank you for insight, as always!

blling:
Friendly ping.
Husky have been added in #2072,

simobasso:
Hi @bcowgill sorry if I disappeared. I've been busy with work. I'll submit a new pr asap

blling:
Prettier integration has been done in #2083, that will be fine if you could do some review.
I will close this PR, thanks.

Issue #1613, opened 2017-02-25
Dev Branch: Route sagas are injected twice on Reload/init


raduflp:
The described issue is only reproducible on the current Dev branch.

On reload (F5), the route sagas are injected twice. As an effect, you get duplicate network calls. 
When moving between paths, it seems to be working correctly.

I'm just starting with this boilerplate, I'll try to look into it myself if I get a chance,

Radu-Filip:
Looks like dependencies update broke this in #1501 . Maybe @gihrig or someone else more experienced than myself can look into it.

gihrig:
@Radu-Filip Thanks for researching this. 

It's a known issue, corrected in #1301 where react-router-redux was rolled back to 4.0.6. The update to 4.0.7 in #1501 was an error 😞 .

Have you confirmed 4.0.8 no longer has the problem by actual test?

Thanks again for your help. ✨

Radu-Filip:
Yes, I've tested it with 4.0.8 and I can't reproduce the issue. Looks like it was an issue with the way History 3.x  works, and that was fixed [here](https://github.com/reactjs/react-router-redux/commit/a2825ac0b8e34662624455ab98e29e2eb27f8096)

justingreenberg:
this should be resolved by #1746 — feedback is welcome! thank you!

Issue #1554, opened 2017-02-04
Yet another global- or App-level saga discussion :-/


marcandrews:
To not take away anything from @RobertSheaO's amazing work in PR #1545, I created this issue to continue a discussion about where the code for global- or app-level sagas should reside.

In response to @gihrig's [comment](https://github.com/react-boilerplate/react-boilerplate/pull/1545#issuecomment-277191565) ...

In this case, we're using sagas, but no matter which method you choose to manage side-effects, there is still the issue of where the _code for "global" side-effects_ reside. 

The three places where this code could reside (summarized by the issue numbers @gihrig listed in his [comment](https://github.com/react-boilerplate/react-boilerplate/pull/1545#issuecomment-277191565)), seem to be: 
1) Define your actions/reducer/sagas/selectors anywhere in your project, and then inject these sagas along with each route.
2) Define your actions/reducer/sagas/selectors in `/app/containers/App` and inject these sagas in a root-level component route (like this PR).
3) Define your actions/reducer/sagas/selectors in their own container. For example, for authentication purposes, these would be defined in `/app/containers/SessionProvider` similar to how `/app/containers/LanguageProvider` is defined. Then, this SessionProvider is imported into `/app/app.js` and included in the `ReactDOM.render` method (just like the LanguageProvidor).

(1) seems cumbersome.

(2) seems like a better approach, but something like session information and components related to session like sign in interface and account management may clutter the App container (which may be better suited to presentation only). This is where I am in my own project: an App container with it's defined role unclear. So if I move the sign in/account management into their own container, then those sagas should go with them. This leads me to:

(3), a completely modular solution to the problem. My authentication actions/reducer/sagas/selectors are defined in `/app/containers/SessionProvider`, along with all the UI components and additional state necessary session management. This leaves the App container for presentation only, and in the future, I can easily remove/replace/refactor this provider without having to tease it away from the App container.

Looking forward to any input on this.

sbruton:
given that the SessionProvider container isn't mapped to a route, where are you injecting the sagas?

marcandrews:
@sbruton I imported my session provider saga in app/app.js right after the store is initialized:

```
import SessionProviderSagas from 'containers/SessionProvider/sagas';
...
export const store = configureStore(initialState, browserHistory);

SessionProviderSagas.map(store.runSaga);
```

kai23:
Do you have any full example of this ?

gretzky:
Closing due to inactivity

PR #1734, opened 2017-04-27
Add a memoized getComponent wrapper so that components do not get reloaded

trevorharwell:
## React Boilerplate

### DIFF
--- a/app/containers/HomePage/sagas.js
+++ b/app/containers/HomePage/sagas.js
@@ -2,8 +2,7 @@
  * Gets the repositories of the user from Github
  */
 
-import { take, call, put, select, cancel, takeLatest } from 'redux-saga/effects';
-import { LOCATION_CHANGE } from 'react-router-redux';
+import { call, put, select, takeLatest } from 'redux-saga/effects';
 import { LOAD_REPOS } from 'containers/App/constants';
 import { reposLoaded, repoLoadingError } from 'containers/App/actions';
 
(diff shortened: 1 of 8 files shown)


trevorharwell:
As a side note to this PR, this resolves a lot of issues people might have with sagas being re-ran when navigating from dynamic route to dynamic route.

trevorharwell:
Not sure how to resolve the build failure.  It failed on node v5 in a file that I didn't change.

colek42:
"As a side note to this PR, this resolves a lot of issues people might have with sagas being re-ran when navigating from dynamic route to dynamic route."

We were having this issue.  This pattern solved it for us/.  Thanks @trevorharwell

Andarist:
@trevorharwell Aint sure if this memoizing approach is the right way to do this. It might be worth doing in general, but not as the mentioned saga 'problem'.

There is a place for both saga scenarios - one which gets cancelled when you leave the route and one which will stay live forever for the application lifespan. Your solution kinda hides the symptom only and we probably should expose 2 different `injectSagas` APIs (for both use cases) and educate about them.

WDYT?

yeluolei:
how can we fix this issue?

gihrig:
> how can we fix this issue?

There has been a lot of discussion on this and several suggested solutions, but no consensus yet on the best way to go.

For starters, be sure that #1613 is not the problem.

Then check out #1554, #1545 and the issues it references for a thorough background on this issue.

justingreenberg:
this issue should be resolved by #1746 :) would love any feedback! thank you very much again for your contribution! please let me know if you have any questions

Issue #1434, opened 2017-01-10
Sagas are injected twice!?


williamgranli:
For some reason, the sagas are injected twice in a new project I started. I did not have this problem in my previous project, even though the code is the same from what I can tell. 

routes.js

```
import { getAsyncInjectors } from 'utils/asyncInjectors';

const errorLoading = (err) => {
  console.error('Dynamic page loading failed', err); // eslint-disable-line no-console
};

const loadModule = (cb) => (componentModule) => {
  cb(null, componentModule.default);
};

export default function createRoutes(store) {
  // Create reusable async injectors using getAsyncInjectors factory
  const { injectReducer, injectSagas } = getAsyncInjectors(store); // eslint-disable-line no-unused-vars
  console.log("in route");
  return [
    {
      path: '/',
      name: 'consolePage',
      getComponent(nextState, cb) {
        const importModules = Promise.all([
          System.import('containers/ConsolePage/reducer'),
          System.import('containers/ConsolePage/sagas'),
          System.import('containers/ConsolePage'),
        ]);

        const renderRoute = loadModule(cb);

        importModules.then(([reducer, sagas, component]) => {
          injectReducer('consolePage', reducer.default);
          injectSagas(sagas.default);
          renderRoute(component);
        });

        importModules.catch(errorLoading);
      },
    }, {
      path: '*',
      name: 'notfound',
      getComponent(nextState, cb) {
        System.import('containers/NotFoundPage')
          .then(loadModule(cb))
          .catch(errorLoading);
      },
    },
  ];
}
```

ConsolePage/sagas.js

```
export function* root() {
  console.log("starting saga")
  const watcher1 = yield fork(a);
  const watcher2 = yield fork(b);
  const watcher3 = yield fork(c);
  const watcher4 = yield fork(d);
  const watcher5 = yield fork(e);

  yield take(LOCATION_CHANGE);
  yield cancel(watcher1);
  yield cancel(watcher2);
  yield cancel(watcher3);
  yield cancel(watcher4);
  yield cancel(watcher5);
}

export default [
  root,
];
```
Where a, b, c, d and e are generator function as below:
```
export function* a() {
  while (true) { // eslint-disable-line no-constant-condition
    const action = yield take(X);
    const id = action.payload.id;
    const url = path + ':10020/' + id + '/logs';
    const response = yield call(y, url);
    if (response) {
      const logs = response.data;
      yield put({ type: z, payload: { asd } });
    }
  }
}
```

[...]

PR #1746, opened 2017-05-08
Migrate to react router v4.

### LINKED ISSUE #1033: React router v4
I've been watching the rollout of react router v4 and it looks like a pretty exciting (but dramatic) departure from v2 best practices. Since the router is such a major part of the boilerplate I'm wondering what general opinion is on integrating it. It seems like this would be a major rewrite, but I'm curious what others are thinking about this.

anuraaga:
React router v4 is a big change in paradigm, resulting in big changes in routing-related generators and templates. I'm hoping that future PRs can re-enable true hot-reloading after this is in.

Routes have been migrated to a component, and loading logic has been moved to separate modules instead of inlining to make it easier to define child routes.

App has been made a connected component to ensure it always has the location as a prop to ensure route rerendering even if its changed to a PureComponent or redux component.

Fixes #1033

## React Boilerplate

- [X] You have followed our [**contributing guidelines**](https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md)
- [X] double-check your branch is based on `dev` and targets `dev` 
- [X] Pull request has tests (we are going for 100% coverage!)
- [X] Code is well-commented, linted and follows project conventions
- [X] Documentation is updated (if necessary)
- [X] Internal code generators and templates are updated (if necessary)
- [X] Description explains the issue/use-case resolved and auto-closes related issues


vicbell:
Could you give a suggestion to manage saga lifecycle like #1420 ?

anuraaga:
@vicbell react router callbacks are replaced by just using react component lifecycle. I think it'll be easy to achieve the same effect by updating `AsyncRoute`

- Split loader and injector logic of components
- In AsyncRoute, add `componentWillMount` which injects the sagas and stores a reference on `this`, and `componentWillUnmount` which cancels them.

Given how big this PR is with the minimum to add react router v4, I'd like to work on the saga lifecycle in a subsequent one if that makes sense.

vicbell:
Great work. Thanks !

vicbell:
There is a warning message when start up dev_rag branch

npm run start

react-boilerplate@3.4.0 start /home/bfan/git/react-boilerplate
cross-env NODE_ENV=development node server

Server started ! ✓

Access URLs:
Localhost: http://localhost:3000
LAN: http://172.26.3.83:3000

Press CTRL-C to stop
    
(node:5299) DeprecationWarning: loaderUtils.parseQuery() received a non-string value which can be problematic, see https://github.com/webpack/loader-utils/issues/56
parseQuery() will be replaced with getOptions() in the next major version of loader-utils.

webpack built 1435bb44a8547abe98d8 in 3697ms

anuraaga:
@vicbell I can't reproduce any warning on my Macbook. Did you make sure to run `yarn` to have the same dependencies?

This has no warning
```bash
$ git clone git@github.com:anuraaga/react-boilerplate.git react-boilerplate-test
$ cd react-boilerplate-test
$ git fetch
$ git checkout dev_rag
$ yarn
$ npm run start
```

vicbell:
I get a network trouble if i use yarn..., Mybe it is my mistake at somewhere.
```java
[1/4] Resolving packages...
[2/4] Fetching packages...
warning There appears to be trouble with your network connection. Retrying...
warning There appears to be trouble with your network connection. Retrying...
warning There appears to be trouble with your network connection. Retrying...
warning There appears to be trouble with your network connection. Retrying...
error An unexpected error occurred: "https://registry.yarnpkg.com/compression/-/compression-1.6.2.tgz: tunneling socket could not be established, cause=connect EINVAL 0.0.31.144:80 - Local (0.0.0.0:0)".
```

anuraaga:
Any maintainer able to take a look at this? @KarandikarMihir maybe?

FWIW, in an app where I'm using this code, I've also gone ahead and migrated to react-hot-loader v3 and have perfect live reload with async routes - we can get back to the happier days of react-boilerplate :)

DenisNeustroev:
Create some child route and you have got a not found page.

anuraaga:
@DenisNeustroev Did you follow the updated instructions in the README?

https://github.com/anuraaga/react-boilerplate/blob/3be5a26a752b93d24581bee3c27fb1fbbca6cbd6/docs/js/routing.md

I'm not having any problem (the trickiest part is probably remembering to remove "exact" from the routes.js route for the parent page.

If you still have problems, please push a branch so I can check it out.

DenisNeustroev:
Yes, my bad, its works like a charm. But there is another problem. I don't use PureComponent, and on route change my parent container re-renders twice, the first is on location change, the second when a child route loads itself, so AsyncRoute try load child route twice.

DenisNeustroev:
Also AsyncRoute can pass store in loader func itself instead of doing it in every container

DenisNeustroev:
ezgif com-video-to-gif

anuraaga:
@DenisNeustroev Interesting, that does seem weird. I'd like to repro it, but a simple attempt with parents and children, all Component, with console.log in render() doesn't show anything out of the ordinary. Can you look at ParentPage and ChildPage1/2 in this branch to see what the difference might be with your use case?

https://github.com/anuraaga/react-boilerplate/tree/dev_rag_parent_child
Navigate to http://localhost:3000/parent and click on a child link - in the console, rendering child only shows once.

NantawatS:
When I try using histroty.push() inside a async function it doesn't work. The location was changed but still stay on the same page.

```
<button
    onClick={() => {
        request.get('xxx', ()=> {
            history.push('/afterlogin');
        });
    }}
>link</button>
```

But if I do some think like this, it's work.

```
<button
    onClick={() => {
        request.get('xxx', ()=> history.push('/afterlogin'));
    }}
>link</button>
```

Any ideas?

NantawatS:
I solved the issue using this approach. ([link](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/blocked-updates.md)) It's not this PR issue.

leifdalan:
@anuraaga You said you have react-hot-loader v3 working with this branch? I can't seem to get it to work with stateful components. Do you have a gist of your `app/app.js` file?

anuraaga:
@leifdalan Here's the diff where I added hot-reload to a project.

https://gist.github.com/anuraaga/782b017d493f2cc20af85cd9eb4634d1

The project has Component, PureComponent, and SFC and all seem to update ok - none of them use setState though, if that's what you mean by stateful component, I'm not sure whether that would have a different result. If there's a simple repro, would be happy to take a look.

internals/templates/containers/App/index.js
```diff
@@ -12,18 +12,26 @@
...
       <div>
-        {React.Children.toArray(this.props.children)}
+        <Routes />
       </div>
     );
   }
 }
+
+const mapStateToProps = createStructuredSelector({
+  location: makeSelectLocation(),
```
wer32: Why do we need location here. the variable is not used.

anuraaga: Without having location in the props, App will not rerender on location change if the component is pure or connected, as react-router-v4 relies on normal react rerendering semantics. Even if location isn't used, I think it's a good default to pass location to App to minimize rerender issues.

internals/templates/routes.js
```diff
@@ -2,43 +2,36 @@
...
+class Routes extends React.PureComponent { // eslint-disable-line react/prefer-stateless-function
+  render() {
+    const store = this.context.store;
+    return (
+      <Switch>
+        <AsyncRoute
+          exact path="/" load={createHomePageLoader(store)}
+        />
+        <AsyncRoute
+          exact path="" load={createNotFoundPageLoader(store)}
```
wer32: Does this work?
I've can't enter to NotFoundPage with not registered url.
Probably this should be corrected with "*" value

wer32:
I found issue in current pull request.

v4 router issue

vicbell:
@wer32   I also have the same issue.

anuraaga:
@wer32 @vicbell Fixed the route re-rendering on back button issue, and also switched not-found matching to non-exact (for some reason it was still working with exact for me, though indeed it seems incorrect).

KarandikarMihir:
Oh my god! So much work being done here. Thanks @anuraaga for all your efforts. I'll try it out this weekend! Thanks again.

wer32:
@anuraaga @KarandikarMihir I totally agree. Great work! loader.js for components are awsome and elegant.

peter-mouland:
I too love the approach of the loaders... 
Is slightly off topic, but out of interest to gauge public opinion...
Do people who do this recommend the approach of making _every_ route async, 
**or** 
Do you generally only make less frequented pages async, like admin and or faq's etc?

anuraaga:
@KarandikarMihir Thanks! Was starting to lose hope this would get reviewed, looking forward to it - and have many more changes to follow up with too, so hope to keep up the momentum :)

@peter-mouland I personally make every route async and haven't needed to optimize further yet, but if I did I'd preload the other main components - get fast first-page load while still having fast loads of other main pages. react-router provides a good example of this, it's very simply just calling the loaders yourself in App.componentDidMount

https://reacttraining.com/react-router/web/guides/code-splitting/loading-after-rendering-is-complete

florimm:
After clean I'm getting this error:
ERROR in ./app/app.js
Module not found: Error: Can't resolve 'react-router-scroll' in '......\app'
 @ ./app/app.js 17:0-48
 @ multi main
package 'react-router-scroll' should be removed.

anuraaga:
@florimm Did you happen to have app.js open from another branch and saved after cleaning? I can't think of anything else that would cause that behavior, this branch is working ok even after clean.

The template that gets used after clean is here, which doesn't import react-router-scroll
https://github.com/anuraaga/react-boilerplate/blob/2bcf0978e8256052a76c776991edb5e63d0c77f1/internals/templates/app.js

app/routing/AsyncRoute.js
```diff
@@ -0,0 +1,74 @@
...
+}
+
+Bundle.propTypes = {
+  children: React.PropTypes.node,
+  load: React.PropTypes.func,
+};
+
+const AsyncRoute = ({ computedMatch, path, load }) => (
+  <Route
+    path={path} computedMatch={computedMatch} render={(props) => (
```
mhssmnn: Shouldn't AsyncRoute also pass on the `exact` prop to Route?

anuraaga: Thanks! Fixed to propagate all props other than load (and noticed I had forgotten to fix the template too). Actually, it was still working because computedMatch takes precedence over the path matching of the route (the Switch already did the matching based on the props on AsyncRoute), but this is definitely a small implementation detail and generically passing all default props is better.

santaclauze:
Hey, firstly thanks for the work man!

I have been trying to implement react-router-dom (note that i'm a dev-beginner ).

My app is not displaying anything and after a few logs and comparing with your  **anuraaga:dev_rag**, for some reason my AsynchRoute component seems to call the props loader to early.

router-warning
 
Also, the `load` props it not called in the return of the `(cb)` from the `loader` file like it does on your project. It calls a `function(store)`

I have pretty much followed the same implementation of the PR except I use Logic, but I doubt that changes anything.

The other difference I observe, is that when creating a `createHistory()` my object does contain a `location` object but `location.key` is undefined while on your project it contains a generated key.

1. Why is the props.load called prior to passing in the loader?
1. Why is my props.load calling the whole loader function rather than just the return of the function?
1. Is it normal that my `location.key` is not generated?

My impression is that I am not passing the `story` properly to the `load` props. 

Also, my apologies if this is not the best place for the question.

app/containers/FeaturePage/loader.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Asynchronously loads the components for FeaturePage
+ */
+
+import { errorLoading } from 'utils/asyncInjectors';
+
+export default (store, cb) => {
```
kevin940726: Thanks for your work! This is awesome, I learned a lot. Just a small question here.

Why do we have to use the callback style? Can't we just return a promise?

SanskarSans:
I could not make it work for multiple layout. I get the same homepage when using separate routes file. I have asked this question on SO too but did not get any help. 
```
const render = messages => {
  ReactDOM.render(
    <Provider store={store}>
        <ConnectedRouter history={history}>
          <App />
        </ConnectedRouter>
    </Provider>,
    document.getElementById("app")
  );
};

import Routes from 'routes';
class App extends React.Component {
render() {
  return (
    <div>
      <Navbar userForm={this.handleDialog} />
      <Routes />
    </div>
  );
}
}

routes.js
function Routes({ location }) {
  return (
    <Switch location={location}>
      <AsyncRoute exact path="/" load={loadHomePage} />
      <AsyncRoute exact path="/features" load={loadFeaturePage} />
      <AsyncRoute path="" load={loadNotFoundPage} />
    </Switch>
  );
}```
I want now admin dashboard a complete new page with different layout not children of App component so i did the following
```javascript
class AdminDashboard extends React.Component {
  render() {
    return (
      <div>
        <TopNavigation />
        <SideNavigation />{/* it will have link */}
        <Routes /> {/* it will display page */}
      </div>
    );
  }
}

function AdminRoutes({ location }) {
  return (
        <Switch location={location}>
          <AsyncRoute
            exact
            path="/admin/dashboard"
            load={loadAdminDashboardPage}
          />
          <AsyncRoute exact path="/commission" load={loadCommissionPage} />
          <AsyncRoute path="" load={loadNotFoundPage} />
        </Switch>
  );
}
```
when i hit the url /admin/dashboard i get the app page not the admin dashboard page and same with the  /commission which should be a child of AdminDashboard

How can i make my router work for different layout?

anuraaga:
@Dattaya Thanks for the review! I realized that now that we handle async loading ourselves, instead of react-router, we also have to handle tracking of the loaded routes. I have hooked `Bundle` and `ProgressBar` into the redux store to allow this to happen and the progress bar is working for me again. It's a pretty big change, so can you review the commit once to make sure the approach is ok? Then I'll update the template as well and address the other comments

https://github.com/react-boilerplate/react-boilerplate/pull/1746/commits/9f44f1b95ce584ccf84e3a92c88dac64cd2faf78

@SanskarSans Everything is rendered top-down starting with App - this means that even the admin dashboard parent page must be in the normal routes.js, with a non-exact path. It will be something like

routes.js
```jsx
  <Switch location={location}>  
      <AsyncRoute exact path="/" load={loadHomePage} />
      <AsyncRoute exact path="/features" load={loadFeaturePage} />
      <AsyncRoute path="/admin/dashboard" load={loadAdminDashboardParent} />
      <AsyncRoute path="" load={loadNotFoundPage} />
    </Switch>
```

AdminDashboardParent/index.js
```jsx
return (
  <Switch location={location}>
    <AsyncRoute exact path="/admin/dashboard/commission" load={loadCommisionPage} />
  </Switch>
);
mapStateToProps = ({
  location: makeSelectLocation();
})
```

I tried to document this in https://github.com/anuraaga/react-boilerplate/blob/3be5a26a752b93d24581bee3c27fb1fbbca6cbd6/docs/js/routing.md but let me know if there's any way I can make this documentation more clear.

@santaclauze I made a commit during the progression of this PR that moved store propagation into the context to make it easier to define child routes. This PR and the dev_rag branch are updated - I suspect you may have updated your code to the latest version of the PR but need to regenerate the loader files which also need to be updated.

MilanRgm:
Wow this is really awesome to know, react-router v4 in react-boilerplate. That's also with async injection of component, sagas and reducers. I want to use this boilerplate but i am not an expert so how can we implement public and private route here?

anuraaga:
@MilanRgm I think this is a good example of react-router-v4 with private / public routes - it shouldn't be too hard to wrap AsyncRoute with AuthenticatedRoute which decides between rendering `AsyncRoute` (with all props propagated through) or `Redirect` based on the auth state.

https://github.com/tylermcginnis/react-router-firebase-auth

SanskarSans:
@anuraaga Thanks for your reply. What i wanted to make is, there will be homepage for end user but if the user is admin, he/she would have dashboard which is completely different from top to bottom. 

For end user it would be like this which is working right now .

Image of homepage

for admin it would be like this

Image of dashboard

for homepage what i did is 

containers/App/index

```javascript
class App extends React.Component {
  render() {
    return (
      <div>
        <Routes />
      </div>
    );
  }
}
```
In containers/Homepage/index i imported all the component i need to show in homepage like navbar, destination etc because for route / the path is give to loadHomePage in routes.js 

Now i want AdminDashboard to have complete layout, with parent layout be AdminDashboard/index.js and all the child routes will be rendered there as 

```javascript
const AdminDashboard = () => return ( <SideNavigation /> <Routes />);
```
Now i am not sure where should i define routes. Should i define it in main routes.js file or fetch the state of auth and if the user is admin, show AdminComponent, if the user is owner show OwnerAdmin, otherwise show Homepage component inside App/index? Can you tell me the best way to handle such condition?

MilanRgm:
@anuraaga Do you mean to say something like this ?

routing/AsyncRouting.js

```javascript
const PrivateAsyncRoute = ({ load, authed, ...others }) => (
  <Route {...others} render={props => authed === true ? <Bundle load={load} {...props} /> :
  <Redirect to={{pathname: '/login', state: {from: props.location}}} />
  } />
);

const PublicAsyncRoute = ({ load, authed, ...others }) => (
  <Route {...others} render={props => authed === false ? <Bundle load={load} {...props} /> :
  <Redirect to='/dashboard' />
  } />
);
```

santaclauze:
@anuraaga  Thanks for the feedback, but I am not sure what you mean by  

> need to regenerate the loader files which also need to be updated

Are you talking about `webpack`?

Dattaya:
@anuraaga, before I review the new `ProgressBar`, let's discuss another issue that could potentially impact how it's implemented. Previously the transition was blocking: first, all of the resources were loaded and only then the view was updated. I've checked the 'offline' option in Chrome as it's easier to see this behavior, here I'm pressing the 'Features' button: 
rr3

Now it's not waiting:
rr4


I can't think of an elegant solution to this problem, I've come up with one that requires forking `react-router`'s `Switch` and that's not desirable. But I'm new to RR4, maybe you have a better idea and if not, what does everyone think if we ship 3.5 without this feature (and show `ProgressBar` instead?) ? Is it in a must have category?

anuraaga:
@Dattaya Thanks for the video, it really cleared up the progress bar semantics for me!

Yeah, I don't know if there's a good solution - I think React Router wants to be out of the business of dealing with asynchronous stuff ;) I think it's why in v4 react-router-redux renames locationBeforeTransitions back to location - the warning that the location may not actually reflect the rendered component is still there, but I guess it's a softer warning now.

My only idea is keeping track of the currently rendered component in the store and continue to display it instead of null when loading an async route - it would trigger a full react rerender, but the DOM shouldn't change, and React tree performance at route transition isn't a big deal IMO. However, I don't think this solution would work well with parent/child routes (though I wonder if the current progress bar too has issues with them).

I know it's very web-1.0, but what do you think of just displaying a loading spinner instead of null when loading the component? It's very simple and has the advantage of still working even after `npm run clean` - TBH, I had no idea about the existence of the progress bar until writing this PR and always wished for a spinner during route load rather than just leaving the page frozen, which is the default behavior after a clean.

Dattaya:
I have another idea that involves `context` and normally that's not recommended but as RR4 is heavily based on `context` and parent–child coupling, we can do that too. You can try it out here in this branch (forked from yours except the last commit, I'll get back to it later): https://github.com/Dattaya/react-boilerplate/tree/a-switch, commit: https://github.com/Dattaya/react-boilerplate/commit/4c4b590311c08d4db9f108ab467fa3c27e3b11e5 

It's neither tested nor linted because I need to hear some feedback first. I tried to create an async child route and it worked as well, so it seems that `context` with the same fields in parent as in child component is not destroyed, it just becomes shadowed.

One downside: it's not compatible with regular `Route` but we can recommend people to use [`webpack`'s `eager` mode](https://webpack.js.org/guides/code-splitting-async/#import-mode) (chunk is not generated) for synchronous routes:
```js
export default (store, cb) => {
  import(/* webpackMode: "eager" */ './index')
    .then(cb)
    .catch(errorLoading);
};
```
Another one: `AsyncRoute` can't be used without `AsyncSwitch`. That's also probably fixable.

I wonder if there are more downsides.

anuraaga:
@Dattaya Yep, that concept seems similar to maintaining similar state as you put in the context in the redux store, I believe. I guess both approaches would work equally well. Seems like a reasonable implementation to me if we want to keep the update blocking, so if you'd like to clean-up the commit I can cherry-pick it in (or vice versa).

Dattaya:
So apparently in production (over https) service workers would cache all of the chunks on the first visit in the background. I now don't think it's that important to block transition or even show a progress bar (only if it takes too much time to load, i.e. the chunk is not in the cache for some reason). We just need to enable Application Cache by default again. It is deprecated as was stated in previous discussions (#791) but AC is supported in almost every browser (except IE9) unlike SW (only Chrome and Firefox). 
image

Also, I'm considering integrating a library instead of our own `AsyncRoute`, but need to hear what everyone else is thinking. Some interesting libraries:

- https://github.com/thejameskyle/react-loadable
- https://github.com/ctrlplusb/react-async-component
- https://github.com/zackify/react-router-async-routing
- https://www.npmjs.com/package/react-universal-component
- https://github.com/oneut/async-react-router

gihrig:
>Also, I'm considering integrating a library instead of our own AsyncRoute

I haven't looked deeply into any of these, but [redux-offline](https://github.com/jevakallio/redux-offline) looks promising.

santaclauze:
I have looked in many places but I fail to find how and where the context is passed to the AsyncRoute.js

--> https://github.com/anuraaga/react-boilerplate/blob/9f44f1b95ce584ccf84e3a92c88dac64cd2faf78/app/routing/AsyncRoute.js#L45

Can you please explain to me how this happens?

kevin940726:
@Dattaya 

Our team integrate with `react-loadable` and it works great! It even enable us to preload the component which is pretty cool, but the downside is that the boilerplate code blow up a little (A Loadable component to export and entry file to import for every container/component). Haven't experiment it with SSR since our project doesn't need to support it.

anuraaga:
Out of those options, `react-loadable` looks to have the cleanest API to me. `redux-offline` looks fun to look into at some point but is probably overkill for the react router version upgrade.

@santaclauze `AsyncRoute` defines `contextTypes` which instructs react to automatically make store available from the context. Any component within a redux application (e.g., under `Provider`) can access the store this way using the context, without requiring propagating `store` all the way down the hierarchy. Using context is considered to be a bit hacky, but it's pretty common for redux utilities to use it to avoid users from having to manually deal with propagating `store` - the main reason for doing so here is to make it easier to define child routes, which otherwise would also need to have the store propagated to them.

santaclauze:
@anuraaga Thanks for the precision and sorry for pushing for the answer. Still learning a lot here :) I was logging the wrong thing, but I get it now. 

Thanks for your help and work man!

Dattaya:
@gihrig, it certainly does (I wish it had support for Apollo.) From what I've read in the docs, `redux-offline` doesn't have any code splitting functionality and is designed to cache data received from an API server.

@kevin940726, thanks for sharing your experience with `react-loadable`! Unless I'm missing something, it shouldn't differ from what we have here in this PR. Right now every async component has a `loader` file; with `react-loadable` it'll look like this:
```js
const LoadableMyComponent = RbLoadable({
  loader: (store) => Promise.all([
      import('./reducer'),
      import('./sagas'),
      import('./index'),
    ])
    .then(([reducer, sagas, component]) => {
      injectReducer('home', reducer.default);
      injectSagas(sagas.default);
      return component;
    }),
  LoadingComponent,
});
export default LoadableMyComponent;
```

@anuraaga, agreed.

Dattaya:
I had to wrap everything in a component, this is how `Loader.js` might look like if we integrate `react-loadable`:

```js
/**
 * Asynchronously loads the components for HomePage
 */
import React from 'react';
import { errorLoading } from 'utils/asyncInjectors';

import RbLoadable from 'routing/RbLoadable';

export default () => (
  <RbLoadable
    loader={({ injectReducer, injectSagas }) => Promise.all([
      import('./reducer'),
      import('./sagas'),
      import('./index'),
    ])
    .then(([reducer, sagas, component]) => {
      injectReducer('home', reducer.default);
      injectSagas(sagas.default);
      return component.default;
    })
    .catch(errorLoading)}
    LoadingComponent = {() => null}
  />
);
```

fooddilsn:
Hi guys,

I'm starting with a new project in these days and I'm really interested to use your react-boilerplate with react router v4.
When you think this PR will be stable and ready to merge?

Thanks for all, great work!

MilanRgm:
I am not a expert guy so i don't know if its a bug or i am doing the wrong way. The bug that is happening in my project is 

There is two entities. One is enduser and another is admin one. That is why i have following routes 

```javascript
function Routes({ location, ...props }) {
  return (
    <Switch location={location}>
      <AsyncRoute exact path="/" load={loadHomePage} />
      <AsyncRoute path="/signup" load={loadSignupPage} />
      <AsyncRoute path="/login" load={loadLoginPage} />
      <AsyncRoute path="/admin/dashboard" load={loadAdminDashboardParent} />
      <AsyncRoute path="" load={loadNotFoundPage} />
    </Switch>
  );
}
```
The parent component for enduser is homepage where signup and login component is. They are separate component. They have their own reducer, sagas, actions and contstants. When i click on signup, i get to signup component where i can fill the form. After hitting submit button, two time the request will be send. 

Second is the admin part where `loadAdminDashboardParent` is a parent for admin user. It has now its own routes as following 

```javascript
function AdminRoutes({ location }) {
  return (
    <div className="row">
      <div className="col-lg-12">
        <Switch location={location}>
          {/* <AsyncRoute path="/admin/dashboard" load={loadAdminDashboardPage} /> */}
          <AsyncRoute
            exact
            path="/admin/dashboard/commission"
            load={loadCommissionPage}
          />
          <AsyncRoute exact path="*" load={loadNotFoundPage} />
        </Switch>
      </div>
    </div>
  );
}
```

here when i route to /admin/dashboard i am directed to admin dashboard but is shown Not Found Page. If i click on commission link i am shown commission page on admin dashboard which is parent component for them but here two times api is called even though the sagas, reducers, api are defined in the component itself. 

@anuraaga Can you please look at this issue? If i need to share repo, I am ready but you can see always multiple times saga, reducers are called. Also /admin/dashboard is also shown as Not Found Page.

kevin940726:
@Dattaya 
In your example above, as far as my understanding, `react-loadable` requires `loader` to be a function with no parameter. If that's true, we'll have to implement some kind of higher-order component to wrap it inside and pass the store to it. And that's exactly what we had done, so far so good, but that seems a little hacky,  to me at least. I'm curious about how you're going to do that, I bet I'm gonna learn a lot from it!

Dattaya:
> I'm curious about how you're going to do that

I think there will be no difference. You can find the current edition (still a draft) here: https://gist.github.com/Dattaya/7de3c9dc68c5ecfa22c24b0bc305b337 It's quite simple.

I don't consider it to be a hack, otherwise we can call `react-loadable` a hack since it is a function that returns a component as well.

Dattaya:
So my preference is to use `react-loadable` instead of our own `AsyncRoute` because of 

- Rich functionality with even some SSR support in case we need it later
- It doesn't conflate routes with async components which means that a component can be loaded asynchronously as a result of user interaction without changing the url.

But one opinion is not enough to decide, @anuraaga, @justingreenberg, @KarandikarMihir, @gihrig, @samit4me, and every one else can you tell what you prefer?

MilanRgm:
@Dattaya Can you see my issue a bit above? I don't know if its a bug or i am handling routes a wrong way.

Dattaya:
@MilanRgm, I'm sorry, I don't know what exactly is going on, maybe the author of `AsyncRoute` @anuraaga can help?

justingreenberg:
based on feedback in #1776, we will be moving forward with this PR 🎉 

i agree with @anuraaga and @Dattaya that `react-loadable` compliments RRv4 very nicely

just reviewed https://gist.github.com/Dattaya/7de3c9dc68c5ecfa22c24b0bc305b337 — looks awesome! my only concern is using context.... maybe we could preload `asyncInjectors` with curried store during setup? i have been underwater but will try to find a few hours over the weekend to pull this down and properly review

thank you so much for the hard work that went into this, i'm very excited about shipping!

Dattaya:
> maybe we could preload asyncInjectors with curried store during setup?

That means that we treat a store as a singletone. I do that in my project, but are there any projects that use more than one store? If there are, this approach might not work for them.

anuraaga:
@justingreenberg @Dattaya I also lean towards sticking with context - it's a pretty common pattern for components that need direct access to the redux store and does give the flexibility of multiple stores as needed.

If the approach in RbLoadable looks good, would be happy to integrate it, and update all the templates and tests - and we'd go ahead and remove the progress bar?

Incidentally, using react-loadable will also have the same effect as https://github.com/react-boilerplate/react-boilerplate/pull/1734 memoizing the component across reloads. There seem to be some concerns on the long-term design of the sagas injection API, but is there a blocking issue with doing the memoization? I would like to eventually move sagas injection/unjection out of the loader and into `componentWillMount / WillUnmount` but feel it can come at a later stage.

Dattaya:
> I also lean towards sticking with context - it's a pretty common pattern for components that need direct access to the redux store and does give the flexibility of multiple stores as needed.

For those who don't yet know, RR4 has [a far worse problem/gotcha](https://reacttraining.com/react-router/web/guides/dealing-with-update-blocking), they rely on context and `location` can get stale unlike redux's store that shouldn't unless someone intentionally regenerates the store. `react-redux`'s `connect` also receives the `store` from the context: https://github.com/reactjs/react-redux/blob/master/src/components/connectAdvanced.js#L82

> If the approach in RbLoadable looks good, would be happy to integrate it

I plan on sending a PR to your fork, don't know when that happens, I'm also short on time. We can keep `ProgressBar` if we want, that's easy to implement but also involves context.

> is there a blocking issue with doing the memoization

don't think it's going to be a problem

>  sagas injection/unjection out of the loader and into componentWillMount / WillUnmount

Let's discuss it in another issue/pull request we'll need to investigate and discuss this idea thoroughly.

fooddilsn:
I also had a similar issue of @MilanRgm in another project where I'm using `react-boilerplate`.
I fixed multiple injection of the same sagas by editing *injectAsyncSagas* method in `app/utils/asyncInjectors.js` as follow:

```
export function injectAsyncSagas(store, isValid) {
  const asyncSagas = {};

  return function injectSagas(name, sagas) {
    if (!isValid) checkStore(store);

    invariant(
      Array.isArray(sagas),
      '(app/utils...) injectAsyncSagas: Expected `sagas` to be an array of generator functions'
    );

    warning(
      !isEmpty(sagas),
      '(app/utils...) injectAsyncSagas: Received an empty `sagas` array'
    );

    if (!asyncSagas[name]) {
      asyncSagas[name] = [];
    }

    sagas.filter((saga) => {
      if (asyncSagas[name].includes(saga.name)) {
        return false;
      }

      asyncSagas[name].push(saga.name);
      return true;
    }).map(store.runSaga);
  };
}
```

and then use it in `routes.js`:

```
injectSagas('uniqueName', sagas.default);
```

In this way, every saga is registered in an object under a unique name (that usually refers to the container that use it), and it is not added again if already exists, similar to what happens with async reducers injection. So it works without having to worry about adding and removing sagas each time, after their first inclusion.

Maybe this is not the right place for discussing it, but it make sense for you?

Tushant:
@anuraaga Can you once check if sagas are injecting multiple times or not in child routes? If that's the case then @fooddilsn solution plus `store.asyncSagas = {}` has to be added on store.js.

anuraaga:
I believe @dattaya's change to use react-loadable will automatically accomplish the same memorization of sagas so want to check with it first

Dattaya:
> We can keep ProgressBar if we want, that's easy to implement but also involves context.

The implementation is still simple but not as elegant as I hoped because I'd forgotten about async components nested inside other async components. So maybe we should stick with spinners in place of loading components instead of `ProgressBar` at the top. Before I continue, I'd like to hear what others think, especially @anuraaga and collaborators.

New `ProgressBar` can handle components loaded in parallel and nested in one another (those child components that are loaded immediately). Here's a quick walkthrough of how it works: each `LoadingComponent` registers itself with [registerLoadingComponent](https://github.com/Dattaya/react-boilerplate/commit/d442afdff8472874198d78c17b5237a53a254d3f#diff-2db869526bdbdfda72fae18fc1b26f97R56) if it's the first component, `ProgressBar` [starts animation](https://github.com/Dattaya/react-boilerplate/commit/d442afdff8472874198d78c17b5237a53a254d3f#diff-2db869526bdbdfda72fae18fc1b26f97R58) and when the loading is done, `LoadingComponent` [unregisters](https://github.com/Dattaya/react-boilerplate/commit/d442afdff8472874198d78c17b5237a53a254d3f#diff-2db869526bdbdfda72fae18fc1b26f97R65). If it's the last component it doesn't finish animation immediately, it waits another [50ms](https://github.com/Dattaya/react-boilerplate/commit/d442afdff8472874198d78c17b5237a53a254d3f#diff-2db869526bdbdfda72fae18fc1b26f97R77) (should be 300) and if [no async children were discovered](https://github.com/Dattaya/react-boilerplate/commit/d442afdff8472874198d78c17b5237a53a254d3f#diff-2db869526bdbdfda72fae18fc1b26f97R70), it updates `ProgressBar` to 100%. On each route change it [cleans the state](https://github.com/Dattaya/react-boilerplate/commit/d442afdff8472874198d78c17b5237a53a254d3f#diff-2db869526bdbdfda72fae18fc1b26f97R30). And that's it.

As it is described in `react-loadable` docs, by default `LoadingComponent` renders after 200ms delay meaning we will see a progress bar only if it was not previously pre-loaded by Service Workers/App Cache or loaded withing 200ms time frame which is fairly quickly. It's a preferred behavior as that blinking on each route would be annoying.

An example of one async withing another. `HomePage` takes 1s to load and a `Test` child component takes 2s to load:

```js
import Loadable from 'routing/Loadable';

export default Loadable({
  loader: () => new Promise((resolve) => {
    setTimeout(() => resolve(import('./index')), 2000);
  }),
});
```

The `ProgressBar` component waits until both are loaded:
progress-bar

My new favorite is [`LoadingIndicator` as a default `LoadingComponent`](https://github.com/Dattaya/react-boilerplate/commit/b3cd888ebff6158a69a02d4c29a01a3a38ba1edd). Nested async components:
spinner

Which one do you prefer?

In addition in both of the branches (both are WIP) I removed `routes` and moved them to `App.js`, `routes.js` gives the impression that we have a central routing config file and that's not true anymore. And removed `routing` generator, adding new routes is extremely simple now: `<Route exact path="/features" component={FeaturePage} />`.

DJTB:
I've been following this quietly and just wanted to vote for the progress bar. I often barely notice github's bar - making everything *feel* snappy somehow as it swings from left to right - as opposed to watching interminable spinner/loading components.

Hard to tell from your gif with a single nested async component, but is there the potential for an app that has 4 separate async components polluting the screen with 4 spinners (vs. a single progressbar) on a slow connection?

vicbell:
`ProgressBar` is better.

Dattaya:
@DJTB, @vicbell thanks for your feedback, much appreciated. I went ahead with loading indicator even before reading your comments, here's a PR: https://github.com/anuraaga/react-boilerplate/pull/2. That component (`LoadingIndication`) will be only available in the demo app, everyone can implement whatever they prefer in their projects. A default loading component can be set with the help of `routing/DefaultLoadingComponentProvider`. 

> for an app that has 4 separate async components

It would probably be better to show one or two for components that carry the most amount of content.

But I'm fine with `ProgressBar` as well and can continue working on it, but first we'll need to wait for other collaborators' decisions. As I said before, my opinion is not enough.

DJTB:
@Dattaya Thanks for all the great work. I'm currently using your rbp-object fork, and looking forward to updating the routing as well :smile:
> It would probably be better to show one or two for components that carry the most amount of content.

Completely agree. I was just curious (for argument's sake) if that was a potential result with a really granular app structure compared to using progressbar approach.

anuraaga:
Thanks @Dattaya! I agree with going with the simpler implementation, since now that it's very easy to preload, hopefully users will generally not see any spinner or progress bar at all :)

Dattaya:
> and looking forward to updating the routing as well

you can count on that, @DJTB 😃 

@anuraaga is going to merge https://github.com/anuraaga/react-boilerplate/pull/2, would be great if someone can check if that double saga injection problem is still present and if it is, I'd love to see a repro case.

anuraaga:
@Dattaya Merged into this PR - thanks a lot!

anuraaga:
While playing with the new loadable version, I found that hot reload (which I enabled using react-hot-loader in the project) stopped working again, likely due to the component memoization. Oh well...

smashercosmo:
FYI, @anuraaga new version of react-loadable has Loadable.Map helper for loading additional resources (like sagas and reducers)

DJTB:
@anuraaga Extra FYI: `LoadingComponent` option prop in `routing/Loadable.js` needs to be changed to `loading` if you upgrade react-loadable to ^4.

Dattaya:
Thanks, @smashercosmo and @DJTB 

> Loadable.Map helper for loading additional resources 

We can use this API, but nothing much is going to change, we'll still need a HoC to get a store.

Either one level up:

```js
WrapperOverLoadableMap({
  loader: {
    Component: () => import('./my-component'),
    sagas: () => import('./sagas'),
    reducer: () => import('./reducer'),
  },
  render({ injectReducer, injectSagas }, { Component, sagas, reducer }, props) {
    // alternatively pass injectors via props
    injectReducer('home', reducer.default);
    injectSagas(sagas.default);
    return <Component {...props}/>;
  }
});
```

or one level down:

```js
Loadable.Map({
  loader: {
    ComponentWrappedInHocWithStore: () => import('./my-component-with-store-and-injectors'),
    sagas: () => import('./sagas'),
    reducer: () => import('./reducer'),
  },
  render({ ComponentWrappedInHocWithStore, sagas, reducer }, props) {    
    return <ComponentWrappedInHocWithStore reducers={{ home: reducer }} sagas={sagas} {...props} />;
  }
});
```

Currently it's:
```js
export default Loadable({
  loader: ({ injectReducer, injectSagas }) =>
    Promise.all([
      import('./reducer'),
      import('./sagas'),
      import('./index'),
    ])
    .then(([reducer, sagas, component]) => {
      injectReducer('home', reducer.default);
      injectSagas(sagas.default);
      return component;
    }),
});
```

Dattaya:
> While playing with the new loadable version, I found that hot reload (which I enabled using react-hot-loader in the project) stopped working again, likely due to the component memoization

@anuraaga, can you publish it on a separate branch? I'd like to try to find out the reason.

Dattaya:
This PR does not solve the problem with sagas, just like before, when components (now any loadable component, before — route components) are being remounted, they inject sagas again. For example, that happens during locale change as was pointed out by @JK0N in #1434.
I want to refactor it to create a list with active sagas, in `componentWillUnmount` remove a saga from the list and in `componentWillMount` run it only if it's not present in the list. @justingreenberg, do you have a better idea?

Dattaya:
I need some advice from an expert on how to tackle the problem I described above and a well known saga expert here is @justingreenberg :). 

What do you prefer:  [a solution that checks if a task is still running](https://github.com/react-boilerplate/react-boilerplate/issues/1434#issuecomment-310783513) + remove task descriptor in `componentWillUnmount` (if it was cancelled) so that the reference can be garbage collected. 
Or cancel a task in `componentWillUnmount` and handle the cancellation appropriately in the saga:

```js
export function* githubData() {
    ...
    if (yield cancelled()) {
      yield cancel(watcher);
    }
```

Or maybe none of those in case I'm missing some edge cases?

vicbell:
Any saga expert could take a look at this? it is delay a lot .

Tushant:
I solved the multiple injection problem the following way

```javascript
import { errorLoading, getAsyncInjectors } from "utils/asyncInjectors";

export default (store, cb) => {
  const { injectReducer, injectSagas } = getAsyncInjectors(store);
  const importModules = Promise.all([
    import("./reducer"),
    import("./sagas"),
    import("./index")
  ]);

  importModules.then(([reducer, sagas, component]) => {
    injectReducer("uniqueReducerName", reducer.default);
    injectSagas("uniqueSagaName", sagas.default);

    cb(component);
  });

};

```

anuraaga:
As far as I understand, this PR does not have any regression with regards to sagas - they are only injected once, and just as the current version of the boilerplate, they are reinjected on every location change which is why we always use this pattern

https://github.com/react-boilerplate/react-boilerplate/blob/master/app/containers/HomePage/sagas.js#L40

So I think sagas can be addressed in another PR - for now I guess we're mainly waiting on a review of the PR, possibly from @justingreenberg ?

justingreenberg:
@Dattaya thank you for the kind words :) @vicbell i apologize for the delay, my company was undergoing penetration testing and compliance audit this week... tedious infosec stuff. anyway, i understand the outstanding issue with respect to duplication and will review as soon as possible... i have a few simple ideas that i think could solve most of this confusion and that work well with RRv4's component based approach, but i really need a solid 3-4 hours to dedicate for research. i promise this is priority for me and appreciate your patience! 🙏 based on what i've reviewed so far,  @anuraaga and @Dattaya have done some really awesome work, and thank you to all the community testers helping to move this into the end zone

side note, i have previously opined on the  "saga duplication issue" (simple `map` implementation in demo app) here: https://github.com/react-boilerplate/react-boilerplate/pull/744#issuecomment-233738166 and while some things have changed, i think most of what i outlined there still stands. the relevant portion is...

> ``` js
>  // sagas.js
>  export default {
>    run: [], // run on every route load, self-managing, advanced usage
>    watch: [], // async, fork and automagically cancel on location change
>    once: [] // fork and run one time
>  }
>  ```
>   
>   this approach could ie delegate watcher management to injector util 
> 
> i think there is definitely an opportunity to provide some "plumbing" for event channels that manage internals such as background tasks, network io, toast notifications, etc. that are tertiary to application logic

i think this may be a good opportunity to handle route::saga lifecycle bindings in a more durable way

Dattaya:
> they are only injected once

@anuraaga, that exact problem I described (saga duplication on locale change) does not happen here because on full remount `react-router-redux` dispatches `LOCATION_CHANGE`. I don't know what is it exactly, a feature or a bug? But if someone wants to show a loadable component conditionally then hide it and then show again (hence the component is mounted two times) the saga would be injected twice. We need to address it in this PR.  

I'm going to refactor it with a decorator like in @justingreenberg's message. In `componentWillUnmount` we can cancel a saga and even replace a reducer with identity function for much better performance (in a project with heavy and numerous reducers), but that will be a subject of another discussion and PR.

Dattaya:
> import sagas from './sagas'
> import { withSagas } from 'utils/withSagas'
> 
> export class HomePage extends Component {

@justingreenberg, is it OK that we import sagas alongside a component and not load these resources in parallel: `import('./index') ... import('./sagas')` ?

Dattaya:
> i think this is too ambitious at this point

well, we need to somehow address this bug (dev branch):
sagas-bug

relying on a quirk in the new `react-router-redux` alpha is probably not a good idea.

anuraaga:
@Dattaya Thanks for the explanation and screencast, yeah I see that this needs to be addressed now to handle locale change. Adding an HOC for managing sagas injection sounds great.

Also FYI I pushed a small fix to InnerLoadable in https://github.com/react-boilerplate/react-boilerplate/pull/1746/commits/35d775b2dfffbf2d7b7d8a3e281a34227394831a to fix an issue with matched routes I found.

justingreenberg:
>  is it OK that we import sagas alongside a component and not load these resources in parallel: import('./index') ... import('./sagas') ?

@Dattaya bundling sagas with RRv4 route components should be ok—the the "route component" with saga import is still "above" the split point in component tree so webpack static analysis will traverse module imports merge and bundle... v3's may even hoist sagas (and reducers) into component module scope, keen to see how that works out! 

in terms of the i18n bug, it really is an edge case owing to the dynamic and flexible nature of RRv4. without a centralized route configuration, we are pretty limited. one solution is to just create a simple in-memory cache so requests don't fire twice... would just be a super simple change to `utils/request.js`:

```js
// utils/request.js

const cache = new Map();

export function setCache(url) {
  return function setter(json) {
    cache.set(url, json);
    return json;
  }
}

export default function request(url, options) {
  if (cache.has(url)) {
    return cache.get(url);
  }

  return fetch(url, options)
    .then(checkStatus)
    .then(parseJSON)
    .then(setCache(url));
}
```

this way, although route components mount/unmount and fire lifecycle actions, we at least short circut re-rendering the `repoList` or other sub-routes when changing locale... lmk what you think!

if this doesn't feel right, i think we can probably just create a separate issue kick and this one down the road so we can :shipit:

Dattaya:
> webpack static analysis will traverse module imports merge and bundle

You're right, thanks for reminding me about that.

> in terms of the i18n bug, it really is an edge case owing to the dynamic and flexible nature of RRv4

This bug is present in `dev` and `master` and with RR4 it actually does not happen because of that `react-router-redux` quirk I mentioned.

vicbell:
Any news? 
When shoud this be ok? 
Tks for your great work!

Dattaya:
@vicbell, yep, here's a refactor with explanations about why and how: https://github.com/anuraaga/react-boilerplate/pull/4 Now I'm waiting for @justingreenberg's response.

justingreenberg:
@vicbell — @Dattaya has put some epic work into this new saga implementation... we are all very excited to ship this! we're almost there :)

app/containers/App/index.js
```diff
@@ -24,24 +26,26 @@ const AppWrapper = styled.div`
...
+    return (
+      <AppWrapper>
+        <Helmet
+          titleTemplate="%s - React.js Boilerplate"
+          defaultTitle="React.js Boilerplate"
+        >
+          <meta name="description" content="A React.js Boilerplate application" />
+        </Helmet>
+        <Header />
+        <Switch>
```
anuraaga: I think we're almost ready here!

@Dattaya Want to confirm one point - in my previous version, I had made `Routes` a connected component and passed the store's `location` in here as a prop. I was having trouble with reliable route-rerendering without making sure the `Switch` uses our redux state rather than its own. After the change, all rerendering issues I found went away - FWIW the issues would only show up with a production build on a real server, not local with webpack dev server.

Do you have any specific reason not to propagate `location` into here from the store? If not I can make the change to restore the connection.

Dattaya: That's probably because you use `connect` with your `App` container and bump into 'update blocking' issue. In this repro case: https://codesandbox.io/s/G6nWE3X0r try to press on some links and then press a back button, you'll see this bug, now remove `connect(mapStateToProps)(` from `App.js` and it will disappear. The example app does not suffer from it because we don't need to get anything from the store in `App`. I'd rather prefer to improve the docs and recommend to use [`ConnectedSwitch`](https://github.com/ReactTraining/react-router/issues/5072#issuecomment-310184271) instead of regular `Switch` while `react-router-redux` is still in beta.

Dattaya:
For everyone's information: `react-router-redux` v5 is in beta and does not support time traveling for a `LOCATION_CHANGE` action.

app/containers/App/index.js
anuraaga: Got it - yeah I agree recommending `ConnectedSwitch` is a better idea.

app/utils/injectReducer.js
```diff
@@ -0,0 +1,45 @@
...
+    componentWillMount() {
+      const { injectReducer } = this.injectors;
+
+      injectReducer(key, reducer);
+    }
+
+    componentWillUnmount() {
+      const { ejectReducer } = this.injectors;
+
+      ejectReducer(key, mode);
```
vicbell: If we eject the reducer, should we alse delete the data in the redux state?
If we do not delete the data from the redux state, should it cause a memery leak?

Dattaya: Normally, in a default mode, it will not be ejected, only if someone decides to use `RESTART_ON_REMOUNT` mode. And yes, `redux` will delete the data from the state.

aga5tya: State is just an object reference and will represent the application at any point of time. So state shouldn't be touched, whereas reducers can be since they are pure and acting upon passed state. Sagas also exhibit similar properties.

vicbell: @aga5tya Tks!
@Dattaya  `And yes, redux will delete the data from the state.` Is that say redux has a `"Gabage Collection"` same polices and we need not delete the page legacy data manually?

Dattaya: `redux.replaceReducer` either adds or removes a key to/from the state. The fact that the part of the state will be removed is actually a downside, it makes it just slightly more useful than React's state—an ability to query the state in sagas. I might as well remove the `RESTART_ON_REMOUNT` mode completely.

vicbell: When/How to clean out unused data from redux store?

Dattaya: That's for you to decide, not some helper like `injectReducer`; you can clean some data on `LOCATION_CHANGE` for example.

vicbell: `LOCATION_CHANGE` will triggered before `componentWillUnmount`, is that right?

Dattaya: it doesn't matter, your reducer will be functional after `componentWillUnmout`

vicbell: When use mode `RESTART_ON_REMOUNT`, assume `LOCATION_CHANGE` is triggered after `componentWillUnmount`, `ejectReducer` will done in `componentWillUnmout` , when `LOCATION_CHANGE` comes, there is no reducer to clean out the data, am i right?

vicbell: Got it, Tks! ^_^

mxstbr:
Deploy preview failed.

Built with commit cfa9de354254279ba9bb0605708614c1cf6b04a6

https://app.netlify.com/sites/react-boilerplate/deploys/595f15478ebdd90770773aee

Dattaya:
@mxstbr, welcome back! Will you be able to review it? 

> Deploy preview failed.

Are there any details? All I can see is:
image

vicbell:
@anuraaga `Deploy preview failed`. Maybe you could get the details?
@ALLEXPERT I am sorry to ask again: `Any news`?

app/utils/sagaInjectors.js
```diff
@@ -0,0 +1,68 @@
...
+
+    invariant(
+      isString(key) && !isEmpty(key),
+      '(app/utils...) injectSaga: Expected `key` to be a non empty string'
+    );
+
+    checkMode(mode);
+
+    if (mode !== DAEMON && Reflect.has(store.injectedSagas, key)) {
+      store.injectedSagas[key].task.cancel();
```
vicbell: Maybe we should also do `Reflect.deleteProperty(store.injectedSagas, key);` here?

app/utils/injectSaga.js
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @param {string} key A key of the saga
+ * @param {function} saga A root saga that will be injected
+ * @param {string} [mode] By default (constants.RESTART_ON_REMOUNT) the saga will be started on component mount and
+ * cancelled with `task.cancel()` on component un-mount for improved performance. Another two options:
+ *   - constants.DAEMON—starts the saga on component mount and never cancels it or starts again,
+ *   - constants.ONCE_TILL_UNMOUNT—behaves like 'RESTART_ON_REMOUNT' but never runs it again.
+ *
+ */
+export default ({ key, saga, mode }) => (WrappedComponent) => {
```
vicbell: In most of the usecase we have more than one sagas, so param should be an array of `{ key, saga, mode }`, eg:  `[{key1, saga1, mode1}, {key2, saga2, mode2}]`

app/utils/injectReducer.js
```diff
@@ -0,0 +1,45 @@
...
+ * Dynamically injects a reducer
+ *
+ * @param {string} key A key of the reducer
+ * @param {function} reducer A reducer that will be injected
+ * @param {string} [mode] By default (constants.DAEMON) the reducer is injected on component mount and is never removed.
+ * Another option is constants.RESTART_ON_REMOUNT—the reducer is started when component mounts and is removed on
+ * component un-mount for improved performance.
+ *
+ */
+export default ({ key, reducer, mode }) => (WrappedComponent) => {
```
vicbell: Sometimes we have more than one reducers, so param should be an array of `{ key, reducer, mode }`, eg: `[{key1, reducer1, mode1}, {key2, reducer2, mode2}]`

app/utils/injectSaga.js
Dattaya: You have three options:
1. Start other sagas in a root saga in whatever way you see fit:

```js
export default function* rootSaga() {
    helloSaga(),
    watchIncrementAsync()
}
```
```js
yield all([
  takeLatest(LOAD_REPOS, getRepos),
  takeLatest(LOAD_USERS, getUsers),
]);
```

2. Decorate your component with `injectSaga` multiple times:

```js
export default compose(
  injectSaga({ key: 'home', homeSaga });
  injectSaga({ key: 'test', testSaga });
)(MyComponent);
```

3. Create your own decorator and use helpers from `utils/sagaInjectors`:

```js
injectors = getInjectors(this.context.store);
componentWillMount() {
  const { injectSaga } = this.injectors;
  sagaDescriptors.map(({ key, saga, mode }) => injectSaga(key, saga, this.props, mode));
}
```

The same is true for reducers, except step 1, to combine reducers into one you can use [`reduceReducers`](https://github.com/acdlite/reduce-reducers) or something like that.
If you have a convincing argument why these three options is not enough, we can change the API.

app/utils/sagaInjectors.js
Dattaya: I plan on doing that but not in the `dev` mode, in `dev` I need access to the latest saga generator.

anuraaga:
I wasn't able to view the netlify error either, but it was probably just flaky. Forced CI and it's passing now

vicbell:
@justingreenberg Can you do some review?

vicbell:
@Dattaya @anuraaga @justingreenberg  Any news?

justingreenberg:
@react-boilerplate/core @anuraaga were there any additional changes? this LGTM after @Dattaya latest PR so if you guys are ok with this, lets merge into dev. want to mAke sure core team has a chance to review and give the 👍

anuraaga:
Ready to merge from my end

Dattaya:
I'm going to push few more improvements and then we'll need two approvals from collaborators [according to the rules](https://github.com/react-boilerplate/react-boilerplate/blob/dev/CONTRIBUTING.md#collaborating-guidelines). I can't give it a thumbs-up it because I'm a co-author.

anuraaga:
@react-boilerplate/core @justingreenberg @Dattaya Merged in the cleanup PR, should be ready for a final review.

justingreenberg:
@anuraaga @Dattaya thank you again guys—I'm traveling today for work but will complete review first thing tomorrow morning 

@KarandikarMihir would have time to review today or tomorrow so we can merge this?

justingreenberg:
this has been a literally been a global effort—on behalf of maintainers and users of RBP i'd like to thank everyone again for their time and contributions!

i'm pulling the trigger and merge into `dev` now :)

anuraaga:
Thanks @justingreenberg this is great! And thanks @Dattaya for basically rewriting my PR to make it much better :)

jstnn:
i thank you all for this effort. kudos

donszek:
@justingreenberg @anuraaga Hello, thanks for this wonderful feature. 

I created an issue at #1899 

Basically I tried to export array of generator function which is like below. However it does not work anymore like the usual case in 3.4. May I know is this supposed to be like this? How should I then export multipler saga-watchers? 

Cheers

``` jsx
export function* watch1() {
  yield takeLatest(CONSTANT1, DO_1);
}
export function* watch2() {
  yield takeLatest(CONSTANT2,DO_2);
}

export default [
  watch1,
  watch2,
];
```

vicbell:
@donszek 
I think you should take a look at [here](https://github.com/react-boilerplate/react-boilerplate/pull/1746#pullrequestreview-48775604), it help me a lot

donszek:
@vicbell that is sure very helpful. thanks for the reference. appreciate it.

williamgranli:
I'm looking for help regarding the workflow of this project. I'm looking to start up a new react project and I've previously used this boilerplate.

I'd like to get react-router v4 changes into my new project and I'm wondering what the best approach would be. Is the dev branch ready to go and relatively safe to use? I can see that it's a couple of commits behind master. 

Or should I simply go for master and fetch the v4 updates as soon as they're merged with master on here.

Thanks!

anuraaga:
I have been using dev in a couple of new projects and haven't had any issues - it fixes many issues compared to master related to saga lifecycle so I'd personally recommend forking off of dev.

The model for dev and master are quite different right now due to this router work, so if you were to go with master now the upgrade later will be a good amount of work.

shic:
@blling Maybe this part of code should be like this 

```
export default compose(
  injectSaga({ key: 'home', saga: homeSaga });
  injectSaga({ key: 'test', saga: testSaga });
)(MyComponent);

```
instead of

```
export default compose(
  injectSaga({ key: 'home', homeSaga });
  injectSaga({ key: 'test', testSaga });
)(MyComponent);
```

Otherwise it will not work.

You inspired me of including different sagas in the same container. But after experimenting, found that the key should always be `saga`.

blling:
@shic Yes, you are right, `saga` is the param name of `injectSaga`, should always be `saga`. If your saga has a different name, you must pass the params like `injectSaga({ key: 'yoursagakey', saga: yoursaganame });`

PR #1765, opened 2017-05-21
use single quotes

jackson-pm:
This PR makes the quote style consistent in this page of documentation.


gihrig:
@jackson-pm 

Good catch!

But...

>  double-check your branch is based on dev and targets dev

Thanks for contributing ✨

justingreenberg:
@jackson-pm good catch, thank you very much!

attempted to update base to `dev`, closing due to conflicts

would you mind creating new PR against `dev`?

PR #1775, opened 2017-05-25
feat(core): styled-components v2 update

justingreenberg:
updates project to use to styled-components v2 🎉 

i'm thinking we should probably update deps and cut a release soon...

edit: congrats on the release @mxstbr!! https://www.styled-components.com/

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,9 @@
   },
   "pre-commit": "lint:staged",
   "babel": {
+    "plugins": [
+      "styled-components"
+    ],
     "presets": [
       [
         "latest",
@@ -236,7 +239,7 @@
     "redux-saga": "0.14.3",
     "reselect": "2.5.4",
     "sanitize.css": "4.1.0",
-    "styled-components": "1.4.3",
+    "styled-components": "2.0.0",
     "warning": "3.0.0",
     "whatwg-fetch": "2.0.2"
   },
@@ -248,6 +251,7 @@
     "babel-plugin-dynamic-import-node": "1.0.0",
     "babel-plugin-react-intl": "2.3.1",
     "babel-plugin-react-transform": "2.0.2",
+    "babel-plugin-styled-components": "^1.1.4",
     "babel-plugin-transform-es2015-modules-commonjs": "6.22.0",
     "babel-plugin-transform-react-constant-elements": "6.22.0",
     "babel-plugin-transform-react-inline-elements": "6.22.0",
--- a/package.json
+++ b/package.json
@@ -251,7 +251,7 @@
     "babel-plugin-dynamic-import-node": "1.0.0",
     "babel-plugin-react-intl": "2.3.1",
     "babel-plugin-react-transform": "2.0.2",
-    "babel-plugin-styled-components": "^1.1.4",
+    "babel-plugin-styled-components": "1.1.4",
     "babel-plugin-transform-es2015-modules-commonjs": "6.22.0",
     "babel-plugin-transform-react-constant-elements": "6.22.0",
     "babel-plugin-transform-react-inline-elements": "6.22.0",
--- a/app/components/Header/A.js
+++ b/app/components/Header/A.js
@@ -1,8 +1,6 @@
-import styled from 'styled-components';
-
 import NormalA from 'components/A';
 
-const A = styled(NormalA)`
+const A = NormalA.extend`
   padding: 2em 0;
 `;
 
--- a/app/containers/HomePage/CenteredSection.js
+++ b/app/containers/HomePage/CenteredSection.js
@@ -1,8 +1,6 @@
-import styled from 'styled-components';
-
 import Section from './Section';
 
-const CenteredSection = styled(Section)`
+const CenteredSection = Section.extend`
   text-align: center;
 `;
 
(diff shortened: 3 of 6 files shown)


KarandikarMihir:
@justingreenberg Just a small query. Is `babel-plugin-styled-components` necessary? Because it says-

> This plugin is only necessary if you're server-side rendering, you can use styled-components perfectly fine without it!

And we're not supporting ssr atm.

Dattaya:
@KarandikarMihir, it doesn't look useful at the moment but it could be in the future once minification is no longer an experimental feature, so I'm going to give this a thumbs up, but if this plugin is removed I'm fine with that too.

package.json
```diff
@@ -248,6 +251,7 @@
     "babel-plugin-dynamic-import-node": "1.0.0",
     "babel-plugin-react-intl": "2.3.1",
     "babel-plugin-react-transform": "2.0.2",
+    "babel-plugin-styled-components": "^1.1.4",
```
Dattaya: Previous convention was not to use semver symbols like caret (^) in `package.json`

Dattaya:
@KarandikarMihir, I've read the description of `babel-plugin-styled-components` again and it says that it's fine to enable minification in production as long as it's enabled in development as well, so I think this plugin is useful after all.

Dattaya:
@justingreenberg, I have one nitpick, otherwise, nicely done :+1: 
I've tested your branch and everything seems to be fine (to be more precise, spotted some bugs but they're unrelated to this PR)

justingreenberg:
@KarandikarMihir & @Dattaya thank you guys!

Re: the plugin - I added it primarily to generate `displayName` (for use with devtools):

>This babel plugin adds the components' name to the class name attached to the DOM node. In your browser's DevTools you'll see `<button class="sc-Button-asdf123 asdf123" />` instead of just `<button class="asdf123" />`.
>
>This also adds support for showing your components' real name in the React DevTools. They will normally show `<styled.button>` for all of your components, but with this plugin they show `<MyButton />`.
>
>This makes it easier to find your components and to figure out where they live in your app. 

ref: https://github.com/styled-components/babel-plugin-styled-components#better-debugging

preprocessing will be a very cool feature, but I would wait until it stabilizes before turning on

KarandikarMihir:
@justingreenberg @Dattaya Alright! Thanks for clearing that up for me. Let's :ship: :smile:

somghosh:
Hello guys, I am trying to use styled-components v2 in the react boilerplate 3.4 (why? because of the better debugging in inspect tools - https://github.com/styled-components/babel-plugin-styled-components#better-debugging) ... I have done all the changes as indicated here in this PR - however it seems that I am getting an error

    index.js:7 Uncaught TypeError: 
    WEBPACK_IMPORTED_MODULE_0_styled_components.default.a.withConfig is not a function
    at Object../app/components/A/index.js (index.js:7)

which is the following line - 
    const A = styled.a

 seems like the styled function is not working any more. This issue goes away when I remove the babel plugin, but then I end up with cryptic class names that dont help debugging

Dattaya:
@somghosh, what is the result of:

```bash
> npm ls styled-components
```

also, that line looks wrong, it should be 
```js
const A = styled.a`
```

somghosh:
Hi @Dattaya 

`> npm ls styled-components` 

gives

```
react-boilerplate@3.4.0 /Users/somghosh/Programming/xeniajs
├── styled-components@2.1.0
└── UNMET PEER DEPENDENCY webpack@2.2.0-rc.3

npm ERR! peer dep missing: webpack@^1.11.0 || ^2.1.0-beta || ^2.1.0, required by eslint-import-resolver-webpack@0.8.0
npm ERR! peer dep missing: webpack@1 || ^2.1.0-beta, required by html-webpack-plugin@2.24.1
```

And yes,

`const A = styled.a``

This is exactly what I am doing, sorry for the typo

Dattaya:
@somghosh, I can see that you have outdated webpack 2 RC. In this case I recommend you to use the `dev` branch, it's the same as `master` but with updated dependencies.

somghosh:
Thanks, yes I figured out thats the case as well. I updated to webpack webpack@2.6.1 -- and what is now happening is the styled-component is adding a 'sc-' prefixed classname, but the remaining part of that classname is still another random text

Using the dev branch might introduce more uncertainties maybe - I simply want nice debuggable classNames in the Developer Console

somghosh:
Can I kindly ask @Dattaya if, when you made this pull request (and was merged), you were seeing "nice" classnames being produced by styled component? 

As I said, I upgraded to webpack 2.6.1 and what I am seeing is as below


image


image

mxstbr:
Can you run `cat node_modules/styled-components/package.json | grep version` ? If you use the Babel plugin with any `2.x` version (apart from some pre-releases) it should work.

somghosh:
"version": "2.1.0",

also `"babel-plugin-styled-components": "^1.1.5",`

somghosh:
thanks @mxstbr , I think it is working in terms of adding the new className, just not a "nice" classname

somghosh:
Apologies @mxstbr @Dattaya  - this works now, I seem to have forgotten to put in the 
```
"plugins": [
 +      "styled-components"
 +    ],

```
bit after making several changes. 

I also quickly checked that this is not in the `dev` branch code in package.json - maybe we should put that in to get the most out of styled-components

Dattaya:
@somghosh I'm glad to hear you figured it all out 👍 

> I also quickly checked that this is not in the dev

I think it is: https://github.com/react-boilerplate/react-boilerplate/blob/dev/package.json#L52

Issue #1774, opened 2017-05-25
Request utils fails on 204 HTTP response


williamdclt:
When using the `request` utils to call an API, this API may return a 204 HTTP response (success, no content).

When this is the case, the `request.parseJSON()` function fails because it is trying to parse an empty string to JSON. This error is unexpected, 204 is a success error code, relatively commonly used, and should be gracefully handled.

I'd expect `parseJSON()` to return `null` when parsing a 204. I'd be pleased to submit a PR if it seems like a valid concern to you!

![image](https://cloud.githubusercontent.com/assets/15254996/26450380/6a33cab6-4157-11e7-8cca-f274150651ff.png)

justingreenberg:
thank you @williamdclt—a PR would be much appreciated!

PR #1780, opened 2017-05-27
Request utils: handle 204 and 205 HTTP response

williamdclt:
As mentionned in [this issue](https://github.com/react-boilerplate/react-boilerplate/issues/1774), the `request` utils fails when it receive a 204 or 205 HTTP response, as it is trying to parse an empty string (the body) into JSON.

This pull request allows `request` to handle those response by returning a `null`.

### DIFF
--- a/app/utils/request.js
+++ b/app/utils/request.js
@@ -8,6 +8,9 @@ import 'whatwg-fetch';
  * @return {object}          The parsed JSON from the request
  */
 function parseJSON(response) {
+  if (response.status === 204 || response.status === 205) {
+    return null;
+  }
   return response.json();
 }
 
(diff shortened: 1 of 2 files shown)


Dattaya:
Thank you, @williamdclt! 
Would it be better to handle all of the json errors: return null or an empty object and print an error message in the dev mode, something like this:

```js
  response.json().catch((error) => { 
    if(process.env.NODE_ENV !== 'production') { 
      console.error(error.message);      
    }
    return null;
})
```

What do you think @williamdclt, @KarandikarMihir?

Dattaya:
Though it shouldn't show any errors in the console for 204 status code so I'm going to approve this PR.

PR #1782, opened 2017-05-28
Fix: Use local instance of shelljs

### LINKED ISSUE #1781: Use local instance of shelljs vs shelljs/global in scripts/clean.js
[scripts/clean.js](https://github.com/react-boilerplate/react-boilerplate/blob/master/internals/scripts/clean.js) uses `shelljs/global` which is a [non-recommended](https://github.com/shelljs/shelljs#global-vs-local) way. Apparently it pollutes the global namespace.

- Was there any specific reason to use the global shelljs?
- If not, can we switch to local instance of shelljs?

KarandikarMihir:
Nothing crazy. Just fixes #1781

### DIFF
--- a/internals/scripts/clean.js
+++ b/internals/scripts/clean.js
@@ -1,58 +1,58 @@
-require('shelljs/global');
+var shell = require('shelljs');
 const addCheckMark = require('./helpers/checkmark.js');
 
-if (!which('git')) {
-  echo('Sorry, this script requires git');
-  exit(1);
+if (!shell.which('git')) {
+  shell.echo('Sorry, this script requires git');
+  shell.exit(1);
 }
 
-if (!test('-e', 'internals/templates')) {
-  echo('The example is deleted already.');
-  exit(1);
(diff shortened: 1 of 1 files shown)


internals/scripts/clean.js
```diff
@@ -1,58 +1,58 @@
-require('shelljs/global');
+var shell = require('shelljs');
```
Dattaya: nit: you can use `const` instead of `var` the same way as `addCheckMark` declared

Dattaya:
I don't like that global require too. Usually in most of programming languages global imports are discouraged. 👍 from me.

KarandikarMihir:
@Dattaya I feel stupid! :sweat_smile: Should be good now.

Dattaya:
💯 Thanks, going to merge as soon as AppVeyor completes the build. 😃

Dattaya:
Appveyor's stuck and somehow I don't have admin access, going to close and reopen to bring it to life

PR #1823, opened 2017-06-24
Bump webpack to 3.0.0 and update related deps

KarandikarMihir:
**Note:** Work in progress :construction: 

Addressing #1815

Quoting @TheLarkInn here-

> Migrating from webpack 2 to 3, should involve no effort beyond running the upgrade commands in your terminal. We marked this as a Major change because of internal breaking changes that could affect some plugins.

For webpack 3 overview, head over to this awesome [medium article](https://medium.com/webpack/webpack-3-official-release-15fd2dd8f07b).

**UPDATE**
Minor reduction in main bundle size after using `ModuleConcatenationPlugin()`- 693kb -> 684kb

**Known Issue**
Looks like `html-webpack-plugin` is incompatible with webpack 3. So expect this warning while installing dependencies.
```
warning "html-webpack-plugin@2.28.0" has incorrect peer dependency "webpack@1 || ^2 || ^2.1.0-beta || ^2.2.0-rc".
```


KarandikarMihir:
@gihrig @Dattaya I came across [this comment](https://github.com/facebookincubator/create-react-app/pull/2574#issuecomment-310285905) after adding `ModuleConcatenationPlugin`. We can add that plugin later when things are stable.

Dattaya:
@KarandikarMihir, agreed. 
If it's not the only bug it would be better to wait until webpack 3 is more stable before mergin.

gihrig:
@KarandikarMihir Yes, I agree as well, stay with webpack 2 until 3 and related packages are stable.

In the case of webpack 1 -> 2 there were features we wanted to use so we went with beta versions. Is there any such use case for webpack 3?

Finally webpack 3 has many [breaking changes](https://github.com/webpack/webpack/releases), updating will require very thorough testing by several people.

KarandikarMihir:
@Dattaya @gihrig Thank you for your inputs. Let's sit on this for a while, then.

pgdejardin:
I've updated my project generated with react-boilerplate to webpack 3.
The main problem was with `html-webpack-plugin` and `extract-text-webpack-plugin`

I had to upgrade them to : 
```
        "extract-text-webpack-plugin": "3.0.0-beta.3",
        "html-webpack-plugin": "2.29.0",
        "webpack": "3.0.0",
```

justingreenberg:
it may be time to take another look at this... things seem to have stabilized, what do you guys think? would be great to include with next release

Dattaya:
> Looks like html-webpack-plugin is incompatible

It supports webpack v3 since 2.29

KarandikarMihir:
@pgdejardin Awesome! @justingreenberg @Dattaya Sorry for the late reply. I'm looking into this as I type.

Dattaya:
Unfortunately hot reloading with Webpack 3.x is still broken (for example, try to modify `HomePage`) [at first I thought it's webpack's fault](https://github.com/react-boilerplate/react-boilerplate/pull/1871#issuecomment-317249006) but now I can't reproduce the bug in a basic hot reloading demo. I'll need to investigate it further.

Dattaya:
Do we even need `ModuleConcatenationPlugin` in development? For the time being we can move it to the production config (main chunk will be reduced down from 668kB to 655kB) and that'll resolve the problem.

Dattaya:
> I can't reproduce 

btw, I did manage to reproduce the bug: https://github.com/webpack/webpack/issues/5510; it's now marked as 'urgent'. Also I've noticed in `3.5.4` release notes this: 'HMR: disable Scope Hoisting for modules using HMR' and we hot reload almost everything. So I still think it's a good idea to move this plugin to the production config and then we can merge this PR.

justingreenberg:
@Dattaya 👏 awesome!!

vicbell:
3.5.5 Released!

Dattaya:
whoa, @ sokra is so fast, amazing work.

KarandikarMihir:
@Dattaya Do we still need to move `ModuleConcatenationPlugin`?

Dattaya:
In my experience incremental builds became around 400ms slower and it's still possible to stumble on edge cases; I experienced one weird bug during hot reloading and had to restart the server to get rid of it, now I don’t remember how to reproduce the bug. So my preference is to use `ModuleConcatenationPlugin` only in production. But if you, or @justingreenberg, or @gihrig want to keep it, I'm fine with that too.

KarandikarMihir:
@justingreenberg @gihrig I'm not sure what exactly `ModuleConcatenationPlugin` does.. Could you shed some light and help close this one?

DJTB:
@KarandikarMihir 
> Scope Hoisting is the flagship feature of webpack 3. One of webpack’s trade-offs when bundling was that each module in your bundle would be wrapped in individual function closures. These wrapper functions made it slower for your JavaScript to execute in the browser. In comparison, tools like Closure Compiler and RollupJS ‘hoist’ or concatenate the scope of all your modules into one closure and allow for your code to have a faster execution time in the browser.

> You can now add the following plugin to your configuration to enable scope hoisting:
```
module.exports = {  
  plugins: [
    new webpack.optimize.ModuleConcatenationPlugin()
  ]
};
```
> Scope Hoisting is specifically a feature made possible by ECMAScript Module syntax. Because of this webpack may fallback to normal bundling based on what kind of modules you are using, and other conditions.
> To stay informed on what triggers these fallbacks, we’ve added a --display-optimization-bailout cli flag that will tell you what caused the fallback.

> Because scope hoisting will remove function wrappers around your modules, you may see some small size improvements. However, the significant improvement will be how fast the JavaScript loads in the browser. If you have awesome before and after comparisons, feel free to respond with some data as we’d be honored to share it!

KarandikarMihir:
@DJTB Thanks! :sweat_smile: I know it's used for scope hoisting.. I was actually referring to-

> In my experience incremental builds became around 400ms slower

Sorry if that was confusing. Did you experience anything similar? We're not using webpack 3 yet in our project, so I don't have any hands-on experience.

@Dattaya Let's keep `ModuleConcatenationPlugin` in both dev and production. That way more people will use it and chances of finding that weird bug will be more. Fixing it is just one line change..

DJTB:
@KarandikarMihir I'm only using it in production, and I haven't benchmarked that since I don't expect it to be speedy anyway :turtle: 

I assume in dev it slows down the incremental builds simply since there's more work for webpack to do - and the benefits of the plugin aren't really necessary there either since the browser code execution is slowed down by prop checks and redux devtools etc regardless

KarandikarMihir:
@DJTB Aah.. Okay. Then let's remove it from dev config without further delay, since we have 2 votes.

PR #1839, opened 2017-06-30
Add babel-polyfill entry to webpack prod config.

SoccerGee:
When doing a production build babel-polyfill will be loaded on entry.  A problem occurs when performing a production build (specifically `npm run start:production` or `npm run build`).  This fixes an error caused by the lack of `Symbol()` support in IE11.

Edit:

Additionally, Babel lists this as a requirement in their documentation.  See [here](https://babeljs.io/docs/usage/polyfill/#usage-in-node-browserify-webpack).

### DIFF
--- a/internals/webpack/webpack.prod.babel.js
+++ b/internals/webpack/webpack.prod.babel.js
@@ -7,6 +7,7 @@ const OfflinePlugin = require('offline-plugin');
 module.exports = require('./webpack.base.babel')({
   // In production, we skip all hot-reloading stuff
   entry: [
+    'babel-polyfill',
     path.join(process.cwd(), 'app/app.js'),
   ],
 


KarandikarMihir:
@SoccerGee We're loading `babel-polyfill` at the very start of the [entry file](https://github.com/react-boilerplate/react-boilerplate/blob/master/app/app.js#L9). Correct me if I'm wrong, but the final effect looks the same to me.

SoccerGee:
@KarandikarMihir you're right but you have to add it to the entry array for production builds.  See the second part of the Usage section here: 

https://babeljs.io/docs/usage/polyfill/#usage-in-node-browserify-webpack

KarandikarMihir:
@SoccerGee Cool. Why is it needed only in production? Why not in base config?

SoccerGee:
@KarandikarMihir I think it's just not a necessary step a for development.  However taking advantage of the babel-polyfill in production environment is more advantageous.  

Also, the issue that I was experiencing only occurred in production after running `npm run start:production` or deploying after `npm run build`.

SoccerGee:
@KarandikarMihir just curious if you spin up a test project and run the project by using `npm run start:production` can you open it in IE 11 without errors?

I cannot.  This fixed that for me.

SoccerGee:
@KarandikarMihir is there anything else I can provide on this?

Dattaya:
@SoccerGee, thanks for contributing! 
`babel-polyfill` should be loaded either with `import ..` or as an entry point, not both.

> can you open it in IE 11 without errors

yes, current dev branch works without any errors in IE 10 and 11, but not 9 (Microsoft Windows [Version 10.0.15063]), started with `npm run start:production`. In IE9 I get: `Unhandled promise rejection TypeError: Unable to get property 'add' of undefined or null reference`

Dattaya:
Oh, actually I do exprerience it, sorry for the misinformation, I was on [another branch](https://github.com/react-boilerplate/react-boilerplate/pull/1871). We have a bug in `package.json`, @SoccerGee, apply these changes and the error will disappear:

```diff
diff --git a/package.json b/package.json
index 80bff27..32f3b12 100644
--- a/package.json
+++ b/package.json
@@ -56,9 +56,7 @@
       [
         "env",
         {
-          "es2015": {
-            "modules": false
-          }
+          "modules": false
         }
       ],
       "react",
```

Dattaya:
That bug was fixed in `dev`. @SoccerGee, please, let us know if you still experience it in IE.

PR #1849, opened 2017-07-09
Serve Dlls via add-asset-html-webpack-plugin

skidding:
This allows tools like React Cosmos–which use the existing webpack config to generate the build–to extend React Boilerplate easily. Right now the Dlls are served via Express middleware, which makes the Node server a dependency for the webpack configuration. 

Dev & prod work well on my end. Happy to test anything else. Thanks!


KarandikarMihir:
@skidding This is great! This was one of the concerns for me in my pet project. I'll have a look at the changes soon.

skidding:
Happy to hear! I want to feature an integration between React Boilerplate & React Cosmos and this was the biggest (or only) roadblock.

Dattaya:
@skidding, there is also [`autodll-webpack-plugin`](https://github.com/asfktz/autodll-webpack-plugin) that would allow us to remove almost all dll related code from RBP.  Is it compatible with React Cosmos?

skidding:
@Dattaya auto-dll-webpack-plugin would also be good. It might have more implications for RBP because of how dll are stored at the moment inside node_modules. I'd prefer it to be a subsequent PR.

@KarandikarMihir do you think you'll have time to look at this soon? The branch got out-of-date and I'd rather not fix the conflicts more than once. Thanks! 🍺

Dattaya:
@skidding, agreed, this is a good first step + `auto-dll-webpack-plugin` is still in the process of maturity, we can add it in 3.6.

DenisNeustroev:
Do we still need to use this function and the cheerio package? [webpack.dev.babel.js: L148](https://github.com/skidding/react-boilerplate/blob/cf262f9e8ff91ca772fbd0c9eb950576f674da10/internals/webpack/webpack.dev.babel.js#L148)

skidding:
@DenisNeustroev good catch! The dlls were added twice.

image

Removed the function along with cheerio, and brought the branch up to date.

KarandikarMihir:
@skidding Apologies for not looking at this. I updated the branch for you :) Now looking at your changes. This is one of the most needed feature.

DJTB:
I did run into a problem with this in my project (already using the webpack 3.0 PR #1823), though I have modified the RBP slightly a bit over time.

Basically, when transpiling to a `web` target some packages throw when trying to access node's `fs`.
The fix was to add the following to the end of the `internals/webpack/webpack.base.babel.js` config:
```
  // Prevent "can't resolve fs" with glob/fs.realpath errors when building dll
  node: {
    fs: 'empty',
  },
```

The clean RBP repo might not run into this issue, but leaving the info here just in case.

skidding:
> Thank you, @skidding ! 

Likewise! ❤️

Now that they're full compatible, I will feature React Boilerplate in the React Cosmos docs!

PR #1859, opened 2017-07-17
Fix plain CSS include

v1adko:
Hi everyone. First off - awesome boilerplate!

I've encountered a small issue with importing plain `.css`, getting  
*“You may need an appropriate loader to handle this file type” with Webpack and CSS* error  

Simply changing include to exclude in webpack config solved it for me.
Plus following [SAAS guide](https://github.com/react-boilerplate/react-boilerplate/blob/master/docs/css/sass.md) it uses exclude (and it's only logical to exclude node modules).

So it seems like a simple typo, that's why I opened a PR without an issue, hope it's ok.
Checked all 3 branches and open PRs, seems no one has addressed it before (guessing not many people still use plain CSS :)

### DIFF
--- a/internals/webpack/webpack.base.babel.js
+++ b/internals/webpack/webpack.base.babel.js
@@ -24,7 +24,7 @@ module.exports = (options) => ({
       // they will be a part of our compilation either way.
       // So, no need for ExtractTextPlugin here.
       test: /\.css$/,
-      include: /node_modules/,
+      exclude: /node_modules/,
       loaders: ['style-loader', 'css-loader'],
     }, {
       test: /\.(eot|svg|ttf|woff|woff2)$/,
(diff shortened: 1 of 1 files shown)


mxstbr:
Deploy preview failed.

Built with commit ec65b3f778670e3d7f75e2fc128c9b872fdf5bae

https://app.netlify.com/sites/react-boilerplate/deploys/596c9b25a700c476b7f82ae5

v1adko:
@mxstbr Oh, I have tried my code on not cleaned repo and look sanitize broke.
So doing include on node_modules breaks simple css import within component and excluding it breaks `sanitize` :)

v1adko:
Ok, figured it out...
```javascript
{
   ...
   include: [/node_modules/, /app/]
   ...
}
```
This works, so the question is should it be allowed or we're relying on `styled-components` doing their magic?)

Maybe add the section about plain CSS (just like SAAS) to the docs then?

PR #1871, opened 2017-07-21
Replace `preset-hmre` with vanilla webpack HMR

### LINKED ISSUE #1068: Switch to react-hot-loader v3

Dattaya:
This is an alternative to https://github.com/react-boilerplate/react-boilerplate/pull/1207.

Downsides:
- does not preserve component's state but Redux state will be in place;
- could be annoying to some devs that injected non-daemon sagas will be restarted on each change in components, though in my experience it's not a big deal.

Upsides:
- very simple setup;
- works with any kind of component unlike deprecated `react-hmre`;
- `react-hot-loader` is still fragile with a lot of edge cases, here we have none;
- injected sagas and reducers will be hot re-loadable for free (no additional setup necessary).

/cc @justingreenberg, @marnusw, @anuraaga

[justingreenberg] edit: resolves #1068

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -80,16 +80,10 @@ const render = (messages) => {
 };
 
 if (module.hot) {
-  // Hot reloadable React components
-  module.hot.accept('containers/App', () => {
-    ReactDOM.unmountComponentAtNode(MOUNT_NODE);
-    render(translationMessages);
-  });
-
-  // Hot reloadable translation json files
+  // Hot reloadable React components and translation json files
   // modules.hot.accept does not accept dynamic dependencies,
   // have to be constants at compile-time
-  module.hot.accept('./i18n', () => {
+  module.hot.accept(['./i18n', 'containers/App'], () => {
     ReactDOM.unmountComponentAtNode(MOUNT_NODE);
     render(translationMessages);
   });
--- a/app/app.js
+++ b/app/app.js
@@ -78,8 +78,13 @@ const render = (messages) => {
   );
 };
 
-// Hot reloadable translation json files
 if (module.hot) {
+  // Hot reloadable React components
+  module.hot.accept('containers/App', () => {
+    render(translationMessages);
+  });
+
+  // Hot reloadable translation json files
   // modules.hot.accept does not accept dynamic dependencies,
   // have to be constants at compile-time
   module.hot.accept('./i18n', () => {
--- a/app/app.js
+++ b/app/app.js
@@ -64,6 +64,7 @@ openSansObserver.load().then(() => {
 const initialState = {};
 const history = createHistory();
 const store = configureStore(initialState, history);
+const MOUNT_NODE = document.getElementById('app');
 
 const render = (messages) => {
   ReactDOM.render(
@@ -74,13 +75,14 @@ const render = (messages) => {
         </ConnectedRouter>
       </LanguageProvider>
     </Provider>,
-    document.getElementById('app')
+    MOUNT_NODE
   );
 };
 
 if (module.hot) {
   // Hot reloadable React components
   module.hot.accept('containers/App', () => {
+    ReactDOM.unmountComponentAtNode(MOUNT_NODE);
     render(translationMessages);
   });
 
@@ -88,6 +90,7 @@ if (module.hot) {
   // modules.hot.accept does not accept dynamic dependencies,
   // have to be constants at compile-time
   module.hot.accept('./i18n', () => {
+    ReactDOM.unmountComponentAtNode(MOUNT_NODE);
     render(translationMessages);
   });
 }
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -56,9 +56,7 @@
       [
         "env",
         {
-          "es2015": {
-            "modules": false
-          }
+          "modules": false
```
Dattaya: That was a bug in `package.json` that should be fixed before 3.5 release

justingreenberg:
@Dattaya looks good to me.... using vanilla webpack hmr is a solid, maintainable approach—as you mentioned, the lack of support for local component state should generally be ok for redux applications since most **data** and **core** application state *should be* lifted into global state... the vastly reduced complexity is worth it

imo, none of the downsides would be a deal breaker for this approach, but i would love feedback from other peeps from core

 - has this been tested with webpack v3 by chance?
 - maybe we should include something in docs with link to webpack hmr docs?

app/app.js
```diff
@@ -78,8 +78,13 @@ const render = (messages) => {
   );
```
justingreenberg: i know it's a bit outside the scope of this PR, but this has been bothering me for a while... i feel like explicitly unmounting component in hmr callback is more appropriate in this case, ie -

```js
 const MOUNT_NODE = document.getElementById('app')

// ...

 const render = (messages) => {
   ReactDOM.render(
     <Provider store={store}>
       <LanguageProvider messages={messages}>
         <ConnectedRouter history={history}>
           <App />
         </ConnectedRouter>
       </LanguageProvider>
     </Provider>,
     MOUNT_NODE
    )

// ...

  module.hot.accept('containers/App', () => {
    ReactDOM.unmountComponentAtNode(MOUNT_NODE)
    render(translationMessages);
  });
```

anuraaga:
LGTM - I had tried using react-hot-loader with `ReactLoadable` and couldn't get it to update the memoized components, but this approach is working great, very nice.

app/app.js
Dattaya: let's do this to be certain the components are unmounted correctly

Dattaya:
@anuraaga, thank you for trying it out.
@justingreenberg, thanks for renaming the title, fun fact: initially I wanted to name the branch `bare-hot` 😁.

> has this been tested with webpack v3 by chance

Since `ModuleConcatenationPlugin` is [incompatible with `DllPlugin`](https://github.com/webpack/webpack/issues/5095)) I had to try one or the other, so webpack v3 + `DllPlugin`—hot reloading works, but webpack v3 with `ModuleConcatenationPlugin` is a completely different story—`react-loadable` throws errors on hot reloading and shows a loading indicator instead of a component that was supposed to be loaded. For some reason imported `HomePage` no longer has `__esModule: true` flag and [`resolve`](https://github.com/thejameskyle/react-loadable/blob/master/src/index.js#L103) returns incorrect result. I think that is a bug in webpack.

image

> maybe we should include something in docs with link to webpack hmr docs?

Added a note with a link in introduction.

internals/webpack/webpack.dev.babel.js
```diff
@@ -29,7 +29,7 @@ module.exports = require('./webpack.base.babel')({
   // Add hot reloading in development
   entry: [
     'eventsource-polyfill', // Necessary for hot reloading with IE
-    'webpack-hot-middleware/client?reload=true',
+    'webpack-hot-middleware/client?overlay=true&reload=true',
```
Dattaya: no reason, I'm going to remove it

Dattaya:
Other collaborators seem to be busy, I'll just go ahead and :shipit:, we can always improve/revert/replace it later.

kaiwa:
Hey, are there any recently added prerequisites to get webpack hmr working, which are not directly included in this PR?

I have a project based on a an earlier verison of react-boilerplate and trying to apply the changes from this pr, but they break hot reloading and a page reload is triggered on any code change instead.

Dattaya:
@kaiwa, could be because RR3 (as well as 2) memoizes routing components, take a look at https://github.com/ReactTraining/react-router/issues/2182, maybe you'll find suitable solution there.

kaiwa:
@Dattaya just managed to upgrade react-router, it is working now with webpack hmr! thx

PR #1886, opened 2017-08-06
Fix bug in font-loader where woff2 is not matched

daveish:
Fixes a bug in the font-loader configuration where woff2 is never matched.


daveish:
Never mind, sorry.

Issue #1745, opened 2017-05-05
I've managed to use sagas but not in react-boilerplate


bernatfortet:
After hours of looking around I can't find out how to deal with redux-sagas inside of react-boilerplate. Here is what I have:
I have a component that when clicked triggers an action.

```js
// containers/someContainer/index.js
import { someAction } from './sagas'
...
  onClick = () => {
    this.props.someAction()
  }
...
const mapDispatchToProps = dispatch => bindActionCreators({ someAction }, dispatch)
export default connect(null, mapDispatchToProps)(someContainer)
```

```js
// container/someContainer/sagas.js
const SOME_ACTION = "SOME_ACTION"


export const someAction = (data) => {
  console.log(`someAction`, data ) // I get this log. 🎉
  return {
    type: SOME_ACTION,
    data: data
  }
}


function* mainSaga(){
  console.log(`mainSaga`) // I don't get this log 😕 
  yield call(someSaga, val )
}

function* someSaga( val ){
  while( true ){
    console.log(`someSaga`) // I don't get this log 😕 
    const action = yield take(SOME_ACTION)
    console.log(`maybe someSaga?`) // I don't get this log either 😕 
  }
}
```

```js
// routes.js
const importModules = Promise.all([
  import('containers/someContainer'),
  import('containers/someContainer/sagas'),
]);
```
What am I not getting about this flow?

alacroix:
@bernatfortet, did you inject your saga in `routes.js`?

```javascript
// routes.js
const importModules = Promise.all([
  import('containers/someContainer'),
  import('containers/someContainer/sagas'),
]);

const renderRoute = loadModule(cb);

importModules.then(([component, sagas]) => {
    injectSagas(sagas.default); // 👈 this line right here

    renderRoute(component);
});
```

PR #1913, opened 2017-08-29
React routerv4/ssr. Merged functionality of #1745 and #1236

ax-hd:
## React Boilerplate
Since there was a vote to merge either react router v4 or server side rendering, I was looking to implement both in this boilerplate. I have merged code from both the commits and got React router v4 working with Server side rendering, solving errors as I encountered them. I know there is bit of work to be done with my pull request. Can someone check this work and tell me what I need to do to get this merged. I have not written corresponding scripts and test (because I dont yet know how), I can learn and implement them. Please suggest other changes to be done. 
To run:
npm run ssr:clean
npm run ssr:generate
npm run start

Sometimes npm run start needs to be done twice. I need to fix it.


TheRusskiy:
@ax-hd this does not actually render components containers on server, I think it is because inside AsyncRoute containers are loaded like promises, and SSR doesn't really wait for these import promises to be resolved.
If you disable JS and load a page you are gonna see that imported components are not rendered

leifdalan:
@TheRusskiy Yes, I noticed that too. The header and footer render but the inner components don't, because they are using Async Component, which relies on lifecycle methods that react SSR does not execute. 

I read this article a while back, it may prove useful, particularly the piece where you essentially have different require statements for server/client for your component containers: https://github.com/react-boilerplate/react-boilerplate/pull/1913

TheRusskiy:
I used @tomazy 's ssr branch of this repo and for lazy component loading on server I used https://github.com/seeden/babel-plugin-dynamic-import-node-sync
but you would have to be careful to make sure that import is not used as a promise anywhere on the server

ax-hd:
OK, I will look into it. Thanks for reporting. @TheRusskiy

blling:
@ax-hd I think this PR is repeat of  #1236, should we close this one?

blling:
Closing as stale, more discussion please see #1236.

PR #1921, opened 2017-09-06
use injectReducer in <App /> and <LanguageProvider /> instead of app/…

kopax:
- Use `injectReducer` in `<App />` and `<LanguageProvider />` instead of injecting them in `app/reducers.js`

We don't have to worry about where is injected the global and language reducer. 

Both components won't be unmounted so I assume this is an improvement as we keep this way `app/reducer.js` free of any business logic.

### DIFF
--- a/app/containers/App/index.js
+++ b/app/containers/App/index.js
@@ -10,12 +10,17 @@ import React from 'react';
 import { Helmet } from 'react-helmet';
 import styled from 'styled-components';
 import { Switch, Route } from 'react-router-dom';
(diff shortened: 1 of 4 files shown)


KarandikarMihir:
I'm not a big fan of Higher order components because it hides away important things. I think the previous setup where reducers are explicitly added is a better approach. That way, developers will also understand that this is how you can add reducers synchronously (unlike `injectedReducers`). But that's just my opinion. Let's see what others have to say.

PR #1928, opened 2017-09-10
chore(repo): Backport `master` to `dev`

Dattaya:
@justingreenberg, @KarandikarMihir, I resolved the conflicts but don't know what's the best way to merge it, I can do that by activating "Allow merge commits" option temporarily, does that mean we'll have two merge commits? Or temporarily removing protection from `dev` and pushing commits directly.


justingreenberg:
@Dattaya thank you! not in front of computer now but can you add yourself to branch protection whitelist to push to dev?

Dattaya:
@justingreenberg, I don't think I can, but I'll figure something out, maybe unchecking 'Include administrators' for few minutes would help.

Dattaya:
So I pushed the commits directly and it marked both of our PR's as merged 😃, interesting. I'm sorry I didn't wait for approvals, now I understand that it would've allowed me to push without the need to uncheck that setting if I had an approval, that's the best way to do it.

PR #1939, opened 2017-09-15
feat(core): Set Referrer Policy to "origin" using document meta tag

JonathanStoye:
Add `<meta name=“referrer” content=“origin”>` to make the application more secure by not sending the full url to servers other than the origin.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -3,6 +3,9 @@
   <head>
     <!-- The first thing in any HTML file should be the charset -->
     <meta charset="utf-8">
+    
+    <!-- Send only the base url to third party servers -->
+    <meta name=“referrer” content=“origin”>
 
     <!-- Make the page mobile compatible -->
     <meta name="viewport" content="width=device-width, initial-scale=1">


KarandikarMihir:
Nice addition. For anyone who wants more info on this, I found a [wiki](https://wiki.whatwg.org/wiki/Meta_referrer).

JonathanStoye:
Weird is, that this tag seems to be still a draft even though all browsers support it. 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta

justingreenberg:
reference - https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-delivery-meta

oscar-broman:
The quote marks are wrong. This will set the values to `“referrer”` and `“origin”`.

justingreenberg:
@oscar-broman good catch... this is using left/right quotations

```
  "   U+0022 QUOTATION MARK  ** this is what we need to be using **
  “   U+201C LEFT DOUBLE QUOTATION MARK
  ”   U+201D RIGHT DOUBLE QUOTATION MARK
```

@JonathanStoye would you mind opening a new PR to fix this?

JonathanStoye:
@justingreenberg I will!

JonathanStoye:
@justingreenberg @oscar-broman https://github.com/react-boilerplate/react-boilerplate/pull/1976

PR #1949, opened 2017-09-26
fix(tests): Updates assertion grammar for Wrapper tests

mohamed-ismat:


app/components/IssueIcon/tests/index.test.js
```diff
@@ -4,7 +4,7 @@ import { shallow } from 'enzyme';
 import IssueIcon from '../index';
 
 describe('<IssueIcon />', () => {
-  it('should render a SVG', () => {
+  it('should render an SVG', () => {
```
KarandikarMihir: I'm not a native English speaker, but I think "an SVG" is correct, because it's pronounced as ES-V-G? :sweat_smile:

mohamed-ismat: yes it was "a SVG" and I changed it to "an SVG" :)

KarandikarMihir: Ah! Sorry, I got it the other way around.

PR #1954, opened 2017-09-27
Update immutablejs.md

mohamed-ismat:


justingreenberg:
@mohamed-ismat thank you!

PR #1963, opened 2017-10-04
feat(webpack): Add heartbeat option to `webpack-hot-middleware`

mikunpham:
Add `heartbeat` option in `webpack-hot-middleware` to remove error ERR_INCOMPLETE_CHUNKED_ENCODING

screen shot 2017-10-04 at 10 38 40 pm

### DIFF
--- a/server/middlewares/addDevMiddlewares.js
+++ b/server/middlewares/addDevMiddlewares.js
@@ -17,7 +17,9 @@ module.exports = function addDevMiddlewares(app, webpackConfig) {
   const middleware = createWebpackMiddleware(compiler, webpackConfig.output.publicPath);
 
   app.use(middleware);
-  app.use(webpackHotMiddleware(compiler));
+  app.use(webpackHotMiddleware(compiler, {
+    heartbeat: 2000,
+  }));
 
   // Since webpackDevMiddleware uses memory-fs internally to store build
   // artifacts, we use it instead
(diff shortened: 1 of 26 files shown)


server/middlewares/addDevMiddlewares.js
```diff
@@ -17,7 +17,9 @@ module.exports = function addDevMiddlewares(app, webpackConfig) {
   const middleware = createWebpackMiddleware(compiler, webpackConfig.output.publicPath);
 
   app.use(middleware);
-  app.use(webpackHotMiddleware(compiler));
+  app.use(webpackHotMiddleware(compiler, {
+    heartbeat: 2000,
```
justingreenberg: heartbeat should be "less than the client's timeout setting - usually set to half its value"
https://github.com/glenjamin/webpack-hot-middleware#middleware

since default timeout is 2s, let's reduce this to 1s **or** increase client timeout to 4s eg by updating query
```
webpack-hot-middleware/client?reload=true&timeout=4000
```

https://github.com/react-boilerplate/react-boilerplate/blob/dev/internals/webpack/webpack.dev.babel.js#L44

mikunpham: Yeah, I've tried the `timeout` option already, but somehow the error didn't appear anymore even when I haven't added it to `webpack-hot-middleware/client?reload=true`.
Anyways, I'm ok with set the client default timeout is 2s, and heartbeat is 1s 👍

mikunpham:
@justingreenberg I saw [this line](https://github.com/glenjamin/webpack-hot-middleware/blob/master/client.js#L6) in `webpack-hot-middleware`, that mean client timeout is not 2s by default, right?

justingreenberg:
@mikunpham yes you are correct :) will merge when build passes—thank you!

PR #1966, opened 2017-10-07
chore(internals): Upgrade enzyme to v3

### LINKED ISSUE #1950: Update to React v16
## Description

React 16 is officially released:

https://facebook.github.io/react/blog/2017/09/26/react-v16.0.html

It would be amazing if Boilerplate could use it by default.

timburgess:
Upgrade to Enzyme 3

This resolves what I believe to be the major blocker for React 16
- [x] Upgrades enzyme to 3.1.0 and the required enzyme adapter (React 15 for now)
- [x] Adapter setup for tests
- [x] Resolves some syntax issues for some tests
- [x] Deal with different `find` behaviour for new static renderer

Resolves #1950

### DIFF
--- a/package.json
+++ b/package.json
@@ -209,6 +209,7 @@
       ".*\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/internals/mocks/image.js"
     },
     "setupTestFrameworkScriptFile": "<rootDir>/internals/testing/test-bundler.js",
+    "setupFiles": ["<rootDir>/internals/testing/enzyme-setup.js"],
     "testRegex": "tests/.*\\.test\\.js$"
   },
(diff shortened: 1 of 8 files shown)


blling:
Could you update `enzyme` to the lastest version and use `enzyme-adapter-react-16` instead of `enzyme-adapter-react-15`  ?

package.json
```diff
@@ -264,7 +265,8 @@
     "circular-dependency-plugin": "3.0.0",
     "coveralls": "2.13.1",
     "css-loader": "0.28.4",
-    "enzyme": "2.8.2",
+    "enzyme": "3.1.0",
+    "enzyme-adapter-react-15": "1.0.1",
```
blling: Should use `enzyme-adapter-react-16` instead of `enzyme-adapter-react-15`

blling: En, I am sorry, Maybe should do this in a follow up PR.

PR #1976, opened 2017-10-10
fix: quotes in referrer policy

JonathanStoye:
Use the right quotes for the referrer policy.

Sorry for this very weird fault.


justingreenberg:
@JonathanStoye @jay-bricksoft thank you!

PR #1977, opened 2017-10-10
fix(docs): Update `route` prop description for Button component

theploki:
*In: app/components/Button/index.js*

Prop **route** was changed to **handleRoute** in Jan 19, 2016, but the commented instructions were not updated.


justingreenberg:
@theploki thank you!

Issue #1985, opened 2017-10-19
Add short_name to mainfest.json


uniquename:
Adding short_name to manifest.json would create a better lighthouse result.

justingreenberg:
resolved in #1988

PR #1988, opened 2017-10-20
feat(core): Add short_name to manifest.json

gretzky:
Addresses #1985 by adding a `short_name` field to the app manifest. 

When there's insufficient space to display an app's full name, it gets truncated. This happens with app launcher and new tab in Chrome for Android. The `short_name` field allows a 12 character or less abbreviation. It also addresses any problems in testing the PWA in Lighthouse.

### DIFF
--- a/app/manifest.json
+++ b/app/manifest.json
@@ -1,6 +1,6 @@
 {
   "name": "react boilerplate",
-  "short_name": "react bp", // 12 chars or less
+  "short_name": "react bp",
   "theme_color": "#b1624d",
   "background_color": "#fafafa",
   "display": "standalone",
--- a/app/manifest.json
+++ b/app/manifest.json
@@ -1,5 +1,6 @@
 {
   "name": "react boilerplate",
+  "short_name": "react bp", // 12 chars or less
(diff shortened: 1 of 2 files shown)


app/manifest.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "react boilerplate",
+  "short_name": "react bp", // 12 chars or less
```
justingreenberg: @gretzky thank you for this :)

one nit: would you mind removing inline comment (since not JSON spec compliant)

maybe we could explain this constraint in `general/gotchas.md`?
https://github.com/react-boilerplate/react-boilerplate/blob/dev/docs/general/gotchas.md

ie:

```
## Using short_name in Manifest.json (PWA)

When there's insufficient space to display an app's full name it is truncated.
This happens with app launcher and new tab in Chrome for Android.
The `short_name` field allows a *12 character or less abbreviation*.
It also addresses any problems in testing the PWA in Lighthouse.
```

gretzky: Done and done!

justingreenberg:
@gretzky thank you! :)

PR #2003, opened 2017-10-28
feat(Messages): Add scope to messages to simplify adding new messages

Droow:
I have changed the messages structure and their generators to use single reference to their scope. This way I can easily change their scope or add new messages with live templates.

For example in my IDE i have set this live template to add new messages which wouldn't be possible without this change: 
vyber_048

I hope this will save some time when changing/adding messages and it is a step forward to the DRY principle. 

Also, thanks @timburgess for this idea.

### DIFF
--- a/app/components/Footer/messages.js
+++ b/app/components/Footer/messages.js
@@ -5,13 +5,15 @@
  */
 import { defineMessages } from 'react-intl';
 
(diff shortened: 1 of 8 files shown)


blling:
This is done in `dev` branch.

Droow:
Sorry, my bad, forgot to check.

PR #2047, opened 2017-12-03
fix(.nginx.conf) moving root path to the main server block

Mensae:
With the addition of the 'Do not cache sw.js' block, the root path, that is in the 'Always serve index.html' block, should move to the main server block or the sw.js will not be found.


blling:
LGTM, Thanks.

blling:
@gretzky  Need your review :)

gretzky:
Thanks @blling - just gonna update the branch to latest and then it should be good to go

PR #2061, opened 2017-12-12
Update redux reducer doc noting unique key

danactive:
Multiple reducers can not be applied to store key


danactive:
Sure, I start this change now

Issue #1983, opened 2017-10-16
Boilerplate 3.5.0, Sagas executes Api several times


furedal:
## Description

Hey, i've based my application on this boilerplate but having issues where my Api-calls are executed multiple times (Usually around 4).
I upgraded from boilerplate 3.4.0 last week but had similar issues there.

Something thats even weirder, if i put a `console.log` in my saga before the Api call, the `console.log` will only be run once.

The project only has one saga file
Here is my index-file where i import the sagas and reducers:  https://pastebin.com/m2fWmLFt
My saga: https://pastebin.com/8j30HjM4

I am not sure how to debug this, if anyone have any directions?

# Versions

- React-Boilerplate 3.5.0
- Node/NPM: 6.10.0/3.10.10
- Browser: Chrome/Safari

sourcesoft:
Check the dispatched actions logs in redux chrome extension, to see if Saga is actually being called multiple times. If not I feel like `const api = new Api();` doesn't look cool outside of saga being sent through watcher instead.

sourcesoft:
Oh also check where you're dispatching `UPDATE_PERIOD`, maybe it's being called somewhere outside of componentDidMount that gets executed multiple times.

furedal:
@sourcesoft I used the chrome extension. It tells me that `UPLOAD_PERIOD` is called once while `UPLOAD_PERIOD_SUCCESS` is called 5 times.
`UPLOAD_PERIOD` is executed in an onClick method.

I will check `const api = new Api();` but I don't think that's the problem. I have an react-native application that uses a similar boilerplate where the API is declared and used the same way.

furedal:
I noticed that every time i make an update in my files so hot reload executes, the number of sagas executed each time gets larger. Probably that the injected sagas gets imported more than once for some reason?

furedal:
Ok, I have identified the problem.
My Reducer and Sagas are injected in multiple files. When checking the injectors, it looks like they are supposed to ignore a Saga/Reducer if the key already has been declared, but I might be wrong.

Anyway, I have wrapped my components into one component that will inject the global sagas/reducers before passing them to the their corresponding Route component.

PR #2063, opened 2017-12-13
fix(sagaInjectors) Prevent sagas to be injected multiple times

lvarayut:
This PR prevents the sagas to be injected if its key is duplicate with the key inside the `injectedSagas` object.

When a duplicate key passed to the `injectSaga()` function, the current saga in the `injectedSagas` object will be replaced with the newly created one. So, there is no way we can refer to the old saga and cancel it. Therefore, we need to be careful to not inject a saga that has a duplicate "key". 

Related to #1983 and #2061. 


mxstbr:
Deploy preview failed.

Built with commit a1eabe0b9842fe26f732a6cd0e9cd54e112e56b2

https://app.netlify.com/sites/react-boilerplate/deploys/5a315f45a1147764782632d0

app/utils/sagaInjectors.js
```diff
@@ -29,6 +29,15 @@ const checkDescriptor = (descriptor) => {
   );
 };
 
+const checkRunningSaga = (store, key) => {
+  const descriptor = store.injectedSagas[key];
+  const isRunning = descriptor && descriptor.task && descriptor.task.isRunning();
+  invariant(
+    !isRunning,
```
blling: If developers accidentally introduced the same key saga So the original saga will be covered, I think we should not be allowed to pass the same key saga

blling:
cc @Dattaya

app/utils/sagaInjectors.js
lvarayut: If the saga has already cancelled, the developers should be allowed to pass the same key to start the saga again. Therefore, `checkRunningSaga` should check only if the saga is duplicate and also running.

PR #1969, opened 2017-10-07
feat(core): Upgrade to React v16

timburgess:
## React Boilerplate

This PR extends from the Enzyme 3 upgrade PR.
- [x] Upgrades to React 16.0 and updates related dependencies
- [x] Upgrades Enzyme 3 adapter to 16 and adds shim required for React 16 testing
- [x] Adds ErrorBoundary component and uses in App

Build size goes from 690kB -> 630kB.

I'm not sure exactly if/why/how we should test ErrorBoundary.

Before opening a pull request, please ensure:

- [x] You have followed our [**contributing guidelines**](https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md)
- [x] double-check your branch is based on `dev` and targets `dev` 
- [x] Pull request has tests (we are going for 100% coverage!)
- [x] Code is well-commented, linted and follows project conventions
- [ ] Documentation is updated (if necessary)
- [x] Description explains the issue/use-case resolved and auto-closes related issues

Be kind to code reviewers, please try to keep pull requests as small and focused as possible :)

**IMPORTANT**: By submitting a patch, you agree to allow the project
owners to license your work under the terms of the [MIT License](https://github.com/react-boilerplate/react-boilerplate/blob/master/LICENSE.md).

PR #2071, opened 2017-12-16
feat(core): Upgrade to React16 - update dependences

blling:
Should update dependences first and then add errorboundary and log intercept as follow up PRs.
- [x] Update the dependences to React16 
- [x] Fix test warning: React depends on requestAnimationFrame.
- [x] All seems good in my local test.


blling:
cc @mxstbr @justingreenberg  @Dattaya @gretzky @gihrig

avdeev:
What about #1969?

blling:
@avdeev I think we should upgrade to react 16 first and new features will add in the follow PRs. Will close #1969 when all it`s determined features are done :)

BTW:
Is there core teamer could give some advise of new features ?

PR #2083, opened 2017-12-24
feat(internals): Add prettier integration

### LINKED ISSUE #1945: RFC: Prettier
i'm surprised nobody has brought this up yet... it's been a game changer for my team. would be nice to integrate with #1711 husky pre-commit hooks. imo this is a no-brainer, kinda wish it shipped with v3.5. maybe we can tighten up the releases and cut another minor in the next couple of weeks

blling:
JG: resolves #1945

This PR complete the following task:
- [x] Add prettier integration.

Prettier configs as follows:
```javascript
  prettier: {
    jsConfig: {
      patterns: ['**/*.js'],
      ignores: ['node_modules/**', 'build/**'],
      options: {
        trailingComma: 'es5',
        bracketSpacing: true,
        semi: true,
        singleQuote: true,
        jsxBracketSameLine: false,
        arrowParens: 'always',
      },
    },
  },
```
You can add other configs to the `prettier` config node if you like.

- [x] Run prettier on the entire codebase and fix problems after run prettier.

Two main changes:
1. Add `stylelint` rules:
```javascript
    "rules": {
      "rule-empty-line-before": null,
      "declaration-block-semicolon-newline-after": null
    }
```
2. Replace `eslint-disable-line` to `eslint-disable-next-line` in some special lines. 

- [x] Fix generate.

Add a custom plop actiontype `prettierjs` and use prettier API to do prettier after plop `add` actions complete. This is important to make CI pass.


gretzky:
Not sure why AppVeyor failed- can anyone help with that? I've never used it

package.json
```diff
@@ -34,18 +34,27 @@
...
-    "generate": "plop --plopfile internals/generators/index.js",
+    "generate": "plop --plopfile internals/generators/index.js && npm run prettier:js",
     "lint": "npm run lint:js && npm run lint:css",
     "lint:eslint": "eslint --ignore-path .gitignore --ignore-pattern internals/scripts",
     "lint:js": "npm run lint:eslint -- . ",
-    "lint:css": "stylelint \"./app/**/*.js\"",
+    "lint:css": "stylelint './app/**/*.js'",
     "lint:staged": "lint-staged",
     "pretest": "npm run test:clean && npm run lint",
+    "prettier:js": "prettier --write './{app, internals}/**/*.js'",
```
sourcesoft: How about all the folder except `build`

package.json
```diff
@@ -34,18 +34,27 @@
...
+    "trailingComma": "es5",
+    "bracketSpacing": true,
+    "semi": true,
+    "singleQuote": true,
+    "jsxBracketSameLine": false,
+    "arrowParens": "always"
+  },
   "lint-staged": {
     "*.js": [
       "lint:eslint",
```
sourcesoft: We can add prettier to pre-commit hook
```
    "**/*.js": [
      "prettier --write",
      "git add"
    ]
```

package.json
sourcesoft: we can even format `package.json` and markdowns

blling: Good idea.

package.json
blling: `git add` will add all unstaged files, if the workplace is dirty, then it will add all the dirty files?

blling:
I think we need to use `pretter api` for doing this.

package.json
sourcesoft: Yes it formats all the files and stage everything so not cool.

sourcesoft:
1. Use husky and remove `lint-staged` and `pre-commit` dependency and instead use `husky`.
2. Add a bash file like
```
touch .pre-commit.sh
chmod +x .pre-commit.sh
```
3. Add this to the file
```
git diff --staged --diff-filter=dx --name-only HEAD | grep ".*\.js$" | xargs -I % sh -c './node_modules/.bin/prettier --write %; git add %'
```
4. Add the bash file to husky pre-commit script
```
"precommit": "./.pre-commit.sh",
```
This only formats staged js files no matter where they are.

blling:
I think it is ready for review.

blling:
@sourcesoft thanks for your suggestions

blling:
cc @justingreenberg @gihrig i closed #1711 , could you do some review in this PR?

justingreenberg:
thank you so much @blling awesome work :) i had been traveling and then catching up on work, apologies and thank you for driving this project forward... very exciting! 

everything works, LGTM 👍

since this is a pretty big diff, would be good to get one more review before merging

app/utils/tests/sagaInjectors.test.js
```diff
@@ -103,7 +101,7 @@ describe('injectors', () => {
       process.env.NODE_ENV = originalNodeEnv;
     });
 
-    it('should not remove daemon saga\'s descriptor in production', () => {
+    it("should not remove daemon saga's descriptor in production", () => {
```
blling: This looks a little strange

justingreenberg: so prettier's default formatting behavior (both with and without --no-semi option) is to prefer double quotes instead of escape characters... it converts string literals to use double quotes so there's no need to escape the tick...
```js
it('should not remove daemon saga\'s descriptor in production', () => {})
```
becomes
```js
it("should not remove daemon saga's descriptor in production", () => {});
```
and 
```js
it('should not remove daemon saga descriptor in production', () => {});
```
stays the same since no escape char
```

afaik there's no way to change this

Comment (blling) [reply app/utils/tests/sagaInjectors.test.js]: Yes there is no way to change this. I the prettier [docs](https://prettier.io/docs/en/options.html#quotes) it gives the notes:
>If the number of quotes outweighs the other quote, the quote which is less used will be used to format the string - Example: "I'm double quoted" results in "I'm double quoted" and "This \"example\" is single quoted" results in 'This "example" is single quoted'.

@justingreenberg do you thinks this is a risk?

Comment (blling) [reply app/utils/tests/sagaInjectors.test.js]: BTW, I also found:
`it("should remove non daemon saga\"'s descriptor in production", () => {});`
becomes
`it('should remove non daemon saga"\'s descriptor in production', () => {});`

**Comment by blling** [comment]
cc @react-boilerplate/core @react-boilerplate/collaborators  I need one more review:)
@gihrig ?

**Comment by blling** [comment]
cc @gretzky

**Comment by blling** [comment]
cc @vigzmv

File: app/utils/tests/sagaInjectors.test.js (hunk 3 above)
Comment (justingreenberg) [reply app/utils/tests/sagaInjectors.test.js]: @blling totally cool with me—small price to pay imo :)

**Comment by azz** [comment]
I would recommend keeping the config simple.

- `Remove internals/scripts/prettier.js` and the `package.json` scripts
- Use the `plugin:prettier/recommended` ESLint config: https://github.com/prettier/eslint-plugin-prettier#recommended-configuration
  `$ yarn add --dev eslint-plugin-prettier eslint-config-prettier`
  ```json
  // .eslintrc.json
  "extends": ["plugin:prettier/recommended"],
  ```

**Comment by blling** [comment]
@azz Then when and where should i do `prettier write` ? Now i do it in the `precommit` hook.

**Comment by azz** [comment]
You could lint-staged + husky for that: https://prettier.io/docs/en/precommit#option-1-lint-staged-https-githubcom-okonet-lint-staged

**Comment by blling** [comment]
@azz Thanks!
I use `internals/scripts/prettier.js` for doing these:
1. List: all files need to do prettier
2. Prettier write: all files need to do prettier
3. Prettier write: all changed files ahead of `dev` branch which need to do prettier
4. Prettier write: all `git staged` files which need to do prettier in `precommit` stage
if use `plugin + lint-staged + husky` i do not know how to implement the thirth and fourth step.

BTW:
`react-boilerplate` has a template generator for code generate, it seems that i must use `prettier API` for prettier the generated js files, if not, the file will not pass `yarn run test`(ESLint check will fail).
One of the CI steps is here: https://github.com/react-boilerplate/react-boilerplate/blob/dev/appveyor.yml#L38-L41

**Comment by azz** [comment]
I don't think you need to worry about incremental linting until there are a very large amount of files.

**Comment by blling** [comment]
@azz Thanks again.
My previous vesion is use `plugin + lint-staged + husky` and i run `prettier` on the whole codebase every time the `generate` is complete, but the CI is always fail in these steps https://github.com/react-boilerplate/react-boilerplate/blob/dev/appveyor.yml#L38-L40, so i decided to change to use `Prettier API` totally.

The CI log is here, i do not known why prettier can not find files, maybe the root directory is missmatch.
https://ci.appveyor.com/project/justingreenberg/react-boilerplate/build/365/job/n3ia0uh73pfihtxb

So i think i must use `Prettier API` instead of use some `plugin` for now for these reasons:
1. Avoid unnecessary `prettier` for the whole codebase. 
2. Run `prettier` for the `git staged` files and do `git add -u` for the prettiered files on every precommit stage, for prevent add dirty files.
3. Can specify the base git branch name and find changed files for running prettier, but not all of the codebase files.

**Comment by azz** [comment]
```
azz at macOS in ~/code/react-boilerplate on master [!]
$ time prettier --write "**/*.js" &>/dev/null

real	0m2.922s
user	0m2.507s
sys	0m0.242s
```

It only takes 3 seconds to check all files, if you integrate it into ESLint you won't even notice.
`lint-staged` + `husky` will mean you won't have to write a custom script to check for changed files.

azz:
R.e. the build failure you mentioned, could you send a link to the log?

blling:
Here is the log https://ci.appveyor.com/project/justingreenberg/react-boilerplate/build/365/job/n3ia0uh73pfihtxb

azz:
The single quotes are probably the issue, I'd use double quotes to wrap the patterns, but I thought you said you tried with ESLint?

blling:
@azz You are right, thanks a lot.
I will close this PR in favor of  #2095.

blling:
cc @justingreenberg @azz waiting for your review again in #2095 :)

PR #2091, opened 2018-01-06
Remove react-intl default span tags.

uberfabulous:
This allows users to see that they can change the default behavior of
`<Formatted*>` tags when using react-intl. Default span tags are
insidious and adds unnecessary clutter.


justingreenberg:
>Default span tags are insidious and adds unnecessary clutter.

@uberfabulous amen dude.

app/containers/LanguageProvider/index.js
```diff
@@ -17,7 +17,7 @@ import { makeSelectLocale } from './selectors';
 export class LanguageProvider extends React.PureComponent { // eslint-disable-line react/prefer-stateless-function
   render() {
     return (
-      <IntlProvider locale={this.props.locale} key={this.props.locale} messages={this.props.messages[this.props.locale]}>
+      <IntlProvider locale={this.props.locale} key={this.props.locale} messages={this.props.messages[this.props.locale]} textComponent={React.Fragment}>
```
justingreenberg: cool trick, thanks for taking the time to contribute and share :)

please remember to update the corresponding template file...

https://github.com/react-boilerplate/react-boilerplate/blob/master/internals/templates/containers/LanguageProvider/index.js

justingreenberg:
@uberfabulous another thought... instead of wrapping in an h1 tag, can we pass raw text to language component fixture with strict comparison to unit test fragment?

://github.com/uberfabulous/react-boilerplate/blob/bdb30f2bb58c7abfe2a9c4db636680579153c1ad/app/containers/LanguageProvider/tests/index.test.js#L28

app/containers/LanguageProvider/index.js
uberfabulous: I've updated the internals LanguageProvider template file with the change and force pushed the rebased / amended commit to my repo (hoping no one is working off of this branch as of yet!).

While looking through the test however,  I felt like changing the <ConnectedLanguageProvider /> test to look for a string, is probably better than changing the <LanguageProvider /> test. Reason being <LanguageProvider /> is normally used by wrapping it around the entire App, so I would think that it should always expect ReactElements and not a string as it children. That said, when I went to go update the <ConnectedLanguageProvider> test, I went down a bunny hole where I think I found that the `store` variable isn't really being set by `configureStore`, and `browserHistory` was something that was used in React Router v3 and not recommended in v4? So, because its a little over my head for now (kind of still starting out with the React ecosystem, but learning fast because of wonderful folks like you!), I left the tests untouched. =(

Not sure if you have any recommendations / suggestions on how to go forward with this one. 

In any case, here's the trick I found to remove span tags. Thanks for looking into it!!

**A bit of warning however, I think React Fragments were only introduced / "stable" as of React v16.2.0.**

uberfabulous: Also, should `defineMessages()` be used to set `messages` in the tests? I thought that `defineMessages()` was mainly used to let `babel-plugin-react-intl` know one of the places scrape throughout the App for Intl Messages. Is `babel-plugin-react-intl` being invoked anywhere in the larger test process being ran by `npm test`?

internals/templates/containers/LanguageProvider/index.js
```diff
@@ -17,7 +17,7 @@ import { makeSelectLocale } from './selectors';
 export class LanguageProvider extends React.PureComponent { // eslint-disable-line react/prefer-stateless-function
   render() {
     return (
-      <IntlProvider locale={this.props.locale} key={this.props.locale} messages={this.props.messages[this.props.locale]}>
+      <IntlProvider locale={this.props.locale} key={this.props.locale} messages={this.props.messages[this.props.locale]} textComponent={React.Fragment}>
```
smashercosmo: What's the benefit of using React.Fragment instead of just 
```jsx
<IntlProvider textComponent={children => children} />
```

uberfabulous: Since we sometimes want `children` to only render strings, if you passed a string into `this.props.children`, like "Hi I'm a string" and then tried to render it as a React Component   object inside `textComponent={this.props.children}`, it would be similar to doing `<Hi I'm a string>Hi I'm a string</Hi I'm a string>`, in JSX. Correct me if I'm wrong, but I think that would always blow up. If you're passing an actual React Component into `this.props.children` within `textComponent={this.props.children}` on the other hand, I guess that would work(?), but then you can end up having some people stuck with inconsistent / multiple dynamic renderings of the parent elements for `<Formatted*>` tags, spread out between different files within a scaled out project. Or another way of looking at it, it obfuscates the parent elements rendered for the `<Formatted*>` tags making it harder for a new developer within a project to see what's going on without digging into other files. It also makes the developer have to do a few extra acrobatics to make a different implementation of the `<Formatted*>` tags they want, whereas wrapping it inside a `React.Fragment` would simply allow the developer to just implement `<Formatted*>` on the exact same file as the Component they're using it in, exactly the way that's intuitive to them with the rest of React. A `React.Fragment` would allow new developers to just ignore the details of the implementation of `<IntlProvider>` to get something simple they want done right away within a larger project. People will also get a feel of `React.Fragment`s and the wonderfulness in potential that they can provide as First-Class Component citizens! =) That's my very long answer for a short question. =D Cheers!

smashercosmo: Sorry, my code above is wrong. This is the right one:
```jsx
<IntlProvider textComponent={props => props.children} />
```
This should always work both with strings (since react 16 supports rendering strings) and components.

uberfabulous: One weird thing `textComponent={props => props.children}` does is that it creates a `<textComponent>` Element inside the React Component Tree (inspecting with React Dev Tools) but **not** the actual rendered DOM itself. Fragments do not have this extra element in the React Component Tree, but I honestly don't know which one is more efficient in terms of processing. React-Intl usage of Context gives me a headache. =D Trying to think of use-cases for hijacking `textComponent` functionality using `props.children` with my own Components inside the children of `<IntlProvider>`... If there is one, would it make more sense to just create a dumb Component within the LanguageProvider folder and throw that into `textComponent` instead? Sorry, more questions than answers.

danielnass:
Hi folks!

Now `<FormattedMessage />` accepts prop `tagName` passed as element or string.

https://github.com/yahoo/react-intl/pull/1118

Issue #2073, opened 2017-12-19
Add jsconfig.json to improve editor support


mjbvz:
## Issue Type

- [ ] Bug (https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md#bug-reports)
- [X] Feature (**https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md#feature-requests**)

## Description

[`jsconfig.json`](https://code.visualstudio.com/Docs/languages/javascript#_javascript-project-jsconfigjson) files are used to define JavaScript projects . The presence of a `jsconfig.json` file tells an editor to treat all `*.js` files in the workspace as part of the same project, which improves the editor experience by enabling features such as `find all references` and `go to definition`. A `jsconfig.json` can also be used to change project wide settings, such as only allowing functions from the ES5 standard library, or even enable [more advanced type checking for JS files](https://code.visualstudio.com/docs/languages/javascript#_type-checking-and-quick-fixes-for-javascript-files)

Editors that use TypeScript to power their JavaScript / TypeScript language support can benefit from a `jsconfig.json`. This includes:

- [VS Code](https://github.com/Microsoft/vscode) (I'm the developer of VS Code's built-in JS/TS  support)
- [Atom TypeScript](https://atom.io/packages/atom-typescript)
- [Sublime TypeScript](https://github.com/Microsoft/TypeScript-Sublime-Plugin)
- Visual Studio
- Others?

I propose adding a basic `jsconfig.json` to the root of this project. This will improve the getting started experience for users of an editor with `jsconfig.json` support, and the file can be safely ignored otherwise

vigzmv:
Hey @mjbvz! Thanks for the contribution.
While `jsconfig.json` is beneficial and can provide more features to the developer, I feel this is something developer dependent and doesn't need to be in the repository. Developers can create it locally and customize it to their preferences as per their needs.
This is just my opinion, let's us see what others think.

mjbvz:
@vigzmv Yes, for us the challenge is discoverability. Many developers using vscode for example are not aware of `jsconfig` in the first place, or may not understand how much creating one can improve the editing experience. Having a simple, generic, jsconfig will make the editor experience better for these users and also hint that they can configure the `jsconfig` further to better match their needs

vigzmv:
https://code.visualstudio.com/docs/languages/jsconfig

blling:
+1.
[...]

PR #2097, opened 2018-01-08
Make it work with hyperclick

piuccio:
[Hyperclick](https://atom.io/packages/hyperclick) is an amazing plugin for Atom / Sublime that allows to cmd+click on a variable and go to the declaration.

It doesn't work on RB because we don't have relative imports but rely on webpack to automagically prefix './app/' to all imports that are not inside `node_modules` [here](https://github.com/react-boilerplate/react-boilerplate/blob/master/internals/webpack/webpack.base.babel.js#L102)

This PR configures [js-hyperclick](https://github.com/AsaAyers/js-hyperclick#i-configured-babeleslintflowwebpacketc-to-avoid--in-my-imports-how-can-i-configure-js-hyperclick) accordingly


blling:
This way is better, do you think so?
https://github.com/react-boilerplate/react-boilerplate/issues/2073#issuecomment-352818193

piuccio:
Jsconfig won't work in this case.

Only changing the docs is an option, but this package being a boilerplate, I think it should come with such basic configuration

PR #2098, opened 2018-01-08
Allow for an optionally namespaced reducer

dfredriksen:
Setting isNamespacedReducer will pass the namespace key into the reducer function, which should be modified to be wrapped in a lambda to receive the namespace value.

Example withReducer definition:
`const withReducer = injectReducer({ key: 'form', reducer, isNamespaced: true });`

In this case, "form" will become the default which can be overridden on a per component basis by passing in a unique reducerKey prop.

Example namespaced reducer definition:

`const formReducer = (namespace) => (state = initialState, action, props) => {
    switch(action.type) {
        case 'SOME_ACTION':
           if(action.namespace !== namespace) {
               return state
           } else {
               return state
                  .set('something', action.someval)
           }
        default:
            return state
    }
}`

Of course you would have to pass the namespace into the action.

Would there be a good reason not to do this? Otherwise feel free to use or leave up for anyone who might want to optionally use.

## React Boilerplate

- [ ] You have followed our [**contributing guidelines**](https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md)
- [ ] double-check your branch is based on `dev` and targets `dev` 
- [ ] Pull request has tests (we are going for 100% coverage!)

### DIFF
--- a/app/utils/injectReducer.js
+++ b/app/utils/injectReducer.js
@@ -11,7 +11,7 @@ import getInjectors from './reducerInjectors';
  * @param {function} reducer A reducer that will be injected
  *
  */
-export default ({ key, reducer }) => (WrappedComponent) => {
+export default ({ key, reducer, isNamespaced = false }) => (WrappedComponent) => {
   class ReducerInjector extends React.Component {
     static WrappedComponent = WrappedComponent;
     static contextTypes = {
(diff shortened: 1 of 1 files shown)


Droow:
I think it can be also solved by this: https://github.com/ioof-holdings/redux-subspace
Especially with `redux-subspace-saga` and `react-redux-subspace`

So question is if this should be integrated in RBP or anyone who needs this feature can use `redux-subspace`. Of course, I will leave that decision to the maintainers.

PR #2109, opened 2018-01-16
Allow to pass additional parameters to the injected sagas

piuccio:
This PR needs some additional tests, but I wanted to have your feedback first.

In RBP 3.4 you would create your sagas as

```js
function* mySaga(action) {
   console.log(action); // { type: 'MY_ACTION' }
}
```

In 3.5 instead if you're using the `injectSaga` you end up with

```js
function* mySaga(whatsThis) {
   console.log(whatsThis); // it's the whole props object
   // turns out that the action is now the second parameter
}
```

As I'm updating my code, I really don't fancy modifying all my sagas to add an extra argument.

Moreover, I don't really like using props inside my saga, feels brittle. I thought the whole point of using actions was to pass parameters to a saga that doesn't have to depend on the props.

Finally, in another project based on 3.4 I need to access a firebase instance (irrelevant) and I want to have some sort of indirection. So instead of `import` the instance I need, I pass it as an argument to `injectSaga` inside `routes.js`. This option is now gone, there's no way to pass anything to the saga.

Getting to the point, I think it's handy to be able to pass stuff down the saga.

The proposed implementation works like this

```js
injectSaga({ key: 's1', saga: saga1 }); // same as the current dev branch
injectSaga({ key: 's2', saga: saga2, args: [] });
injectSaga({ key: 's3', saga: saga3, args: [{ banana: ' yellow'}] });

function* saga1 (props, action) {}
function* saga2 (action) {}
function* saga3 (fruit, action) {}
```

What do you guys think?

I don't like too much that there's no way to have `fruit` and `props` at the same time, so maybe there should be a flag to turn `props` on or off?

```js
injectSaga({ key: 'flag1', saga: saga1 }); // same as the current dev branch
injectSaga({ key: 'flag2', saga: saga2, props: false });
injectSaga({ key: 'flag3', saga: saga3, args: [{ banana: ' yellow'}], props: false });
injectSaga({ key: 'flag4', saga: saga4, args: [{ banana: ' yellow'}] });

function* saga1 (props, action) {}
function* saga2 (action) {}
function* saga3 (fruit, action) {}
function* saga4 (fruit, props, action) {}
```

For what it's worth, I don't think I'll ever use `props` in my saga

### DIFF
--- a/app/utils/injectSaga.js
+++ b/app/utils/injectSaga.js
@@ -13,9 +13,13 @@ import getInjectors from './sagaInjectors';
  * cancelled with `task.cancel()` on component un-mount for improved performance. Another two options:
  *   - constants.DAEMON—starts the saga on component mount and never cancels it or starts again,
  *   - constants.ONCE_TILL_UNMOUNT—behaves like 'RESTART_ON_REMOUNT' but never runs it again.
- *
+ * @param {array} [args] Arguments passed to the saga once called
+ * By default your saga will receive
+ *   - component props
+ *   - action
+ * If defined, the saga will receive those args instead of the component props
  */
-export default ({ key, saga, mode }) => (WrappedComponent) => {
+export default ({ key, saga, mode, args }) => (WrappedComponent) => {
   class InjectSaga extends React.Component {
     static WrappedComponent = WrappedComponent;
     static contextTypes = {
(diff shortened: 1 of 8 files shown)


gretzky:
travis build is failing -- if that could be taken care of, i'd be glad to look at this PR

piuccio:
Sorry, I know it's failing, let me quote my first line

> This PR needs some additional tests, but I wanted to have your feedback first.

Can you please just comment on the message above first? I don't have time to work on a PR that will eventually be forgotten because it's not considered useful

gretzky:
@piuccio Everyone here is doing this in their spare time. If you do not have time to work on a PR that you feel will be forgotten, then I'm unsure as to why you opened this PR?

The TravisCI build shows that it failed in the file you edited, so with that info it's hard for me to be able to give feedback on this PR that's failing from the beginning. When there is a working solution, I'll be able to give feedback. 

If you choose not to fix, I'll close this PR.

piuccio:
@gretzky I fixed the tests

piuccio:
thanks

PR #2124, opened 2018-01-31
Feature/add styles constant

### LINKED ISSUE #2115: RFC:  Styles constant
## Issue Type

- [ ] Bug (https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md#bug-reports)
- [x] Feature (https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md#feature-requests)

## Description

Common style properties are currently defined in many places throughout the boilerplate.  For example, the color #41addd is written 7 times in 3 different files in the example application.  This becomes messy in larger applications.  With many teammembers, maintaining consistent colors and breakpoints becomes challenging and a source of headache.

A solution we've used with react-boilerplate is to create a styles constant, which can be imported and referenced to avoid having to write out this common css multiple times.  An example styles constant could look like:

```js
export default {
   color: {
      primary: '#41addd',
   },
   breakpoint: {
      tablet: '768px'
   },
}
```

These would be used as follows:

```
import styles from 'styles/constants`

const buttonStyles = css`
   color: ${styles.color.primary};
   width: 100%;

   @media only screen and (min-width: ${styles.breakpoint.tablet}) {
      max-width: 300px; 
   }
`
```

The concern I see being raised is that this approach ends up tying components together a bit.  The advantage brought in is better style scalability and potential for theming, while keeping the code a bit more DRY.

I'm happy to open a pull request for this, however wanted to have the idea vetted by the maintainers first.

# Versions

- React-Boilerplate (see `package.json`): 3.5.0

stabback:
Closes #2115 

I'd like to leave discussion of the idea itself to the issue (which is an RFC).  If approved, this will be my first contribution to react-boilerplate.  I'm happy to take all pointers and address all concerns.


gretzky:
I think because this is a bit bigger than a regular PR it makes sense to have its own branch outside of dev. Hopefully that'll further a conversation about changing up styles.

I've created a new branch -- `styles-rfc`. If you could switch your base branch to that one, I'd be more than happy to merge this in.

avdeev:
@stabback @gretzky What about SCREAMING_SNAKE_CASE for constants?

Related with https://github.com/airbnb/javascript/pull/255

PR #2146, opened 2018-03-04
Test matrix update

MariusRumpf:
Updates test matrix to remove end of life node v7 and include the new stable version v9. 
Add maxWorkers argument to travis tests to [improve build time](https://facebook.github.io/jest/docs/en/troubleshooting.html#tests-are-extremely-slow-on-docker-and-or-continuous-integration-ci-server).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: node_js
 
 node_js:
+  - 9
   - 8
-  - 7
   - 6
 
 script:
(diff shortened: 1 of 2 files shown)


julienben:
I can't find much evidence for the maxWorkers=4 trick (cf. [here](https://github.com/facebook/jest/issues/3765) or [here](https://github.com/facebook/jest/issues/4004)) but in practice it's yielded very short build times on our Travis so I agree with adding this. Also agree with the update to targeted node versions.

julienben:
@gretzky You got any thoughts on this?

PR #2164, opened 2018-03-18
Webpack v4 addendum

julienben:
# Update Log

This is a follow up to the Webpack v4 upgrade thread [here](https://github.com/react-boilerplate/react-boilerplate/pull/2148).

The PWA manifest.json is now generated by the webpack-pwa-manifest plugin. I updated the documentation accordingly and also upgraded some dependencies.

I also disabled JPEG optimization via image-webpack-loader as it was breaking the build in Travis CI (no matter the version). I left a NOTE explaining how to enable it quickly.

## Tooling Versions

* Node 8.9.4
* npm 5.7.1
* yarn 1.5.1
* Mac OS 10.13.3
* Chrome 65.0.3325.162 (64-bit)

## :package: Version Diff

**[0] PATCH UPDATES**

```
 hoist-non-react-statics                    2.3.1  →  2.5.0
 enzyme                                     3.2.0  →  3.3.0
 stylelint-config-standard                  18.0.0 → 18.2.0
 stylelint-processor-styled-components      1.1.1  →  1.3.1
```

**[1] MINOR UPDATES**

```
 chalk                                   2.3.0  →    2.3.2
 compression                             1.7.1  →    1.7.2
 cross-env                               5.1.1  →    5.1.4
 express                                4.16.2  →   4.16.3
 invariant                               2.2.2  →    2.2.4
 lodash                                 4.17.4  →   4.17.5
 prop-types                             15.6.0  →   15.6.1
 react-redux                             5.0.6  →    5.0.7

 css-loader                            0.28.10  →  0.28.11
 enzyme-adapter-react-16                 1.1.0  →    1.1.1
 eslint-import-resolver-webpack          0.8.3  →    0.8.4
 html-webpack-plugin                     3.0.4  →    3.0.6
 rimraf                                  2.6.1  →    2.6.2
 style-loader                           0.20.2  →   0.20.3
 webpack                                 4.1.0  →    4.1.1
 webpack-cli                            2.0.10  →   2.0.12
 webpack-dev-middleware                  3.0.0  →    3.0.1
```

**[3] MAJOR UPDATES**

```
 image-webpack-loader                  3.6.0  →   4.2.0
```

**[4] NEW DEPENDENCIES**

```
 webpack-pwa-manifest                           3.6.2
```

### Warnings: Observed on `yarn install` after removing node_modules

*  warning "stylelint > postcss-html@0.11.0" has unmet peer dependency "postcss-sass@>=0.2.0".

PR #2185, opened 2018-04-17
Adding missing svg-url-loader dependency to package.json

julienben:
I added svg-url-loader to the webpack plugins in #2164 but forgot to add it to package.json. This is a fix for that.


julienben:
@gretzky Github won't let me suggest you as a reviewer but if you have a second, this is a VERY short one ;)

gretzky:
@julienben I'm offended it won't let me review!

gretzky:
@julienben all set buddy

julienben:
Thanks!

PR #271, opened 2016-04-08
Updated all deps

oliverturner:
In honour of React v15 being released I went nuts and upgraded all the things :)

PR #2202, opened 2018-05-02
Updating order of module resolution in Webpack config

stern-shawn:
As encountered [here](https://github.com/ReactTraining/react-router/issues/6126), if someone uses and extends the boilerplate to create a singleton of say, `history` so it can be used in sagas/epics and so on, the module resolution order of apps -> node_modules means that react-router components like Redirect will import from the wrong instance of `history` (the local declaration in /app instead of the actual library in /node_modules).

Switching the order doesn't seem to cause any harm and means less chance of this affecting users.

Also splitting up the fields in the array onto multiple lines since it fits with all the other nearby arrays 😄 

## React Boilerplate

### DIFF
--- a/internals/webpack/webpack.base.babel.js
+++ b/internals/webpack/webpack.base.babel.js
@@ -125,7 +125,10 @@ module.exports = (options) => ({
     }),
   ]),
   resolve: {
-    modules: ['app', 'node_modules'],
+    modules: [
+      'node_modules',
+      'app',
+    ],
     extensions: [
       '.js',
       '.jsx',


julienben:
I read through your issue in `react-router` but couldn't reproduce it in my apps even with the same history setup. The redirects were still working.

That being said, I checked why we have module resolution in its current order and it seems to have been done in #271 to deal with an Enzyme issue. Since this doesn't break any tests and fixes an issue for you, I would be inclined to approve.

@gretzky Any quick thoughts?

@oliverturner What about you since you wrote #271?

stern-shawn:
@julienben Just to satisfy my curiosity, did you run `npm run build:dll` after adding the new `app/history.js` implementation? This originally got past me because I was working on it in local development and it'd been awhile since doing npm i/run build so I had an old dll with the correct history association. After building a new dll with the original resolution order is when I ran into the odd import in webpack

```js
var _history = __webpack_require__(/*! history */ "./app/history.js");
```

Either way, thanks!

julienben:
Yeah I don't think it changed anything for me.

PR #2223, opened 2018-05-28
add Lock app template

gretzky:
I've added [Lock](https://probot.github.io/apps/lock/), a bot that locks closed issues that are a certain age old. I've set the number of days to be 30, because I constantly get notifications of people who are commenting on closed issues. Most of these comments / issues aren't an rbp-related issue, rather just asking for help. I want to curb convos on old, closed issues and then if/when someone opens a new issue, we can either solve the problem or direct them to StackOverflow. 

Let me know if this is overkill, but I think it just takes some of the weight off.

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -1,0 +1,29 @@
+# Configuration for lock-threads - https://github.com/dessant/lock-threads
+
+# Number of days of inactivity before a closed issue or pull request is locked
+daysUntilLock: 30
+
+# Issues and pull requests with these labels will not be locked. Set to `[]` to disable
+exemptLabels: []
+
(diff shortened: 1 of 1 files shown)


julienben:
I'm all for this. Just wondering: Does it also lock 30 day old PRs and RBP-related discussions? And is it reversible?

gretzky:
Yeah it's reversible-- as in we can override it. It doesn't lock PRs, but it'll close RBP related discussions if they're in issues and over 30 days old

gretzky:
I think locking the RBP discussions is fine since I think a large majority, if not all of them, of them are stale

julienben:
Thanks for the 520 emails 😂

PR #2228, opened 2018-05-30
omg make it stop (changed lockbot config file)

gretzky:
I changed the lock bot config file to add a label instead of a comment so we don't get hundreds and hundreds of notifications 😭


julienben:
The problem is that it's still working its way through our backlog. It's locking hundreds of old+closed threads every hour. The benefits of this bot might show up once it's done with that so let's maybe give it another shot? In the meantime, I've set up a quick Gmail filter to auto-delete anything coming from lock-bot. Wanna maybe try that too?

gretzky:
Agreed

PR #2235, opened 2018-06-08
v3.6.0 🎉

gretzky:
Hola amigos,

I've successfully cherry-picked a bunch of commits that should suffice as a minor version upgrade. It contains the following

* React v16
* Redux v4
* Webpack v4
* Enzyme v3
* Minor code adjustments due to the above upgrades
* Some documentation upgrades
* Reconfiguration of the way we test styled-components, using [jest-styled-components](https://github.com/styled-components/jest-styled-components)

So far, the only issue I'm running into is the reducers.js test failing. I'm sure I'm just overlooking something, but if anyone could pull this branch down and fix it, it would be much appreciated.

I think this is a good thing to push out soon because of all the major dep upgrades. After, we can then focus on more of the commits in the dev branch and other PRs.


julienben:
Congrats on the huge PR!! I'll review and also try to fix tests in the next few days. (Currently traveling again :)

gretzky:
Failing checks, I'll take a look at those soon

nguyenvanhoang26041994:
Why not React 16.3?

package.json
```diff
@@ -209,92 +209,111 @@
...
+    "react-test-renderer": "16.2.0",
+    "rimraf": "2.6.2",
     "shelljs": "0.7.7",
-    "style-loader": "0.18.1",
-    "url-loader": "0.5.8",
-    "webpack": "3.5.5",
-    "webpack-dev-middleware": "1.11.0",
-    "webpack-hot-middleware": "2.18.0"
+    "style-loader": "0.20.3",
+    "stylelint": "8.3.1",
```
MariusRumpf: You added the `stylelint` packages, but no configuration for them.

package.json
```diff
@@ -35,7 +35,7 @@
     "clean:all": "npm run analyze:clean && npm run test:clean && npm run build:clean",
     "generate": "plop --plopfile internals/generators/index.js",
     "lint": "npm run lint:js",
-    "lint:eslint": "eslint --ignore-path .gitignore --ignore-pattern internals/scripts",
+    "lint:eslint": "eslint --ignore-path .gitignore --ignore-pattern internals/scripts --fix",
```
MariusRumpf: If a staged file is linted before commiting it is fixed, but the changes are not staged and the commit is made. This way you can commit incorrectly formatted files. `lint-staged` should not use the `--fix` flag.

julienben:
I wanted to push some code to your branch but it's from your own fork of the repo, not an actual RBP branch, so I'm not authorized. Pushed to a separate branch instead.

Check it out here: https://github.com/react-boilerplate/react-boilerplate/commit/fa12ee68199ea8adec2ecd49078c78551d135876

It's just a start but I believe I fixed the remaining issues with the reducers and testing.

- The first failing test (in reducers.test.js) was just missing an export on the routeReducer function.
- And the second... well it didn't make much sense to me so I rewrote it. In any case, the original version of the test was pushing an object as initial state when it's supposed to be Immutable (hence the "state.merge is not a function" error).
- I also provide initialState to the selectors. Happy to discuss the rationale for this but the app simply doesn't work for me without it.

Cheers!

package.json
julienben: Agreed! There's an alternative way to fix it though (which I've been using for a long time):

image

julienben: (This will add the fixes to your commit.)

package.json
julienben: Here's the one I use if it helps:
```json
{
  "extends": "stylelint-config-standard",
  "rules": {
    "selector-pseudo-class-no-unknown": [
      true,
      {
        "ignorePseudoClasses": [
          "global"
        ]
      }
    ],
    "no-descending-specificity": null,
    "declaration-colon-newline-after": null,
    "value-list-comma-newline-after": null,
  }
}
```

julienben:
I believe we should also add the changes from #2146.

julienben:
You can use #2240 if you want.

package.json
julienben: @gretzky What do you think of this? Remove the --fix flag or keep it but make sure they're added to the commit?

PR #2241, opened 2018-06-11
v3.6.0 -- new branch

gretzky:
This moves #2235 to its own branch so @julienben and I can hack away at it. It just pulls in my changes into a new branch here.

Merge it regardless of CI because its not going to master til those pass


gretzky:
@julienben all set, we can work on this branch

PR #2095, opened 2018-01-07
feat(internals): Add prettier integration

blling:
JG: resolves #1945

Replace #2083

This PR complete the following task:
- [x] Add prettier integration.

@azz's suggestions:
Use the `plugin:prettier/recommended` ESLint config: https://github.com/prettier/eslint-plugin-prettier#recommended-configuration
`$ yarn add --dev eslint-plugin-prettier eslint-config-prettier`
```json
// .eslintrc.json
"extends": ["plugin:prettier/recommended"],
 ```
And just add one custom option now:
```json
"prettier": {
  "singleQuote": true
},
```

And add  `lint-stage` as follows:
```json
  "lint-staged": {
    "*.js": [
      "prettier --write",
      "lint:eslint",
      "stylelint",
      "git add"
    ]
  },
```
`git add` in the `lint-stage` will add the correct file:)

- [x] Run prettier on the entire codebase and fix problems after run prettier.

Two main changes:
1. Add `stylelint` rules:
```javascript
    "rules": {
      "rule-empty-line-before": null,
      "declaration-block-semicolon-newline-after": null
    }
```
2. Replace `eslint-disable-line` to `eslint-disable-next-line` in some special lines. 

- [x] Fix generate CI failure.

Add `prettier` action for plop, in order to format the js file after generated.

PR #2242, opened 2018-06-12
v3.6.0 -- the definitive branch

julienben:
Let's discuss our ongoing progress here.

Refer back to #2235 for the initial discussion on v3.6.0.


mxstbr:
Deploy preview for *react-boilerplate* failed.

Built with commit 85a0d105bf7674077202f2c92b09b86f928114d3

https://app.netlify.com/sites/react-boilerplate/deploys/5b1fc2e9b312747ea679a960

julienben:
I pushed a few things. None of them major. 

Major deps we might still want to update for this release:
- styled-components (v2 to v3)
- eslint
- jest

I'll go over them ASAP and fix the relevant breaking changes (as well as any corresponding docs).

Open questions:
- Are we keeping stylelint? If yes, we need to give it a config.
- Anything else?

Before release, we'll need to do an in-depth review, testing and make sure docs are up-to-date.

gretzky:
IMO...

- we should ditch stylelint for now-- I can't remember who but someone contributed a great config and its on its own branch, so lets revisit that particular branch instead. It dealt with styles as a whole as well.

- i prefer the `--fix` tag but i'm open to anything here, just my preference

- none off the top of my head now but i'll come back later

mxstbr:
styled-components v2 to v3 should be a painless upgrade fwiw, almost no breaking changes

julienben:
stylelint has been in here since v3.0 according to Changelog.md. I think it would be a flub to remove it without replacing it with something else.

Were you referring to #2095? Cause it has both some stylelint and prettier-related stuff, both of which I would be happy to add into 3.6. Unless there are objections.

julienben:
Sounds good for the `--fix` flag! I added a `git add` cmd to ensure fixes get added to the commit.

julienben:
@mxstbr I quickly read through the releases of `styled-components` from 3.0 to the latest 3.3.2 and didn't find the need for any changes in the codebase. Tests still pass and everything looks good in the browser. Could you let us know if there's anything we missed maybe? I'm not a styled-components user so I don't know its ins and outs.

mxstbr:
Nope, shouldn't require any changes for y'all!

julienben:
I upgraded a bunch more dependencies after going through their change logs. Will continue tomorrow.

Mensae:
I've been [using](https://github.com/Mensae/portfolio-react-boilerplate) styled-components, ^3.1.5, for quite a while now. No issues at all.

julienben:
@gretzky I added a few updates (mostly react and offline-plugin which now supports webpack v4). Nothing huge. However, I opened a new branch with v3.6.0 as the base. In it, I started the process of upgrading eslint.

It's in its own branch because it requires lots of changes and will make this PR's diff unreadable. So I suggest merging the eslint branch into v3.6.0 at the end, before release. Sounds good? (It's #2244)

Second question: Any objection to me using that same branch to add prettier?

gretzky:
PLEASE add Prettier!!!!!

gretzky:
Great work @julienben, let me know if there's anything specific you want me to look at. Otherwise, I'm going to plug away at a CRA-RBP [combo repo](https://github.com/gretzky/cra-rbp)

mxstbr:
> Otherwise, I'm going to plug away at a CRA-RBP combo repo

Yesssssss :tada: :tada: :tada:

julienben:
I just pushed what should hopefully be the last major commit. Still has a minor issue with node versions and dependencies but I'm on it. It includes:

- A few more dependency upgrades
- `webpack-cli` and `jest-cli` have moved major versions but thankfully the breaking changes don't affect us.
- The dependencies which weren't upgraded to their latest versions (`ngrok` and `plop`) couldn't be because of Node 8+ requirements. 
 - I also added a `stylelint` config and command (as recommended by the [styled-components website](https://www.styled-components.com/docs/tooling#stylelint)) and wrote some docs for it.

I believe we're pretty much ready for final review. The only things still open in my mind are:

- Any more doc updates we want to make?
- Let's write a changelog. I'll do it unless someone wants to volunteer.
- Let's do some in-depth testing and reviewing to catch whatever AppVeyor and Travis can't. Anyone who wants to help with this is more than welcome. Especially those with unique setups (*ahem* Windows *ahem*) or older versions of Node. We need:
  - Full Regression Testing [as described in the docs](https://github.com/react-boilerplate/react-boilerplate/blob/master/docs/maintenance/dependency.md)
  - Checking that the template app was properly updated in the same way as the sample app.
  - Checking the generators.

@react-boilerplate/collaborators @mxstbr Final stretch! Let's get this out there!

julienben:
PS: Don't forget to use the the no-whitespace feature for a MUCH shorter diff:

[https://github.com/react-boilerplate/react-boilerplate/pull/2242/files?w=1](https://github.com/react-boilerplate/react-boilerplate/pull/2242/files?w=1)

gretzky:
@justingreenberg agreed on json files

gretzky:
I combed through the docs and I couldn't think of anything that needed to be added-- I added a few in the original cherry picks before so those should cover most new material (from what I remember). Users are always great at pointing out any lacking docs as well.

I can write the changelog right now

justingreenberg:
@gretzky ok, i will add to pretierignore

i'm working on cross-OS compatability to fix AppVeyor build... it does not like some of the filesystem stuff we are doing. i can't reproduce the error locally, so i will be triggering builds to test it.

maybe this has already been addressed and i missed it, but is there really a good reason to continue supporting node 6?

on another note, in dealing with the `extract-intl` issue, it was kinda crazy that the `DEFAULT_LOCALE` was hidden in a deeply nested folder (even in generated app) which is used in build scripts, etc.

```js
import { DEFAULT_LOCALE } from '../containers/App/constants';
```

so, i moved it to root `i18n.js` which i think makes a lot more sense

julienben:
It's late here so just a quick comment about JSON and prettier: I added it to the pre-commit hook as well so you'll want to remove it there too.

That being said, do you have a quick example of a JSON that is made less readable by prettier? Only cause I'm not sure what you mean. I use it in my projects because (for regulation purposes - medical field) I have to keep logs of all my tests on every release and Jest outputs a giant JSON with no formatting whatsoever. So the commit hook fixes that for me in a completely effortless way.

Also, could your issue with it be fixed/mitigated by using a larger char width on JSON files?

gretzky:
@justingreenberg I'm all for removing support for Node 6. Personally, I know of many projects that still use it but they're all having out-of-date dependency issues that would be solved by upgrading. I say we nuke node 6.

justingreenberg:
this is almost ready, just need to update the plop Regex to conform to new file updates... i will try to wrap it up tonight

ref https://github.com/react-boilerplate/react-boilerplate/blob/master/internals/generators/language/index.js

julienben:
Some updates:
- Upgraded the last dependencies that needed Node v6. Had to update the ngrok setup with it.
- Ran the Full Regression Testing and everything went fine except for the generators needing a small fix in the selectors. Done.
- Had to add an .nvmrc per [Netlify docs](https://www.netlify.com/docs/) to trigger a build on Node 8 instead of 6.

Small issue:
- When you use a generator, the code coming out of it isn't always passing linting. We can't fix that permanently because it depends on the length of the component/container name. The only thing we could do would be to run "prettier --write" on the files coming out of the generator. Seems like a bit of overkill but feasible as it's already happening in `generate-templates-for-linting.js`.
- Prettier + JSON is still alive for now.

mxstbr:
baee117 should've probably not been a squash commit, which means all the contributors' work is now credited to @gretzky if I understand that correctly? Shall we undo that before merging this so every contributor gets their name on the contributors list?

julienben:
@mxstbr How does one do that? In a completely new branch?

julienben:
I can't approve my own PR so just letting you (@gretzky and @justingreenberg) know that I'm good with the codebase as it stands now. Nothing I feel the need to add or change before this becomes v3.6. You?

mxstbr:
Not sure, I'm not really a git wizard 😅 There's gotta be a way to take a new branch without squashing all the commits and then moving all these commits over there—maybe just gotta cherry pick them all?

julienben:
Closed in favor of #2248

PR #2244, opened 2018-06-14
ESLint upgrade + Prettier

julienben:
See #2235 for reasoning behind this branch.

Note that this won't pass checks yet as it has some outstanding linting issues. I committed using the --no-verify flag on git commit.


mxstbr:
Deploy preview for *react-boilerplate* failed.

Built with commit ac193efe0268409ad0136dcd8a704fb2b87180e8

https://app.netlify.com/sites/react-boilerplate/deploys/5b276e4b3672df201372cfcf

julienben:
This PR contains the following changes:

- Upgraded ESLint from v3 to v4 (including all related dependencies + small rules changes)
- Set up Prettier in full (including config and ignore files)
- Integrated ESLint and Prettier using the relevant base config and plugin
- Moved ESLint config to a JS file to import the Prettier setup into ESLint's rules
- Made all the code updates/formatting required by new ESLint + Prettier
- Updated the generators to push out code more in line with the new linting setup
- Had to use ES5-only trailing commas to maintain Node 6 compatibility. Can be adjusted individually by Node 8+ users.
- Added a pre-commit step for JSON linting and autofixing
- In `generate-templates-for-linting.js`, added a step to prettify generated code

Reasoning for the prettify step in `generate-templates-for-linting.js`: Generated JS is not always formatted exactly per Prettier's requirements. This is due to Prettier's fixed printWidth clashing with the (varying) component name length. Thus, we prettify each component after generation. Otherwise, CI would break.

Still missing: Some documentation on the new setup

julienben:
Since this branch will be merged into v3.6 (#2242) right before release, it will be rebased at that time and will need a new approval. However, it's still ready for an actual review.

So @gretzky @mxstbr @justingreenberg, you (and anybody else here from the team) are more than welcome to review this.

You can see my post above for a detailed list of all the changes. I'm just trying out a fix for AppVeyor at the moment but other than that, I believe it's ready. Unless you can think of some relevant additions or fixes for the docs.

PR #2248, opened 2018-06-20
V3.6.0 with attribution

julienben:
Per @mxstbr's suggestion on #2242, we're introducing a final v3.6.0 PR to fix contributor attribution in git history.

@gretzky @justingreenberg Let's pick up the conversation here. This contains exactly the same changes as the previous branch (as proven by the diff stats).


mxstbr:
Yesss this is the much better! 🙌 How'd you make that happen?!

julienben:
Merged with rebase (instead of squash) from @gretzky's rbp fork and then lots of cherry picking. It had exactly the same number of additions and deletions as the last v3.6 branch (until we made changes here) so I'm hoping it means I was accurate and didn't forget anything.

julienben:
I think @mxstbr's intention was for the stats on [this page](https://github.com/react-boilerplate/react-boilerplate/graphs/contributors) to be accurate. It's possible that they would've been despite the squash commit anyway though. I'm not sure.

We also have a MAINTAINERS.md btw. I added both our names to it, @gretzky.

Speaking of which, do you have any final changes or are we ready to release? Do you want to do the honors?

gretzky:
Got it got it. I think we're ready! I'm pulling it 🔫 . Thanks for letting me do the honors @julienben, and kickass work you've put in. ✋ hi-5.

julienben:
✋ Major props for actually initiating this release and seeing it all the way through!

(I just tagged it btw cause you had only merged.)

Mensae:
Especially to @julienben and @gretzky Thanks for all the work!

gretzky:
@julienben my right hand man

PR #2249, opened 2018-06-21
Bring the dev branch back up to speed

julienben:
Now that we've released v3.6, the next logical step is to merge master into dev, right?

Any conflicts, we'll keep the changes from master. The two branches diverged in September and are looking quite different now.

PS: This is in anticipation of #2300 "Why is this project not on React 19 yet?" 😜


gretzky:
why aren't we on webpack v1400 yet?????????

jk but for real this dev branch and outstanding PRs need to be pruned bad. i'll start going through it and open some RFC discussion for certain things if need be.

in addition to that, i want to beef up the webpack. it can be better. i've started on the [webpack-optimization branch](https://github.com/react-boilerplate/react-boilerplate/tree/webpack-optimization)

julienben:
So I'll resolve the conflicts but are we in agreement that it means keeping the changes from master each time?

julienben:
(Any other branches based on dev will have to rebase on top of this)

julienben:
Closed in favor of #2262

PR #2254, opened 2018-06-25
Fix the eslint-fix script

gretzky:
When running `lint:eslint:fix` nothing happens because we aren't invoking eslint like in all other scripts and resolving the root dir.

I reworked the script so it just covers `app/` and runs the `--fix` flag

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "lint": "npm run lint:js && npm run lint:css",
     "lint:css": "stylelint './app/**/*.js'",
     "lint:eslint": "eslint --ignore-path .gitignore --ignore-pattern internals/scripts",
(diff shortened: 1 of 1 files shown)


julienben:
This command isn't meant to be run without adding a path or a blob. It's used by the pre-commit hook right now and as per [their docs](https://github.com/okonet/lint-staged) they pass "the list of currently staged files as arguments." So I believe this would break the pre-commit hook.

```json
"lint-staged": {
    "*.js": [
      "npm run lint:eslint:fix",
      "git add --force"
    ]
  }
```

To lint manually, one can use `npm run lint` or `npm run lint:eslint:fix -- .` to fix. There's a short explanation in the docs [here](https://github.com/react-boilerplate/react-boilerplate/blob/master/docs/general/commands.md#linting).

PR #2272, opened 2018-07-10
Get the dev branch right

gretzky:
Alright, let's see if this works?


gretzky:
If this doesn't work, I'm down to just recreate the dev branch. Maybe create a new dev branch, rename the old dev branch, and then set the new default branch as the freshly created dev branch.

Man that's a tongue twister.

But that way, we can still keep the old dev branch and cherry-pick any old PRs from it.

gretzky:
Actually, I'm just gonna do that.

gretzky:
I was able to get around the conflicts by making a [new branch](https://github.com/react-boilerplate/react-boilerplate/tree/new-dev). We could un-protect the dev branch, rename it, and then make this branch the new dev?

julienben:
> Man that's a tongue twister.

And yet it made perfect sense.

> We could un-protect the dev branch, rename it, and then make this branch the new dev?

Agreed. Good idea to keep the original dev branch.

Question though: Did you create new-dev from the old dev branch? It ought to be from the master branch no? Since the dev branch is currently a bit confused about its identity and we want the basis of any new work to be the v3.6 release.

gretzky:
`new-dev` branch was branched off of master, and then `dev` merged in. it should be 100-something commits ahead of master, no commits behind so it's equal

gretzky:
But if we did that, we'd be ditching all of the PRs that other people made to the old dev branch? No? This branch is even with 3.6 and has the rest of the PRs that haven't been merged into a version. 

I'm fine with either way-- I just didn't know if we wanted to drop all of the previous PRs in dev.

Mensae:
Can I suggest to branch off ```master``` into a new ```dev```, call it like ```dev > v3.60``` or something? Perhaps a bit awkward to have to make an update to the current master with a change of the documentation for this, to point out people need to do new PR to that branch, but I am with @julienben and in favor of starting off with an exact mirror of the ```master``` but also with @gretzky that we should not loose pending PR on this ```dev```.
Another, better, suggestion would be to branch off ```dev``` into like ```dev < v3.6.0``` but not 100% sure what will happen with pending PR but hey, you could try and see :-)  ( I'm curious now and think I'll test it on one of my own )

julienben:
Thanks @Mensae!

I agree that we shouldn't lose the PRs that are in dev but which didn't make it into v3.6. However, because there's so much in there, I suggest we do the following:

1. Create a new dev branch to be even with v3.6. Make that the official dev branch.
2. Keep+rename the old dev branch which contains all the PRs that have been merged into dev but didn't make it into v3.6.

This is only so we can carefully review everything we cherry-pick from old-dev into the new dev branch. Agreed?

Mensae:
@julienben Hmmm... I don't think you can rename a branch, can you?

To come back on my second suggestion, with this we will loose, at least they will be closed, all pending PR

julienben:
Yeah, come to think of it, you can't rename a remote branch. Instead, you can rename a local branch (`git branch -m newname`), delete the old remote and create a new remote with the new name. Content and history will be the same but you're right about the PR issue.

That being said, I wasn't talking about open PRs. I was talking about all the things that have already been merged (the 100 or so commits that @gretzky mentions).

Unmerged commits can always be reopened and made to point to the new dev branch.

Mensae:
> That being said, I wasn't talking about open PRs. I was talking about all the things that have already been merged (the 100 or so commits that @gretzky mentions).

@julienben I know :-) The open PR is something I added to the equation, sorry about that ;-)
So yes, people can be asked to reopen and then I agree with your steps, although I guess you want the new ```dev``` to be named ```dev``` right? So then:

1. create a new branch out of the current ```dev```  and name it like ```dev-old```
2. delete the current ```dev```
3. create a new branch out of ```master``` and name it ```dev```
4. cherry pick from ```dev-old``` to ```dev```

Right?

gretzky:
@julienben I think anything will work honestly, just to get around some git things. I'm in favor of whatever makes the most sense without nuking things!

julienben:
Agreed! So to summarize everything, I believe the simplest plan is the following:

1. Create a copy of the current `dev` called `old-dev`
2. Unprotect `dev` temporarily
3. Reset it to the last commit where it was equal to `master` and then bring it back to the level of v3.6 (by doing a PR from `master` into `dev`)
4. Cherry pick from `old-dev` to `dev`

Plan B as described by @mensae is:

1. Create a copy of the current `dev` called `old-dev`
2. Delete the current `dev`
3. Create a new branch out of `master` and name it `dev`
4. Cherry pick from `old-dev` to `dev`

@gretzky, if you give me the go-ahead, I'll implement the first plan.

Maybe @mxstbr would like to weigh in before we do something this drastic?

gretzky:
I'm all in

julienben:
I think everything worked! Sorry for the scary Github emails you might've received in the last 15 minutes... About to restore protection to the `dev` branch.

This is currently the commit graph for the repo:

image

As you can see:
1. `master` and `dev` point to the same commit aka the v3.6 tag
2. The `old-dev` branch contains all the commits that used to be in `dev`

I think this means we're ready for step 4, cherry picking anything we want from `old-dev` into `dev`. (@gretzky, any thoughts on what we should take from `old-dev`?)

We can also start submitting any other work we want to submit on top of `dev` (i.e. Webpack optimization).

Final points:
- @gretzky, cool if I remove `new-dev`? I don't think we need it anymore.
- Existing PRs still point to `dev` but do have a different commit history now. Relevant contributors will need to rebase their commits onto the clean `dev` branch.

PR #2238, opened 2018-06-11
Dockerize

davidthornton:
Following from https://github.com/react-boilerplate/react-boilerplate/pull/753, which was closed due to inactivity, I have reopened this PR and applied some changes that take advantage of the stability and feature updates that Docker has provided in the intervening period.

The main changes over and above the previous PR are:

1. A focus on using Docker in development rather than production (these days k8s, static hosting or similar is probably a better strategy for orchestrating a react container in production)
2. An update to use node 10 (`node:10-slim` is the base Docker image from which this container derives)
3. The addition of a `docker-compose.yml` file which streamlines package dependency and startup installation time. From fAustralia, a `docker-compose build` took ~3 minutes to pull all dependencies and get up and running. The base OS accounts for 67 Mb and a fully installed development environment is about 740 Mb in size.
4. A focus on optimal DX: when using Docker with mounted volumes, touch reloading works perfectly and only three commands are required to update `npm` dependencies from other contributors. Only a single command is required to locally update dependencies: `docker-compose exec yarn add {NEW_PACKAGE_NAME}`.

PR #2282, opened 2018-07-19
Dockerize

davidthornton:
Re-creation of #2238 rebased on the new dev branch.

I've copied the PR description below for clarity and updated it slightly to reflect changes in the existing PR and this PR with the same fundamental intention.

______

Following from #753, which was closed due to inactivity, I have reopened this PR and applied some changes that take advantage of the stability and feature updates that Docker has provided in the intervening period.

The main changes over and above the previous PR are:

1. A focus on using Docker in development rather than production (these days k8s, static hosting or similar is probably a better strategy for orchestrating a react container in production)

2. An update to use node 10 (`node:10-slim` is the base Docker image from which this container derives)

3. The addition of a `docker-compose.yml` file which streamlines package dependency and startup installation time. From Australia, a `docker-compose build` took ~3 minutes to pull all dependencies and get up and running. The base OS accounts for 67 Mb and a fully installed development environment is about 740 Mb in size.

4. A focus on optimal DX: when using Docker with mounted volumes, touch reloading works perfectly and only three commands are required to update `npm` dependencies from other contributors. Only a single command is required to locally update dependencies: `docker-compose exec yarn add <package>`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,21 @@
+version: '3.3'
+services:
+  localdev:
+    build:
+      context: .
+      dockerfile: ./internals/docker/Dockerfile
+    command: npm run start
+    environment:
+      - NODE_ENV=localdev
+    ports:
+      - "3000:3000"
+    volumes:
+      - ./app:/code/app
+      - ./package.json:/code/package.json
+      - ./yarn.lock:/code/yarn.lock
+      - ./server:/code/server
+      - ./.gitignore:/code/.gitignore
+      - ./.eslintrc.js:/code/.eslintrc.js
+      - ./.prettierrc:/code/.prettierrc
+      - ./.prettierignore:/code/.prettierignore
+      - ./.stylelintrc:/code/.stylelintrc
--- a/docs/general/commands.md
+++ b/docs/general/commands.md
@@ -148,3 +148,29 @@ npm run lint:eslint:fix -- .
 ```
 
 Lints your code and tries to fix any errors it finds.
+
+## Using Docker in development
+
+*To use Docker for development, make sure you're within the root directory of the project.*
+
+To build the development container, run:
+
+```docker-compose build localdev
+docker-compose up localdev
+```
+In order to update dependencies, run:
+
+```docker-compose stop localdev
+docker-compose exec localdev yarn add <package>
+```
+
+> Note: Because this command attaches to a running container, it must be run while
+an instance of the development container is up.
+
+In order to receive new dependencies from other contributors, you must
+clear your mounted Docker volumes, by running:
+
+```docker-compose down -v localdev
+docker-compose build localdev
+docker-compose up localdev
+```
--- a/internals/docker/.dockerignore
+++ b/internals/docker/.dockerignore
@@ -1,0 +1,7 @@
+Dockerfile
+docker-compose.yml
+node_modules
+build
+coverage
+stats.json
+npm-debug.log
--- a/internals/docker/Dockerfile
+++ b/internals/docker/Dockerfile
@@ -1,0 +1,12 @@
+FROM node:10-slim
+
+RUN apt-get update && apt-get install libpng12-0 bzip2
+
+WORKDIR /code
+
+ADD package.json package.json
+ADD yarn.lock yarn.lock
+ADD internals internals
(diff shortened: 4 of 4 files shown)


julienben:
Quick question: How come you run `start:production` in your Dockerfile? Shouldn't it be just `start`?

davidthornton:
I think I did it to force a production start when the Dockerfile runs outside `docker-compose.yml`. I suppose there are good questions to be asked about the boundaries between environments and how far we want to go in guiding versus restricting others' development and delivery pipelines.

I personally use `docker-compose` in localdev and kubernetes in production (and I think this is the current state of the art) which means that the `docker-compose.yml` is ignored in production and the `Dockerfile` is the "only" thing a production container orchestrator interprets.

That being said, I think I have seen people moot running docker-compose in production (although I suspect most people would agree that it is an insufficient tool for production container orchestration)

julienben:
While I like Docker and use it at work every day, I'm not sure that I'm for its inclusion here.

I say that because:
1. This setup isn't tested via CI (i.e. it could break tomorrow without us knowing) and it adds one more thing to maintain.
2. It's not clear at first approach whether this is intended for use only in dev (and why) and can thus create confusion. 
3. It adds a custom value for NODE_ENV which, if you're not aware of it and taking it into account during dev, will create hard-to-debug inconsistent behaviors later.

@gretzky Please let me know if you think differently.

davidthornton:
To address your points:

1. It does for me... 
screen shot 2018-07-22 at 6 58 25 pm I didn't include the `docker-compose.test.yml` file to reduce the review overhead for now.

2. Docker is obviously a very powerful tool and I presented it as a dev only tool simply to get a mime viable workflow demonstrated. That being said, if you'd like further updates to docs I can do that.

3. I can remove this.

But in reality, if we don't want a Docker image to help let people quickly pull and run a react boilerplate -- just let me know and I'll step back from advocating for Docker.

There is a considerable amount of work involved in setting up an optimal Docker image while also maintaining an instantaneous developer iteration cycle, which I am willing to support but not if it doesn't make sense in this repository.

julienben:
1. How does it work? These tests run on Travis and Appveyor? Do they just run the current test suite inside a container instead of the current setups specified in appveyor.yml and .travis.yml? I'd be happy to see the PR in full if you have extra stuff.

2. It's up to us the find the balance between adding advanced features and making things reasonably easy to understand for newcomers. In this case, my impression is that we want to make it clearer that this is only for dev purposes OR make the setup compatible with both prod and dev concerns. I would prefer the second option.

3. Definitely a good idea IMO. The addition of new values for NODE_ENV should be left up to the users.

> There is a considerable amount of work involved in setting up an optimal Docker image while also maintaining an instantaneous developer iteration cycle, which I am willing to support but not if it doesn't make sense in this repository.

I understand and this frequently-required maintenance work is part of the problem. Much of the work done in this repo is dedicated to maintenance and I think we're weary of adding to that workload.

As to whether or not to go forward with this PR, I'm afraid I can't give you a yes or no answer. I would like to hear from other contributors (and also any users reading this!?) before committing. Personally, I'm for it if we can find a good solution to everything discussed above.

gretzky:
I personally have been against adding Docker from the beginning, because not every setup uses it. I don't want frequent users of RBP to then have to be tied to Docker if they don't want to be. IMO thats a large piece of the stack to integrate/change and I don't think just adding it is the correct way to go-- it would require input from the community I think. I'd be pretty pissed if one day I upgraded my RBP project and it then has a Dockerfile that I didn't ask for. But that's just me!

julienben:
I agree with you @gretzky. We definitely don't want usage of Docker to become compulsory for anyone.

My understanding was that this wouldn't be that though. Maybe I'm mistaken?

Perhaps a plan B could be to put most of this inside the docs? A Docker.md explaining how to set up Docker for dev and production/deployment purposes within an RBP project.

I don't feel very strongly about this either way.

gretzky:
I always think docs are the best first step-- anyone that wants Docker can find instructions, and then usually if enough people are using it and issues start to open, it can become a bigger feature. That's my 2 cents

gretzky:
I'm going to close this now because it's out of date -- but please feel free to open an RFC issue to further the discussion here!

PR #2293, opened 2018-07-24
Poc react testing library

Mensae:
#### Replacing Enzyme with react-testing-library

As I am not that happy with Enzyme and there is yet no complete React 16 support I thought "Hey, let's give [react-testing-library](https://github.com/kentcdodds/react-testing-library)' a try, heard good things about it". And I have to say, I really like it.
I like for example that you can use normal DOM selectors because tests work with actual DOM nodes. This makes it also that tests are closer to the 'real' world. It has, hardly, any learning costs and does not require any setup.

It proofed its point already in the case of ```containers/RepoListItem/IssueIcon``` that decorates ```components/IssueIcon```.
One of the test cases for ```containers/RepoListItem/IssueIcon``` state that it 'should adopt any attribute'. This test passed by checking the ```props``` with Enzyme so, yes, the property exists in ```props``` but ```components/IssueIcon``` did not pass them.

For me it's a keeper. Please tell me your thoughts

[react-testing-library](https://github.com/kentcdodds/react-testing-library)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,7 @@
 const fs = require('fs');
+const path = require("path");
 
-const prettierOptions = JSON.parse(fs.readFileSync('./.prettierrc', 'utf8'));
+const prettierOptions = JSON.parse(fs.readFileSync(path.resolve(__dirname, '.prettierrc'), 'utf8'));
 
 module.exports = {
   parser: 'babel-eslint',
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,6 +1,6 @@
 # React Boilerplate
 
-Before opening a new issue, please take a moment to review our [**community guidelines**](https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md) to make the contribution process easy and effective for everyone involved.
+Before opening a new issue, please take a moment to review our [**community guidelines**](https://github.com/react-boilerplate/react-boilerplate/blob/master/CONTRIBUTING.md) to make the contribution process easy and effective for everyone involved.
 
 Please direct redux-saga related questions to stack overflow:
 http://stackoverflow.com/questions/tagged/redux-saga
@@ -13,8 +13,8 @@ https://github.com/react-boilerplate/react-boilerplate/issues?q=is%3Aissue+is%3A
 
 ## Issue Type
 
-- [ ] Bug (https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md#bug-reports)
-- [ ] Feature (https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md#feature-requests)
+- [ ] Bug (https://github.com/react-boilerplate/react-boilerplate/blob/master/CONTRIBUTING.md#bug-reports)
+- [ ] Feature (https://github.com/react-boilerplate/react-boilerplate/blob/master/CONTRIBUTING.md#feature-requests)
 
 ## Description
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,6 +1,6 @@
 ## React Boilerplate
 
-Thank you for contributing! Please take a moment to review our [**contributing guidelines**](https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md)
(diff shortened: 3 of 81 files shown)


julienben:
Wow lots of work here! Thanks for sharing!

At first glance, this looks pretty good to me but I'm not sure what exactly is justifying a switch from Enzyme to this. That being said I'm pretty un-opinionated when it comes to this stuff so maybe don't listen to me.

Some points to discuss/change:
- Looks like you have a couple of failing tests but it seems trivial to fix.
- I understand this is a POC but you'll want to update the documentation at some point. `component-testing.md` and `introduction.md` at least.
- Can we also take advantage of this to remove `react-test-renderer`? According to `dependency.md`, it was added because of `enzyme`.
- Did you run the full regression testing? It's described in `dependency.md` and quite important here since it involves changes to the sample app and generators.

Mensae:
Thanks for your comments @julienben 

> Looks like you have a couple of failing tests but it seems trivial to fix.

Ah, yes, forgot about that. ```./internals/scripts/generate-templates-for-linting``` is generating templates with the French language setting. This makes that snapshots differ on Travis. I actually removed this from my personal app.
Any thoughts on this? I'm afraid that it wont be so easy as just changing the language to German there.

> I understand this is a POC but you'll want to update the documentation at some point. component-testing.md and introduction.md at least.

Indeed I will, but I thought lets start a dicussion first, as I like coding more than documentation :-)

> Can we also take advantage of this to remove react-test-renderer? According to dependency.md, it was added because of enzyme.

I still use ```react-test-renderer```  in case of for example ```injectReducer.test.js``` where we use a null component. In this case ```react-testing-library``` does not help

> Did you run the full regression testing? It's described in dependency.md and quite important here since it involves changes to the sample app and generators.

I did :-)

gretzky:
Thanks for all of this work, @Mensae! I agree that I'm also not happy with the way Enzyme has been handled. 

I also agree with @julienben on not being sure about making a switch. I definitely think it's worth exploring other, more well-known options as well (e.g. [ava](https://github.com/avajs/ava)) just because of maintainability / future-thinking.

Thoughts? I'm also open to starting a new branch even with master for react-testing-library and advertising it on Spectrum to get thoughts from users.

Mensae:
> I'm also open to starting a new branch even with master for react-testing-library and advertising it on Spectrum to get thoughts from users.

I'd appreciate it @gretzky

gretzky:
Hey @Mensae, sorry for the radio silence. I just moved to a new state and started a new job so it's been hectic. If you're still interested in extending this, I just need the CI to pass.

Mensae:
Hi @gretzky, Ah, I understand. No problem at all :-) and congrats with your new job.

Yes I am, I'd rather not have to maintain my own fork :-). I have a bit of a trouble to get it passing the CI test though. Like I mentioned to @julienben:

```./internals/scripts/generate-templates-for-linting``` is adding lines like ```...RbGeneratedComponent...``` to existing language templates and generating a French language template. This makes that snapshots differ on Travis and so the tests fail.

I'm having the same issue with my own [website](https://github.com/Mensae/portfolio-react-boilerplate), with enzyme, I removed the ```node ./internals/scripts/generate-templates-for-linting``` line in ```.travis.yml``` to get it pass.

jwinn:
@Mensae Nice work on all this. I agree with @gretzky we should properly evaluate any switch from enzyme. I'll be looking into this over the next week or so for a project I'm doing and provide my thoughts some time next week. Thanks again!

julienben:
@gretzky @jwinn 

Any thoughts on reconsidering the switch to `react-testing-library` proposed by @Mensae?

Please see #2385. After the update to `styled-components` v4, the lack of support for `forwardRef` in `enzyme` forced us to use `enzymeFind` from `styled-components/test-utils` to make our tests work with both libs.

This is how our tests changed as a result:

image

Not super easy to grok IMO...

It also feels like `react-testing-library` has become the popular/default alternative to `enzyme` and many people learn React testing from the creator's courses.

What do you think?

julienben:
@Mensae What do you think of updating this branch so we can merge it into `dev` and include it in the next release? Do you have the time/availability/desire to work on it or would you rather someone else took a stab at it?

Mensae:
@julienben No problem at all. I think I can make some time for it. What's your planning for a next release?

julienben:
No idea yet... I'm assuming we'll want to do a 3.8 release when react 16.7.0 lands.

gretzky:
^ agreed

Mensae:
It was more convenient for me to create a new branche. So here is a new [PR](https://github.com/react-boilerplate/react-boilerplate/pull/2452) :-)

julienben:
Closing in favor of #2452.

Issue #2288, opened 2018-07-20
React Boilerplate Electron Version


mjangir:
This is really not an issue but, I had no idea where to put this. I searched a lot for turning this awesome boilerplate into electron application and finally decided to do changes at my own:

For people:

**Who want to add electron support in their existing application:**
[https://medium.com/@mjangir70/turn-the-famous-react-boilerplate-into-electron-desktop-application-68d91dce8d3a](https://medium.com/@mjangir70/turn-the-famous-react-boilerplate-into-electron-desktop-application-68d91dce8d3a)

gretzky:
Awesome, @mjangir! Great write up. Normally we like to keep this issue channel clean, but I think I'll leave this here for a little bit since it is a great write up and hopefully more people will see it. 

If you want to, feel free to open a PR and include a link to this post in the docs somewhere! Also, feel free to post about it on [Spectrum!](https://spectrum.chat/react-boilerplate).

julienben:
Yup in the long term the best way to do this is probably via a link in the docs.

xeoneux:
I was looking exactly for this! Please add it to the docs 😃

mjangir:
Added complete documentation on creating an electron desktop application out of this boilerplate. Just raised a pull request for that. Kindly review.

gretzky:
Thank you @mjangir!

PR #2306, opened 2018-08-05
add SSR example to docs

gretzky:
super quick/easy change. i added an SSR example to the table of contents so its right there if anyone goes looking for it.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -34,6 +34,7 @@
   - [routing](js/routing.md)
 - [Maintenance](maintenance)
   - [Dependency Update](maintenance/dependency.md)
+  - [SSR Example](https://github.com/gretzky/react-boilerplate-ssr)
 
 ## Overview
 


jwinn:
Should the PR be merged to dev, then to master, or is dev in an unknown state? (Trying to get back up-to-speed as I've been away a while)

gretzky:
Dev is, I think (correct me if wrong @julienben) in a good state. I figured a minor doc change to master regarding SSR wouldn't hurt since we get asked about it pretty often 😅

jwinn:
I don't disagree with it going into master. I was under the impression that we wanted `dev` to be in-sync and ahead of `master`. So the process was: changes are made to `dev`, then propagated to `master`, with any semantic version and changelog updates, where required, made as well--in this case being unnecessary.

My initial impetus for the query, though, was I wasn't sure on the state of `dev`--believing it in a good state, but I wasn't completely sure and wanted to verify. Are we still using gitter to have these kinds of discussions, or is there somewhere else we can chat? :smile:

julienben:
@jwinn We do have a clean dev branch now.

@gretzky I think a link in the TOC is a bit confusing. People click there expecting to remain on the same page, not be taken to a different repo. Could the TOC item link instead to a short paragraph in README.md which briefly explains your SSR project and links to the repo from there?

About merging this directly to master: Fine by me so long as we rebase dev on top of master right away to keep the history clean. Alternatively, we could merge this into dev and do a small 3.6.1 release.

julienben:
It might also be the perfect opportunity to add a link to @mjangir's electron version of RBP (#2288)

julienben:
Closing in favor of #2372.

Issue #2297, opened 2018-07-28
Not throwing an error


joshuaalpuerto:
## Issue Type

- [x] Bug (https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md#bug-reports)

## Description
`console.log` not displaying error.

I added an error under `app/containers/HomePage/actions.js`
```js
/*
 * Home Actions
 * ...
 */

shouldThrowError // <---line 20
```

This should throw an error on `console.log` but it doesn't.

### React Boilerplate v3.4.0
I went to `releases` tried to download version `v3.5.0` but it gives me `3.4.0`.

Anyway, I added the same code and it does throw an error which should be.

<img width="1616" alt="screen shot 2018-07-28 at 7 08 48 pm" src="https://user-images.githubusercontent.com/7126484/43356022-5b92a914-929b-11e8-83f3-5b3176cd1150.png">

I know this can be fix by simply `eslint --fix .` but the thing is this should throw an error. It's quite hard to debug especially if it's just **white** page.

## Steps to reproduce
[https://github.com/joshuaalpuerto/rbp-repro-error-swallowed](https://github.com/joshuaalpuerto/rbp-repro-error-swallowed)

1.  Clone the repository with `git@github.com:joshuaalpuerto/rbp-repro-error-swallowed.git`
2. `yarn`
3. `yarn start`

# Versions
- React-Boilerplate 3.6.0
- Node version 8.9.0 & 10.7.0
- Npm version 5 & 6
- Browser: all browsers

joshuaalpuerto:
It seems `react-loadable` swallowing the error. 

Importing the component directly
```js
import HomePage from 'containers/HomePage/Loadable';

//to

import HomePage from 'containers/HomePage';
```

Will throw the error on `console.log`

It seems there a [PR](https://github.com/jamiebuilds/react-loadable/pull/129/files) exist for this issue.

joshuaalpuerto:
**[UPDATE]**

It seems already merge but the `5.4.1` hasn't been released yet by `react-loadable`. 

**SOLUTION**
Downgrade to `5.3.0` Let me know if this is something I can PR but I think `react-loadble` should release the fix.

Cheers!

joshuaalpuerto:
Hi @julienben,

Thanks for the hard word by the way.

> indicating that there's a 5.4.1 release for react-loadable?

[...]

PR #2324, opened 2018-08-29
Babel 7

julienben:
Babel 7 was recently released so I gave a shot at upgrading it. Followed the advice from the official upgrade guide [here](https://babeljs.io/docs/en/v7-migration). Also, stage presets were removed so I used [babel-upgrade](https://github.com/babel/babel-upgrade) to make sure we had the relevant plugins for stage 0 (as we had until now).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,5 @@
 const fs = require('fs');
-const path = require("path");
+const path = require('path');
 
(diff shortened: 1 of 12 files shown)


julienben:
Fixed the base to be 'dev' instead of 'master' and added a a fix to #2297.

Issue #2172, opened 2018-04-01
Generate container doesn't use proper reducer names everywhere


Floriferous:
When using `yarn generate` to create a container, it appears to wrongly generate the name of the reducer in `injectReducer` and `injectSaga`.

Steps to reproduce:
* run `yarn generate`
* container
* Stateless Function
* name: `TestContainer`
* Yes to all questions

At this point in the `TestContainer` folder, the selectors use `testContainer` (proper camelCase), whereas the `injectReducer` from `index.js` uses `testcontainer` (all lowerCase).

This causes the component to crash on render with the following (horrible) error:

```
ReactCompositeComponent.js:744 Uncaught TypeError: Cannot read property '_currentElement' of null
    at ReactCompositeComponentWrapper._updateRenderedComponent (ReactCompositeComponent.js:744)
    at ReactCompositeComponentWrapper._performComponentUpdate (ReactCompositeComponent.js:723)
    at ReactCompositeComponentWrapper.updateComponent (ReactCompositeComponent.js:644)
    at ReactCompositeComponentWrapper.receiveComponent (ReactCompositeComponent.js:546)
    at Object.receiveComponent (ReactReconciler.js:124)
    at ReactCompositeComponentWrapper._updateRenderedComponent (ReactCompositeComponent.js:753)
    at ReactCompositeComponentWrapper._performComponentUpdate (ReactCompositeComponent.js:723)
    at ReactCompositeComponentWrapper.updateComponent (ReactCompositeComponent.js:644)
    at ReactCompositeComponentWrapper.receiveComponent (ReactCompositeComponent.js:546)
    at Object.receiveComponent (ReactReconciler.js:124)
```

Edit: I think these are the offending lines that should have a matching case:

* https://github.com/react-boilerplate/react-boilerplate/blob/5b58dd97d0e9dc39799c345f1bcbf1333bde92b1/internals/generators/container/stateless.js.hbs#L71
* https://github.com/react-boilerplate/react-boilerplate/blob/5b58dd97d0e9dc39799c345f1bcbf1333bde92b1/internals/generators/container/selectors.js.hbs#L6

julienben:
Thanks for reporting! This was already fixed in #2055 though. It will be released to master in v3.6 so, in the meantime, you can make the changes directly to the generator files in your project.

Issue #2303, opened 2018-08-03
Generate container doesn't use proper reducer names everywhere (not merged in 3.6)


anwfr:
Hello,

I reproduce #2172 problem, which was supposed to be fixed in 3.6 by PR #2055.
Looks like this PR was merged in dev branch, but was not integrated in 3.6.

Indeed we still have lowercase reducer key (instead of camelCase) for stateless containers:
> const withReducer = injectReducer({ key: '{{ lowerCase name }}', reducer });
https://github.com/react-boilerplate/react-boilerplate/blob/v3.6.0/internals/generators/container/stateless.js.hbs#L71

My suggestion is to merge #2055 again for next version.

gretzky:
Great catch, thank you!

julienben:
This will be merged in #2325 soon. Closing. Merci @anwfr!

PR #2325, opened 2018-08-29
Fix generator case

julienben:
Cherry picked a commit from #2055 to fix issues #2172 and #2303.


jwinn:
For some reason GH isn't updating the status of the completed travis pr process. Otherwise I'd approve and merge.

julienben:
@justingreenberg Is there a setting to cancel queued builds once a newer version of a branch or PR is pushed? There is in Travis and it makes things a lot faster but I don't have access to the Appveyor config for RBP.

julienben:
Thanks!! I enabled "Rolling builds". Cryptic name but this is how it's described in their docs:

> “Rolling builds” are great for very active OSS projects with lengthy queue. Whenever you do a new commit to the same branch OR pull request all current queued/running builds for that branch or PR are cancelled and the new one is queued. Other words, rolling builds make sure that only the most recent commit is built.

Sounds like what we need.

PR #2326, opened 2018-08-29
Upgraded ESLint to 5.4.0 + other dependencies

julienben:
Nothing crazy. Just trying to keep things up to date.

I've been running all these upgrades in my personal projects for a while with no issues.

ESLint rule updates:
- **react/destructuring-assignment**: New rule. Disabled it by default. It expects you to ALWAYS destructure `this.props.whatever` or `this.state.whatever` using `const {whatever} = this.props;`. Overkill IMO.
- **no-unused-vars and react/jsx-uses-vars**: Explicitly enabled no-unused-vars and added jsx-uses-vars since it does not detect variables used in JSX anymore.
- **jsx-a11y/label-has-for**: Deprecated. Errors that can't be fixed if you use custom components like `<Input>` or `<Label>`
- **jsx-a116/label-has-associated-control**: Similar to above but allows you to specify custom component names so kept it.


jwinn:
LGTM

PR #2328, opened 2018-08-29
Changing to use current and LTS versions for testing; should remove t…

jwinn:
Changing to use current and LTS versions for testing. Should remove the need to update every time a new node major version comes out or an LTS is EoL.


justingreenberg:
please squash merge and update commit message to follow conventional commit format

Issue #2327, opened 2018-08-29
Getting ready to release the next version


julienben:
Hey everyone,

Our last release 3.6 was already over 2 months ago. We haven't made any serious changes to the architecture or the stack since but I propose releasing 3.7 anyway.

It includes quite a few bug fixes as well as major upgrades for Babel and ESLint.

I think it would be great if we could include @gretzky's work from #2257 if we can finish it up soon-ish.

It's also a good opportunity to update the docs and add details/links about SSR, Electron and Docker as discussed in various issues. Relevant contributors are invited to take care of that in the relevant places (#2306, #2282, #2288).

Thoughts anyone? Have I missed anything?

jwinn:
@julienben I'd like to update our CI to use latest and LTS versions and make sure #2304 is fixed for windows. I'll create a branch for these two items...

jwinn:
Also, what's everyone's thoughts on running `npm audit` (`npm audit fix`), which requires adding `package-lock.json` file?

julienben:
You can create a `package-lock` with `npm i --package-lock-only` but if we're gonna do this, are there any reasons not to switch to `npm` entirely?

jwinn:
I know, wanted to more discuss `yarn` vs. `npm`, but should probably saved that for a later discussion. Wanted to see if `npm audit` had enough of an advantage (more from a maintenance perspective) to switch back.

jwinn:
Agreed. I'm not sure how big of a "change" this is as to whether it requires maintainer buyoff per our guidelines. @mxstbr thoughts?

justingreenberg:
fwiw i support switching back to npm, there were very good reasons to use yarn 2 years ago but npm has really come a long way... yarn workspaces are useful for monorepo/library ecosystem maintenance, but i think new  features (especially adding vuln auditing and `npm ci` for ) will add value to end users of this project, and as mentioned lockfile ensures builds are deterministic

the one caveat is that yarn does seem to still be slightly faster, especially when packages are cached (https://github.com/pnpm/node-package-manager-benchmark).. may be prudent to benchmark before making a final decision

michaelcuneo:
ooh. I like this. I didn't know this about Yarn vs NPM. I'll switch back to NPM.

julienben:
Update:

We've implemented the switch to `npm` and still working on the Babel 6 -> 7 transition (in #2331 especially).

[...]

PR #2329, opened 2018-08-30
Switching from yarn to npm

julienben:
Following our conversation in #2327, I did some quick benchmarking on yarn vs npm (3 runs each, clean node_modules, with both lock files present). Yarn was faster but only by 10-12 seconds.

IMO, this switch still makes sense. First, for the reasons discussed earlier (audit and ci) but also because it just makes our code and docs more consistent. Some of our commands require `yarn` (mostly installing dependencies) and while everything can be done with `yarn`, our docs go back and forth and are overwhelmingly `npm`-oriented.

I took advantage of this to refactor `setup.js`. We were using `parseFloat` to check node and npm versions. This was inaccurate since our requirement is 8.10 and parseFloat will return 8.1. We're now using a semVer comparison library.


jwinn:
Thank you for taking the time to do this. Great work! 👍

A few thoughts/questions:

Should `appveyor.yml` use `npm ci` instead of `npm install`?

Regarding `npm audit`:
* While out of this project's direct control, I get `https://nodesecurity.io/advisories/598` referenced 9 times, related to `tunnel-agent` used by `image-webpack-loader`--necessitating `image-webpack-loader` to be updated with `>=tunnel-agent@0.6.0`.
* Should we think about having `npm audit` as part of the commit pipeline?

Should `docs/general/faq.md` be updated?
* `I'm using Node v0.12 and the server doesn't work?` -- Is this really relevant anymore? If it is, should we change the language to be more along the lines of "We settled on supporting the LTS and Current versions of Node.js for the boilerplate..."
* `Use CI with bitbucket pipelines` -- Should `npm install` be changed to `npm ci`?

julienben:
> I'm using Node v0.12 and the server doesn't work? -- Is this really relevant anymore? If it is, should we change the language to be more along the lines of "We settled on supporting the LTS and Current versions of Node.js for the boilerplate..."

Good catch. The "engines" entry in package.json will take care of warning users on old Node.js versions and we also have version checks before install and before the setup script. IMO, this is more than enough and makes this question irrelevant.

> Use CI with bitbucket pipelines -- Should npm install be changed to npm ci?

Yup. Changing that.

> Should appveyor.yml use npm ci instead of npm install?

That too.

> Regarding npm audit

Can you expand on what you mean by having audit as part of the commit pipeline? Are there other projects that do this? How does it work?

julienben:
(Note that Travis already uses `npm ci` by default.)

justingreenberg:
with respect to `npm audit`, as of v6.4 [`audit` can be configured](https://github.com/npm/cli/pull/31) to fail based on vuln score, eg. `npm audit --audit-level=high` will exit 0 if only low or moderate level vulnerability is detected. this should resolve the issue with current deps, however this would also mean dropping support for npm v5.x so it may not make sense at this time

we can add audit in future release @julienben awesome work

julienben:
Good to know about npm audit! Let's remember for whenever npm v6.4 ships with node LTS.

@jwinn I'll restore appveyor to using npm install in the babel fixes branch.

Issue #2330, opened 2018-09-01
TypeError [ERR_INVALID_ARG_TYPE]


FlorinSenoner:
## Issue Type

- [x ] Bug

## Description
running `npm run extract-intl` throws errors: `TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefined`

![screen shot 2018-09-01 at 19 49 51](https://user-images.githubusercontent.com/23100806/44948570-38cdd480-ae20-11e8-8ab8-d66f823254ef.png)

## Steps to reproduce
* download dev branch
* run `npm install` to install the dependencies
* run `npm run extract-intl`
* get error: `Error: Requires Babel "^7.0.0-0", but was loaded with "6.26.3". If you are sure you have a compatible version of @babel/core, it is likely that something in your build process is loading the wrong version. Inspect the stack trace of this error to look for the first entry that doesn't mention "@babel/core" or "babel-core" to see what is calling Babel.`
* run `npm install --dev babel-core@^7.0.0-bridge.0` to add missing dependency (it's declared in "resolutions")
* run `npm run extract-intl` again and get the above error

# Versions

- React-Boilerplate: 3.6.0
- Node/NPM: v10.9.0/6.2.0

julienben:
Thanks a lot for reporting!

Nice catch, I'm adding the missing babel-core "bridge" dependency in #2331.

I personally don't have time to look into the extract-intl script bug just yet but in any case, I strongly recommend that you use the `master` branch. The `dev` branch is not ready for release.

If you do stick with `dev` though, it would be awesome if you could share with us any other issues you run into.

gretzky:
Any updates on this? I haven't had time to investigate myself

yuchaosydney:
I just ran this morning, the issue still exists in the dev branch. It turns out to be missing filename parameter in babel-core transform option object.

Wish to contribute to this repo, so created #2362 pr. 

Hope it is all good.

julienben:
@gretzky, the remaining issue here is the extract-intl script. I've started a fix for it in #2331 and I see a separate PR claiming to fix it entirely (woot!). Will review properly ASAP and make sure this is all closed properly.

julienben:
Closing this as the bug only exists in `dev`, which is currently a WIP towards the next release and not meant to be used for new projects.

PR #2331, opened 2018-09-02
Babel fixes

julienben:
Not ready yet. Just opened this to work on #2330 and potentially other issues with the babel upgrade.

Specifically, jest deps still require babel 6 and it's causing some instability. Looking into it.


jwinn:
Still receiving `TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefinedError transforming file: app/...`, but no babel error (#2330) and the translation files are created.

julienben:
Is this in reference to #2330? You got this error when running the extract-intl script with this branch checked out? 

FYI, this branch isn't trying to fix the issue with the extract-intl script. I just added here a small babel fix (which I later remembered I'd already added in #2329 anyway). My plan is to use this branch to continue adding fixes to the updated Babel setup. Some packages are still dependent in Babel 6.

jwinn:
Ok. 👍  Yes, I was referencing #2330, as I wasn't sure, so just checked `npm run extract-intl` and got the errors and empty translation files, but no `babel`-related error mentioned in #2330.

julienben:
Link to the relevant babel plugin: https://github.com/yahoo/babel-plugin-react-intl

julienben:
Alternative solution: Use this new plugin instead: https://github.com/akameco/extract-react-intl-messages

appveyor.yml
```diff
@@ -24,8 +24,10 @@ install:
   - choco install -y googlechrome --ignore-checksums
   # Install the latest stable version of Node
   - ps: Install-Product node $env:nodejs_version
-  - npm install npm -g
-  - npm ci
+  - npm install
+
+# TODO: Switch to using npm ci instead of install
+# once Node LTS ships with npm 5.7 (with ci)
```
piuccio: node 8.12 (lts) comes with npm 6.4

internals/scripts/extract-intl.js
```diff
@@ -94,44 +92,34 @@ for (const locale of locales) {
...
-const extractFromFile = fileName => {
-  return readFile(fileName)
+const extractFromFile = fileName =>
+  readFile(fileName)
     .then(code => {
       // Use babel plugin to extract instances where react-intl is used
-      const { metadata: result } = transform(code, { presets, plugins });
+      let messages = [];
+      try {
+        const output = transformSync(code, { presets, plugins });
```
piuccio: In my code I'm using `await transform(code, { presets, plugins, filename: fileName });` note the `await` instead of sync and the additional `filename` that prevents some warnings

appveyor.yml
julienben: Thanks for the notice! It wasn't in 8.11. Will check if Appveyor provides 8.12 already and update accordingly!

internals/scripts/extract-intl.js
julienben: `filename` made it into the new version of the script merged in #2350 ;)

julienben:
@piuccio I like the idea of breaking babel out of package.json. Mostly cause package.json is just too long with all these configs in it and I don't think it does much for us to keep it all in there. We already had to break out ESLint for the prettier integration. Might do the same for Jest and Babel in a different branch.

Closing this PR in favor of the already merged #2350.

PR #2350, opened 2018-09-24
Fix extract-intl for babel 7

fanixk:
Fixes extract-intl script to work with Babel 7. Uses async version of transform.
Based on #2331


gretzky:
Thank you @fanixk!

PR #2072, opened 2017-12-18
Migrate from pre-commit to husky

avdeev:
Related with #2056 

Work correctly

```bash
Mac-mini-Alexey:react-boilerplate alexeyavdeev$ git commit -m "feat: migrate from pre-commit to husky"
husky > npm run -s precommit (node v8.9.0)

 ↓ Running tasks for *.js [skipped]
   → No staged files match *.js
```

PR #2374, opened 2018-10-09
Dependency upgrades and fixes

julienben:
- Fixes
- Linting/prettier
- Dependency upgrades
- Switch Appveyor to `npm ci`
- More incoming


julienben:
@gretzky I agree with you about the high number of Babel plugins. It's the list we reached when going through Babel's official upgrade to 7 instructions/script. Not sure how to slim down the list just yet but I've at least moved the Jest and Babel configs to their own files to get `package.json` down from 300 to 200 lines.

gretzky:
I'll take a look at the babel plugins

satishjhanwer:
I saw that dev branch doesn't have husky as a dependency anymore and its being remove while merging this PR. Husky was added in [ #2072](https://github.com/react-boilerplate/react-boilerplate/pull/2072)

@gretzky @julienben, Let me know your thoughts on [commit-lint](http://marionebl.github.io/commitlint/), I see we are not following proper commit messages which makes changelog generation a hard job.

julienben:
@satishjhanwer What's your point regarding Husky? Does it need to be restored as a dependency?

Re:commit-lint, I wouldn't be against having it for our work here in the RBP repository itself but it might be too much to force it on all the people who clone the repo for their own use. 

This is a problem that arises out of our being a project that needs to be cloned rather than one that can be installed as a dependency...

PR #2375, opened 2018-10-09
Migration from react-router-redux to connected-react-router

julienben:
`react-router-redux` is deprecated and hasn't seen an update in 2 years so we're migrating to the newest and shiniest, `connected-react-router` ([more here](https://github.com/supasate/connected-react-router)). Please review extensively as this isn't a trivial change.

@gretzky has suggested a migration to `reach-router` and this isn't in any way a contradiction to that. Just trying to stay up-to-date for now as a full router switch would be a much larger endeavor.

Also note that this PR is still incomplete. Just trying to see what people think of it. Once we're happy I'll do some regression testing and update the docs.


app/reducers.js
```diff
@@ -2,49 +2,22 @@
...
  * Creates the main reducer with the dynamically injected ones
  */
 export default function createReducer(injectedReducers) {
-  return combineReducers({
-    route: routeReducer,
-    global: globalReducer,
-    language: languageProviderReducer,
-    ...injectedReducers,
-  });
+  return connectRouter(history)(
```
justingreenberg: one minor nit: assigning some variables here may help readability... when first reviewing it took me a few seconds to find the `connectRouter` integration

```js
export default function createReducer(injectedReducers) {
  const mergeWithRouterState = connectRouter(history);
  const rootState = combineReducers({
    global: globalReducer,
    language: languageProviderReducer,
    ...injectedReducers,
  });
  return mergeWithRouterState(rootState);
}
```

unrelated, but it might make sense to assign a default parameter for `injectedReducers`...

```js
export default function createReducer(injectedReducers = {}) {
  //...
}
```

or add guard, eg.. 

```js
import isObject from 'lodash/isObject';
import isFunction from 'lodash/isFunction';
import invariant from 'invariant';

export default function createReducer(injectedReducers) {
  invariant(
    isObject(injectedReducers) &&
    Object.keys(injectedReducers) &&
    Object.values(injectedReducers).every(isFunction),
    '(app/reducers) createReducer: `injectedReducers` must be an object of reducer functions'
  )
  // ...
}
```

mxstbr:
> re: build preview, i'm not sure who has access to netlify cc: @mxstbr is it possible to add RBP org?

I've added you (j*k*g@gmail.com) and @gretzky as admins on Netlify, I think it worked?!

gretzky:
Can confirm it worked @mxstbr

gretzky:
Good work @julienben !

app/reducers.js
julienben: Absolutely to both comments! Thanks!

I went for the default param as combineReducers already has an error message if it doesn't receive an object of reducer functions + using our own guard meant providing an empty object at initial store creation, which felt like a contrived complexity.

julienben:
> Assuming the LOCATION_CHANGE (with hot reload) issue is no longer of concern with connected-react-router, this LGTM.

@jwinn I removed the REDUX_DEVTOOLS thing because I couldn't reproduce the issue after the switch. But now that I look into it further, I can't reproduce the issue before the switch either. Can you (or someone else) give me the steps to reproduce it? I'd like to make sure this is indeed fixed by the switch.

PR #2376, opened 2018-10-10
Babel cleanup

gretzky:
This removes the numerous unnecessary babel plugins added by the babel 7 upgrade. Keeps everything necessary / what it existed beforehand as well.


gretzky:
Yup, was an oversight-- thanks! And yep, we're thinking alike about the proposal plugins. 

Also, @mxstbr wants you to get a Twitter 😄

julienben:
Where did he say that?

julienben:
Ok nvm found it https://twitter.com/mxstbr/status/1049918611358400512

Issue #2378, opened 2018-10-10
Postinstall - build:dll - wrong path creation for react-boilerplate-dlls folder


Bastek333:
## Issue Type

Bug

## Description

Postinstall script which is running build:dll with node ./internals/scripts/dependencies.js is throwing an error - as path for creating appropriate folder is wrong. This is only on development environment as on production this script is skipped !!! 

> aem-authoring-ui@1.0.0 build:dll C:\Projekt\aem-dc-view-spike
> node ./internals/scripts/dependencies.js

mkdir: cannot create directory **C:/Projekt/aem-dc-view-spike/C:/Projekt/node_modules/react-boilerplate-dlls**: Not a directory
Building the Webpack DLL...
fs.js:646
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);

**And I did quick checking:**

**dllConfig:**  { [Boolean: false]
  exclude:
   [ 'chalk',
     'compression',
     'cross-env',
     'express',
     'ip',
     'minimist',
     'sanitize.css',
     'react-icons',
     '@opuscapita' ],
  include: [ 'core-js', 'eventsource-polyfill', 'babel-polyfill', 'lodash' ],
  path: 'C:\\Projekt\\node_modules\\react-boilerplate-dlls' }
**outputPath:**  C:\Projekt\aem-dc-view-spike\C:\Projekt\node_modules\react-boilerplate-dlls
**dllManifestPath:**  C:\Projekt\aem-dc-view-spike\C:\Projekt\node_modules\react-boilerplate-dlls\package.json

## Steps to reproduce
 just npm install - or 
node ./internals/scripts/dependencies.js

# Versions

- React-Boilerplate (see `package.json`): "version": "1.0.0"
- Node/NPM: 6.4.1

julienben:
Thanks for filling in the issue template!

I'm afraid I can't reproduce your error though. (We have lots of automated testing so if `npm install` were failing for everyone, it would have been noticed right away.)

This is most likley due to an "incompatibility" between your computer setup and this project. The result is that your output path for the dll file is broken (since `C:\` is repeated twice).

Let's see if we can fix it!

On your computer, could you open `dependencies.js` inside of `internals/scripts/` and replace all three instances of `path.join` with `path.resolve`?

Then rerun `npm install` and report back.

Bastek333:
Thank you for quick answer.

Looks like path.resolve has resolved issue :) - now the script is running fine.

[...]

PR #2379, opened 2018-10-10
Path fix and lodash babel dev

julienben:
Two small changes:

First, a fight to the death between `path.join` and `path.resolve` (per #2378). I think `path.join` wins. Unless it creates another issue I didn't think of?

Second, I added lodash's official [babel-plugin-lodash](https://github.com/lodash/babel-plugin-lodash) to production plugins in Babel. It replaces `import { get } from 'lodash';` with `import get from 'lodash/get';`. I know we already use the latter by default but in case a user forgets to do the same, it saves them from adding an additional 70k to their bundle.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -1,26 +1,33 @@
 <!doctype html>
 <html lang="en">
-  <head>
-    <!-- The first thing in any HTML file should be the charset -->
-    <meta charset="utf-8">
 
-    <!-- Make the page mobile compatible -->
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+<head>
+  <!-- The first thing in any HTML file should be the charset -->
+  <meta charset="utf-8">
 
-    <!-- Allow installing the app to the homescreen -->
-    <meta name="mobile-web-app-capable" content="yes">
+  <!-- Make the page mobile compatible -->
+  <meta name="viewport" content="width=device-width, initial-scale=1">
 
-    <link rel="icon" href="/favicon.ico" />
-    <title>React.js Boilerplate</title>
-  </head>
-  <body>
-    <!-- Display a message if JS has been disabled on the browser. -->
-    <noscript>If you're seeing this message, that means <strong>JavaScript has been disabled on your browser</strong>, please <strong>enable JS</strong> to make this app work.</noscript>
+  <!-- Allow installing the app to the homescreen -->
+  <meta name="mobile-web-app-capable" content="yes">
 
-    <!-- The app hooks into this div -->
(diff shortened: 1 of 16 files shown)


justingreenberg:
@gretzky i think we can keep the full lodash toolchain available, since we are using module imports eg `import isEqual from 'lodash/isEqual'` webpack tree-shaking excludes any unused modules from the final build... so, even with current setup the output bundles would be the same as if we were installing the individual modules from npm, which may not be the most ergonomic solution imo. while i'm also reticent to add anymore plugins, i think in this case it's probably ok since it's just a development dependency

@julienben if we are adding lodash babel transform, let's update existing [lodash module imports](https://github.com/react-boilerplate/react-boilerplate/search?q=lodash) to named exports, eg.

```js
import conformsTo from 'lodash/conformsTo';
import isFunction from 'lodash/isFunction';
import isObject from 'lodash/isObject';
```

can be

```js
import { conformsTo, isFunction, isObject } from 'lodash';
```

since the plugin will transform this to former module import format anyway :)

gretzky:
@justingreenberg makes sense, I'm only asking because we could significantly reduce the size of modules by removing the full lodash toolchain and just adding the modules we need-- lodash is ~530k (64k gzipped) and all of the modules we currently use here combine to 181k (8k gzipped).

justingreenberg:
@gretzky totally agree regarding DX. unfortunately devs using RBP will be still have to install the full lodash package since it's part of the extended graph (lodash is also a dependency of our other deps like webpack, babel, enzyme, plop etc.)

julienben:
The way we imported `lodash` modules until know (`import xxx from 'lodash/xxx'`) already ensured that we benefit from tree-shaking. This was only meant as a protection for users who forget to import this way. But now (per @justingreenberg 	) it allows us to make our imports look cleaner and make sure we always benefit from tree-shaking, no matter how we import.

The plugin itself is only used in production so there's no hit on dev performance. (We have some leeway in dev anyway since we just removed a whole bunch of plugins 😅)

My reasoning was based on [this article](https://nolanlawson.com/2018/03/20/smaller-lodash-bundles-with-webpack-and-babel/) where he shows how to reduce the lodash part of a bundle. I didn't add `lodash-webpack-plugin` as it has a few gotchas and should be a per-project decision. I also tried `lodash-es` but that came with its own set of issues.

@gretzky I'll wait for you to let me know if you're still on the fence about this or if it makes sense to you before I merge.

gretzky:
Yeah makes sense to me overall, I'll merge

Issue #2092, opened 2018-01-06
RFC: Remove Immutable.js


justingreenberg:
I hope everyone had a happy and healthy holiday  🎉 back to work

So, one of the biggest complaints maintaniers hear about RBP relates to Immutable.js

There has been a lively discussion recently in #2015. People seem to be pretty split, with strong opinions both ways. The idea of this RFC is not really to evangelize one way or the other, but to start a conversation and provide a platform for maintainers and community at large to weigh in.

Give the thumbs up (👍 ) to vote to REMOVE Immutable.js
Give the thumbs down (:-1: ) to vote to KEEP Immutable.js
As well, please leave any comments below.

Thank you for taking the time to provide feedback  😄

gretzky:
My main take on immutable is that it causes more problems than solves issues, and to me a boilerplate maintains basics (albeit some are opinionated) and immutable isn't a basic imo. I also think that any project that needs/requires it that uses this bp will be able to still re-add it back, since its been here for so long as a use case.

blling:
I think immutable.js is great, the main problem is that we do not find the right way to use it.

JulienMelissas:
The last two projects I've used this on I removed Immutable. The first one was because I didn't feel comfortable using it due to lack of experience. The second one was because the team decided it was not necessary. It has it's place for sure, but I think opt-in (maybe in the form of a nice guide) would be the best way to go.

schettino:
Immutable is totally **optional**. There is a lot of discussion about get it working properly. Whereas its may be convenient to enforce best practices, developers tend to work around it just to make it work. 
Also, writing immutable states with bare metal javascript is not that bad and if you are designing your state properly, you won't find any problem here.

Removing Immutable is a great step a head if this project is about making a standard react setup, many of us just remove it on every single fresh install (I just forked this repo to stopping doing that). Just JS is fine, Immutable also is fine if you like it, and there are a tons of different helpers, including Lodash, that is already included in this boilerplate. So, please, let's get it out of here! Thanks

ralphstodomingo:
Anecdotal: I've never had problems with Immutable. It felt like it clicks. While I'm certain things can be made to work without it, here's what the intro on RBP says, on top:

> A highly scalable, offline-first foundation with the best developer experience and a focus on performance and best practices. 

[...]

PR #2380, opened 2018-10-11
Immer

julienben:
It might take a long time to reach a consensus in #2092 so after reading @mweststrate's [intro to Immer](https://hackernoon.com/introducing-immer-immutability-the-easy-way-9d73d8f71cb3) I gave a shot at migrating the sample app just so we can see what that would look like.

Immediate benefits:
- Immutable data without the need to use Immutable's custom data types
- Essentially no learning curve
- Removes over 50k from our builds
- Trivial to remove for anyone who doesn't want it. We could even add it as an option in the generators.

Thoughts? Opinions?

NOTE: Tests will fail. This is a WIP until we can agree on a direction.


gretzky:
I guess I'm a bit confused as to why there's an immediate need to move away from Immutable if there doesn't seem to be a consensus on which direction we should head in? 

I don't use any immutable library in any of my own projects so I can't say much-- I only have one opinion: I don't agree about the 'no learning curve' piece. The syntax of this library seems pretty different from immutable (i.e `draft`?) and for anyone that uses this boilerplate as a part of their daily development, that's going to be a big difference.

Mostly I just don't want to assume our users learning curve. Although this is a production-ready boilerplate, I know it's used for reference quite a bit and Immutable is a big part of this project, it changes the way someone learns/builds/thinks about a redux store. I have a soft spot for people who learn React development from this boilerplate (it _really_ helped me back in the day) and experiencing Yet Another JS Library™ fatigue is real.

Every other bullet point I like and agree with tho 🎉 

If I'm the minority here, I have no problem going with this solution. It's tough because we aren't gaining a whole lot of traction with RFCs and immutability is a huge topic 🤷‍♀️

mxstbr:
I didn't know there were RFCs—ping me Twitter when you have important RFCs and I'll share them with the wider React community? If there is one for Immutable.js, tweet it and I'll RT

julienben:
It's #2092 but it may be better to start with a fresh one?

julienben:
@gretzky If your main concern is the learning curve, maybe we should skip my next PR? I won't divulge what's in it but let's just say that it includes renaming the project to `angular-boilerplate` 😜😂

More seriously, I'm not trying to express a need to move to Immer, immediate or not. We can totally release 3.7 with Immutable and then any changes to store management can wait for a big 4.0 until we've had more time to discuss and agree.

At that time though, I would probably at least be suggesting a move away from Immutable.js. For several reasons that I'm happy to list (although I believe most of them have been discussed in #2092).

Btw, the main one is probably the huge learning curve associated with Immutable. IMO, a move to Immer would lower the overall learning curve of the project. (Take a quick look at the Immer intro and you'll see it has a tiny API.)

I also completely agree with you about this project being used as a reference and I also learned a TON from it. Seeing as the overall React community's consensus seems to be going towards seeing Immer as "the best way to do immutable state", perhaps "being a reference" means it would make sense for us to follow that trend. (Unless I'm not making sense?)

All that being said, I don't have a strong preference between Immer and another solution. @justingreenberg's suggestion of using [redux-immutable-state-invariant](https://github.com/leoasis/redux-immutable-state-invariant) to raise errors if data is mutated in dev is also a good one IMO (although it would be nice it we could find a maintained version).

In any case, I'm mostly trying to stimulate the discussion and as I said this can totally wait til the next version. :)

(Sorry I wrote so much!!)

gretzky:
Happy to merge this whenever appropriate

julienben:
> Happy to merge this whenever appropriate

@gretzky You're happy to switch to Immer? If so, I'll finish this branch soon.

gretzky:
@julienben yeah I'd say so-- I still don't use immutability but I see the value. Removing it would be a bigger task imo. Take your time!

julienben:
@gretzky I believe this is ready for review!

gretzky:
All set!

PR #2384, opened 2018-10-15
replaced immutablejs with immer in tests and internals

robertaird:
Apologies for the duplicate pull request, I wanted to fix the botched history.

Not to step on anyone's toes, but I love this boilerplate and I definitely want to use it with immer on my next project. I saw this branch and I wanted to see how it would look as far as testing goes, as well as having it implemented in any components we scaffold with reducers, so I went ahead and replaced the rest of the instances of ImmutableJS within the code. I haven't touched documentation, but I thought it might save you the bit of time it takes to refactor, should you all decide to go this route.

All Jest tests are passing.

### DIFF
--- a/app/containers/App/tests/reducer.test.js
+++ b/app/containers/App/tests/reducer.test.js
@@ -1,19 +1,20 @@
-import { fromJS } from 'immutable';
+import produce from 'immer';
 
 import appReducer from '../reducer';
 import { loadRepos, reposLoaded, repoLoadingError } from '../actions';
 
+/* eslint-disable default-case, no-param-reassign */
 describe('appReducer', () => {
   let state;
   beforeEach(() => {
-    state = fromJS({
+    state = {
       loading: false,
       error: false,
       currentUser: false,
-      userData: fromJS({
+      userData: {
         repositories: false,
-      }),
-    });
+      },
+    };
   });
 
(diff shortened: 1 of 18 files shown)


julienben:
Thanks a lot for this! No toes stepped on. This is a welcome contribution.

I haven't reviewed in depth yet as no decisions have been made regarding a switch to `immer` but thought you might want to know about a couple small things you missed.

First off, the generator in `internals/generators/container/reducer.js.hbs` is incorrect. You shouldn't be returning the state from inside the `produce` function.

Second, the linting of generated components didn't pass so the tests weren't able to run on the CI tools. Take a look at the Travis or AppVeyor configs. You'll see that before running linting and tests, a script called `generate-templates-for-linting` is run to check that the generators produce "valid/clean" code.

If you have time to fix it, great. If not, I'll merge this into the `immersion` branch when relevant and apply the fixes there myself. Thanks again!

robertaird:
Thanks for the feedback! I went ahead and fixed the botched `reducer.js.hbs` generator and brought it in line with the others in the demo repository. I also had a typo in the `selectors.js.hbs` generator and as well as some trailing spaces in the `reducer.test.js.hbs` generator. I ran the `generate-templates-for-linting` script and linting passed on my machine.

PR #2394, opened 2018-10-18
Selector bug fix

julienben:
I noticed that the router state is under the key `router` and not `route`. Not sure when we missed this but not surprising that we missed it since the selector's not actually in use (mostly there for demo purposes AFAIK).

Also added some small dependency upgrades.


jwinn:
@julienben IIRC (and could be completely wrong), the use of `route` (instead of `router`) was related to an earlier `react-router` `react-router-redux`, and subsequent `LOCATION_CHANGE`, related issue/workaround.

justingreenberg:
@gretzky seems like an issue with some deps like enzyme, the build starts failing here:
https://ci.appveyor.com/project/justingreenberg/react-boilerplate/builds/19626822/job/k2t1ioixxrrsvryt#L157

```
TypeError: (0 , _testUtils.enzymeFind) is not a function
```

there also seems to be some outdated snapshots and an issue with SC global styles

no idea why tho since it's passing on travis, i will trigger a new build now

PR #2396, opened 2018-10-19
switched to terser-webpack-plugin from uglifyjs-webpack-plugin

robertaird:
I've been incorporating some changes from the dev branch into a project I built with the previous release, and I ran into this issue at build time: https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/359

Turns out uglifyjs-webpack-plugin is no longer using uglify-es on the backend as it's no longer actively maintained, and they recommend switching to terser. There may be a good reason to stay on Uglify, I honestly haven't dug in too deeply, but this configuration is working on my project and it might save people from some small headaches in the future.

### DIFF
--- a/internals/webpack/webpack.base.babel.js
+++ b/internals/webpack/webpack.base.babel.js
@@ -115,7 +115,7 @@ module.exports = options => ({
   },
   plugins: options.plugins.concat([
     // Always expose NODE_ENV to webpack, in order to use `process.env.NODE_ENV`
-    // inside your code for any environment checks; UglifyJS will automatically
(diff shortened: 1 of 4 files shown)


julienben:
Same! Looks good to me! Thanks for noticing and opening a PR for this @robertaird!

Please just fix the `package.json` conflict with the correct versions:

```
"terser-webpack-plugin": "1.1.0",
"url-loader": "1.1.2",
"webpack": "4.21.0",
```

and update `package-lock.json` accordingly and then we'll merge it!

robertaird:
@julienben Should be all set!

PR #2398, opened 2018-10-19
replace windows line ending with empty string before comparing node version

### LINKED ISSUE #2397: npm run start on the dev branch fails on Windows
# React Boilerplate

Before opening a new issue, please take a moment to review our [**community guidelines**](https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md) to make the contribution process easy and effective for everyone involved.

Please direct redux-saga related questions to stack overflow:
http://stackoverflow.com/questions/tagged/redux-saga

For questions related to the boilerplate itself, you can also find answers on our gitter chat:
https://gitter.im/mxstbr/react-boilerplate

**Before opening a new issue, you may find an answer in already closed issues**:
https://github.com/react-boilerplate/react-boilerplate/issues?q=is%3Aissue+is%3Aclosed

## Issue Type

- [x] Bug (https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md#bug-reports)
- [ ] Feature (https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md#feature-requests)

## Description

I'm getting an error when running `npm run setup` on the dev branch that comes from the `compare-versions` module claiming `v10.10.10` is not valid semver:

image

It appears that some new code that checks the node version is not properly replacing windows line endings coming from stdout. Pull request to follow.

## Steps to reproduce

Run `npm run setup` in Powershell or cmd.exe on windows

# Versions

- React-Boilerplate (see `package.json`): vNext
- Node/NPM: node v10.10.0, npm v6.4.1
- Browser: N/A
- Windows: 10 version 1803 build 17134.345

autechr3:
I'm not sure why the `\r\n\t` case exists. There doesn't appear to be a tab character in the output from `node --version` on my machine. There is, however, a plain windows line ending (`\r\n`) that gets passed into `compareVersions()` because the current regex doesn't replace it.

I just added another condition to the regex. That seems to be the safest option.

Fixes #2397


julienben:
The \r\n\t case was there because of [this stackoverflow thread](https://stackoverflow.com/questions/10805125/how-to-remove-all-line-breaks-from-a-string) (second answer which I now see doesn't match its [source](https://www.textfixer.com/tutorials/javascript-line-breaks.php)). I'm on Mac so wouldn't have seen this error. Thanks for reporting.

autechr3:
In that case, should we just go with the first answer's suggestion and make it  `/\r?\n|\r/gm`?

julienben:
I like the longer version more personally. Just because of readability. And it's not like performance matters here.

julienben:
I didn't notice your stdout.trim() suggestion. Happy to merge that too. Whichever your prefer.

autechr3:
Lets go with that then. If the goal is readability, then we should avoid a regex replace all together if we can.

julienben:
AppVeyor failed on Node LTS. Could be due to random network glitches on their side as often happens. Rebuilding now.

PR #2410, opened 2018-10-25
 import .htaccess file in templates

ngtan:
Please correct me if I'm wrong. There is a tiny distinction between `app/app.js` and `internals/templates/app.js`. Was it missed?


julienben:
Yes indeed. We need this or file-loader adds a .bin extension to .htaccess on build. Thanks for catching it.

PR #2412, opened 2018-10-25
Fix .htaccess bug in template app

julienben:
@gretzky The .htaccess fix from @ngtan is a non-trivial bug fix for anyone using apache. (Without the fix, the file created in build will be .htaccess.bin...) Should we release it straight to master? Feel free to merge this if you agree.


gretzky:
There's a segfault in the CI build, no idea why

julienben:
I restarted the job. No reason for this to happen with the changes in this branch.

PR #2413, opened 2018-10-25
Code splitting with Suspense

julienben:
With [React 16.6](https://reactjs.org/blog/2018/10/23/react-v-16-6.html#reactlazy-code-splitting-with-suspense) we can load components asynchronously using `lazy` and `Suspense`. Negligible difference in bundle size but it's still nice to use native React features instead of an additional dependency.

@gretzky @justingreenberg What do you think? Should we implement this? I'll finish the PR if relevant!

### DIFF
--- a/app/utils/loadable.js
+++ b/app/utils/loadable.js
@@ -1,6 +1,6 @@
 import React, { lazy, Suspense } from 'react';
 
-const loadable = (importFunc, { fallback }) => {
+const loadable = (importFunc, { fallback = null }) => {
   const LazyComponent = lazy(importFunc);
 
   return () => (
--- a/app/utils/loadable.js
+++ b/app/utils/loadable.js
@@ -1,0 +1,13 @@
+import React, { lazy, Suspense } from 'react';
+
+const loadable = (importFunc, { fallback }) => {
+  const LazyComponent = lazy(importFunc);
+
+  return () => (
+    <Suspense fallback={fallback}>
+      <LazyComponent />
+    </Suspense>
+  );
+};
+
+export default loadable;
--- a/app/utils/loadable.js
+++ b/app/utils/loadable.js
@@ -1,11 +1,11 @@
 import React, { lazy, Suspense } from 'react';
 
-const loadable = (importFunc, { fallback = null }) => {
+const loadable = (importFunc, { fallback = null } = { fallback: null }) => {
   const LazyComponent = lazy(importFunc);
 
-  return () => (
+  return props => (
     <Suspense fallback={fallback}>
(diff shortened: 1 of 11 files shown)


gretzky:
Agreed w/ the changes but I think we should [wait for 16.7](https://twitter.com/sophiebits/status/1055127361367076864)

app/utils/loadable.js
```diff
@@ -0,0 +1,13 @@
+import React, { lazy, Suspense } from 'react';
```
gretzky: This is rly cool

julienben:
Will 16.7 bring anything that makes Suspense obsolete?

gretzky:
No, but it (I think) will bring suspense more into the mainstream, and there's always the possibility of change. I think it just might make sense to include this as part of an upgrade we'll need to do anyway once its released.

julienben:
Sounds good. I'll keep this open in the backlog until then.

danielnass:
Hi folks!

As I know, Suspense stuff doesn't work with SSR yet.

julienben:
We don't support SSR but this change won't be merged anytime soon anyway.

Mostly we're waiting for all the new features of React to become a bit more popular. I'm also fairly certain that `utils/loadables.js` is working for this example but won't work in a number of edge cases.

julienben:
(NOTE: `utils/loadable.js` was fixed and show now work in all situations.

gretzky:
Howdy, I think it's time we revisit this-- @julienben what do you think?

gretzky:
Makes sense to me!

joey-siftlogic:
Generating a container like so

    npm run generate container

will generate a `Loadable.js` with this content

    /**
    *
    * Asynchronously loads the component for Form
    *
    */

    import loadable from 'utils/loadable';

    export default loadable(() => import('./index'));

So I assumed it is implemented yet in the generators... ? or is it just not necessary for the generator to use the fallback LoadingIndicator component as HomePage/Loadable.js have.

nvm me if that was the case.

julienben:
The loading indicator is only in the sample app as an example. It's too opinionated to be part of the "clean" app. So the generator looks exactly as you posted it.

I'll add some details about how to use a fallback component in the docs. Just having a very intense work period and will get to it soon after.

julienben:
@gretzky I updated the docs and rebased this branch on top of the branch from #2579. That means if you approve and merge #2579, this one can be merged into dev immediately after (with no conflicts).

gretzky:
All set @julienben

julienben:
Turns out it wouldn't let me merge it as is... Can you please re-review?

Issue #2415, opened 2018-10-25
Suggestion to generate a passing test with components and containers


mjhost:
# React Boilerplate

Before opening a new issue, please take a moment to review our [**community guidelines**](https://github.com/react-boilerplate/react-boilerplate/blob/master/CONTRIBUTING.md) to make the contribution process easy and effective for everyone involved.

Please direct redux-saga related questions to stack overflow:
http://stackoverflow.com/questions/tagged/redux-saga

For questions related to the boilerplate itself, you can also find answers on our gitter chat:
https://gitter.im/mxstbr/react-boilerplate

**Before opening a new issue, you may find an answer in already closed issues**:
https://github.com/react-boilerplate/react-boilerplate/issues?q=is%3Aissue+is%3Aclosed

## Issue Type

- [ ] Bug (https://github.com/react-boilerplate/react-boilerplate/blob/master/CONTRIBUTING.md#bug-reports)
- [x] Feature (https://github.com/react-boilerplate/react-boilerplate/blob/master/CONTRIBUTING.md#feature-requests)

## Description

Thanks for the great boilerplate. I've used it in the last year and I found out a nice variant for the generated component/container tests. It basically spies on the console, tries a shallow render and checks that there's no call to console.error. I would like to send a pull request of this if you are interested.

gretzky:
Hi @mjhost, I'd love to see your implementation! Open a PR at your leisure and we can take a look.

Thank you for asking and filling out the template!!

julienben:
Will close this since there's now a corresponding PR.

PR #2421, opened 2018-10-26
Generate passing tests for components/containers

mjhost:
## React Boilerplate

- [ ] Pull request has tests (we are going for 100% coverage!)

as proposed in  #2415 here is my suggestion to generate a basic passing test that asserts that console.error has not been called in a shallow rendering.

### DIFF
--- a/internals/generators/component/test.js.hbs
+++ b/internals/generators/component/test.js.hbs
@@ -10,4 +10,7 @@ describe('<{{ properCase name }} />', () => {
     shallow(<{{ properCase name }} />);
     expect(spy).not.toHaveBeenCalled();
   });
+  it('Expect to have additional unit tests specified', () => {
+    expect(true).toEqual(false);
+  });
 });
--- a/internals/generators/component/test.js.hbs
+++ b/internals/generators/component/test.js.hbs
@@ -1,11 +1,13 @@
-// import React from 'react';
-// import { mount } from 'enzyme';
+import React from 'react';
+import { shallow } from 'enzyme';
 // import { enzymeFind } from 'styled-components/test-utils';
 
-// import {{ properCase name }} from '../index';
+import {{ properCase name }} from '../index';
 
 describe('<{{ properCase name }} />', () => {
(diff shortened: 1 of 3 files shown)


mjhost:
It's just something that i had that helped me during the development. 

1. I don't really know what there should be against generation of a small passing test. You won't stop anyone from writing deeper tests, and it will need to be expanded to test all branches.
2. this one is just testing that there are no console errors, aka it catches `"Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports."` for example

mjhost:
by the way @gretzky wanted to see the implementation :)

julienben:
1. You lose the very explicit notice of the failing test which says "Expect to have unit tests specified". If you have a passing test by default, devs are less likely to write additional tests IMO.

2. I personally like that it fails on React warnings but I'm not sure how popular that would be. @gretzky You have an opinion on this? Or this PR as a whole?

@mjhost I think your test makes sense as an addition to the current test, not a replacement. We should have both yours and another called "Expect to have **additional** unit tests specified" which compares `false` to `true` as it does now. Perhaps it could even have a comment with the link to the test documentation in this repo. And then devs are really encouraged to write more tests and learn about it. What do you think?

(BTW, please take a look at your diff. All the indentation in `stateless.js.hbs` was removed.)

mjhost:
sorry about the indentation, prettier just walked in and i didn't realize. 
about the proposal of making it an addition to the existing test, it makes sense. I'll commit accordingly.

julienben:
Yeah `prettier` likes to invite itself where it's not welcome. I wasn't able to fix it with `.prettierignore` if I remember correctly. Instead, you can add: 

```json
"[handlebars]": {
  "editor.formatOnSave": false,
}
```

to your VS Code config (if that's your IDE of choice).

julienben:
Thanks @mjhost

PR #2452, opened 2018-11-08
Replacing Enzyme with react-testing-library

Mensae:
## React Boilerplate

Started as POC, [see previous pull request](https://github.com/react-boilerplate/react-boilerplate/pull/2293), this PR is updated to version 3.7 of rbp.
In this PR I also adapted the testing documentation.

I chose not to add setup and teardown options to the global test config, [see](https://github.com/kentcdodds/react-testing-library#global-config). Whilst it simplifies things I like things to be explicit.

As I, at a minimum, use a snapshot test most of the time and find myself copy/pasting this test a lot, I added this test to the generator templates. In the spirit of rbp this test is wrapped in the [skip](https://jestjs.io/docs/en/api#testskipname-fn) method so one has to deliberately 'unskip' it.

Any feedback will be appreciated :-)


julienben:
Daamn @Mensae you didn't waste any time!! Awesome job! Will review this Sunday at the latest.

Issue #2314, opened 2018-08-13
SAGA: Cannot read property cancel of undefined


aloysius-tim:
# React Boilerplate

## Issue Type

- [X] Bug (https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md#bug-reports)
- [ ] Feature (https://github.com/react-boilerplate/react-boilerplate/blob/master/.github/CONTRIBUTING.md#feature-requests)

## Description

Hi there,

Just wanted to point out that the following issue has not been resolved yet: https://github.com/react-boilerplate/react-boilerplate/issues/2021

@kai23 solution of adding '&& descriptor.task' on line 75 worked for me.

```
export function ejectSagaFactory(store, isValid) {
  return function ejectSaga(key) {
    if (!isValid) checkStore(store);

    checkKey(key);

    if (Reflect.has(store.injectedSagas, key)) {
      const descriptor = store.injectedSagas[key];
      if (descriptor.mode && descriptor.mode !== DAEMON && descriptor.task) {
        descriptor.task.cancel();
        // Clean up in production; in development we need `descriptor.saga` for hot reloading
        if (process.env.NODE_ENV === 'production') {
          // Need some value to be able to detect `ONCE_TILL_UNMOUNT` sagas in `injectSaga`
          store.injectedSagas[key] = 'done'; // eslint-disable-line no-param-reassign
        }
      }
    }
  };
}
```

julienben:
Hey @aloysius-tim,

Thanks for submitting this. I've never seen it on my end nor do I know how to reproduce it.

Do you have any idea why adding this check fixes it though?

Read: Does this simply bypass the bug or is it a legitimate check?

Cheers!

NathanBWaters:
I just hit this error.  Just want to note that I'm only hitting this in production mode and not development mode.  Adding the `&& descriptor.task` fix also worked for me.

tatemz:
> Note: I don't have a suggested fix, but hopefully the below details can help the next dev better understand the issue.

I too have this issue in `production` mode. The issue occurs when I switch languages using `react-intl`. Some of my components use `InjectIntl` directly in order to translate strings (rather than HTML elements). See [API docs](https://github.com/yahoo/react-intl/wiki/API#injection-api) for more details.

The root of the problem lies within the React component lifecycle.

[...]

Issue #2121, opened 2018-01-26
[suggestion] DAEMON as saga's default mode


FourwingsY:
from #2021 

I thought the problem is on saga's default mode: RESTART_ON_REMOUNT

regarding to https://github.com/facebook/react/issues/11106
garaeon saids:

> Regarding componentWillUnmont.
> 
> The official docs explicitly recommend use of this hook as cleanup, so this will probably break a lot of people unless changed/docs update/some workaround is recommended.
> 
> The docs are talking about clean up on the component level. For example listeners added by a component.
> 
> What you're talking about is cleaning up global state. This is inherently messy and can get very complicated. You're welcome to file an issue or PR in the docs repo to clarify it.
> 
> fire a cleanup action/set a flag to reset a part of the Redux state tree before showing another component that also uses that part of the state tree but needs it to be re-initialized before render
> 
> I've written such code myself before but always regretted it. This is brittle. I recommend refactoring away from patterns where you need this.
> 
> The easiest way to do so is to not reuse the same part of the tree in such a way. Think about it: components are generally expected to be independent. What happens if you render two such "exclusive" components at the same time? They will conflict. This doesn't seem like the right component thinking.
> 
> I understand it is more code to avoid such patterns, but in my experence replacing a single state tree branch with an object or an array was enough. For example you can key it by ID (even a dynamically generated one), and then instead of tearing down the old one in time for the new component mounting, you can just point the new component to the branch with new ID, while keeping the old one alive. Then eventually you can emit an action that cleans up unused state branches and only leaves the one with the most recent ID.

So ejecting sagas with componentWillUnmount is unnecessary. Sagas are on global-level, and It should not controlled by component-level actions. Sagas should only injected just like reducers. 

I think we should remove RESTART_ON_REMOUNT options, and set DAEMON as default.
Actually mods are not necessary, if we do this.

How do you guys think about?

blling:
Saga injection issue is so sneaky, i think we need experts :)
cc @Dattaya @justingreenberg

svitekpavel:
I just run into similar issue as in #2132. I was trying to use saga from other container to update data through API. The problem was that the saga was ejected in the middle of the API call and I didn't see any information in the console.

To the topic:
I don't think it's necessary to make all sagas DAEMONS, but I would suggest two things:

1. Add a global daemon sagas into `containers/App/saga.js` and mount it in `containers/App/index.js` like this:
```
import { DAEMON } from 'utils/constants';
[...]

PR #2467, opened 2018-11-14
Fix saga multiple injection problem(#2314)

tonyex52:
## purpose

According to the issue #2314, 
using `componentDidMount` and `Promise` to prevent saga multiple injection problem.

- `componentWillMount` has the render ordering problem ([comment](https://github.com/react-boilerplate/react-boilerplate/issues/2314#issuecomment-426911261)).
- Cuz of React v16, `componentWillMount` is unsafe and we should not use it.

Using this modification, we can assure that the order sagas are injected is correct

## order before fix
1. `Page A(1)`'s `componentWillMount`
2. `switch language (to make Page A re-Mount)`
3. `Page A(2)`'s `componentWillMount`
4. `Page A(1)`'s `componentWillUnMount`

## oder after fix 
1. `Page A(1)`'s `componentDidMount`
2. `switch language (to make Page A re-Mount)`
3. `Page A(1)`'s `componentWillUnMount`
4. `Page A(2)`'s `componentDidMount`

### DIFF
--- a/app/utils/injectSaga.js
+++ b/app/utils/injectSaga.js
@@ -27,21 +27,33 @@ export default ({ key, saga, mode }) => WrappedComponent => {
       WrappedComponent.name ||
       'Component'})`;
 
-    componentWillMount() {
-      const { injectSaga } = this.injectors;
+    state = {
+      isReady: false,
+    };
 
-      injectSaga(key, { saga, mode }, this.props);
+    componentDidMount() {
+      const { injectSaga } = this.injectors;
+      this.isComponentMounted = true;
+      injectSaga(key, { saga, mode }, this.props).then(() => {
+        if (this.isComponentMounted) {
+          this.setState({ isReady: true });
+        }
+      });
     }
 
     componentWillUnmount() {
       const { ejectSaga } = this.injectors;
 
+      this.isComponentMounted = false;
       ejectSaga(key);
     }
 
     injectors = getInjectors(this.context.store);
 
+    isComponentMounted = false;
+
     render() {
+      if (!this.state.isReady) return null;
       return <WrappedComponent {...this.props} />;
     }
   }
--- a/app/utils/sagaInjectors.js
+++ b/app/utils/sagaInjectors.js
@@ -25,39 +25,46 @@ const checkDescriptor = descriptor => {
 
 export function injectSagaFactory(store, isValid) {
   return function injectSaga(key, descriptor = {}, args) {
-    if (!isValid) checkStore(store);
+    return new Promise((resolve, reject) => {
+      try {
+        if (!isValid) checkStore(store);
 
-    const newDescriptor = {
-      ...descriptor,
-      mode: descriptor.mode || DAEMON,
-    };
-    const { saga, mode } = newDescriptor;
+        const newDescriptor = {
+          ...descriptor,
+          mode: descriptor.mode || RESTART_ON_REMOUNT,
+        };
+        const { saga, mode } = newDescriptor;
 
-    checkKey(key);
-    checkDescriptor(newDescriptor);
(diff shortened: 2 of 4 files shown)


app/utils/injectSaga.js
```diff
@@ -27,21 +27,33 @@ export default ({ key, saga, mode }) => WrappedComponent => {
...
+    };
 
-      injectSaga(key, { saga, mode }, this.props);
+    componentDidMount() {
+      const { injectSaga } = this.injectors;
+      this.isComponentMounted = true;
+      injectSaga(key, { saga, mode }, this.props).then(() => {
+        if (this.isComponentMounted) {
+          this.setState({ isReady: true });
+        }
```
julienben: I don't understand. What's the purpose of `this.isComponentMounted`? In case the component is unmounted before the injectSaga promise is resolved?

In any case, lines 38-40 would look cleaner if you did:

```js
this.setState({ isReady: this.isComponentMounted });
```

app/utils/sagaInjectors.js
```diff
@@ -25,39 +25,46 @@ const checkDescriptor = descriptor => {
...
+        if (!isValid) checkStore(store);
 
-    const newDescriptor = {
-      ...descriptor,
-      mode: descriptor.mode || DAEMON,
-    };
-    const { saga, mode } = newDescriptor;
+        const newDescriptor = {
+          ...descriptor,
+          mode: descriptor.mode || RESTART_ON_REMOUNT,
```
julienben: The default mode has been switched to DAEMON in #2430. Please make sure you don't change it back to RESTART_ON_REMOUNT as it appears here.

app/utils/sagaInjectors.js
```diff
@@ -25,39 +25,46 @@ const checkDescriptor = descriptor => {
...
+          /* eslint-disable no-param-reassign */
+          store.injectedSagas[key] = {
+            ...newDescriptor,
+            task: store.runSaga(saga, args),
+          };
+          /* eslint-enable no-param-reassign */
+        }
+        resolve();
+      } catch (error) {
+        reject(error);
```
julienben: You have a `reject` here, which makes sense. But no error handling in the `injectSaga` HOC. I think that should be added.

app/utils/injectSaga.js
```diff
@@ -27,21 +27,33 @@ export default ({ key, saga, mode }) => WrappedComponent => {
...
+      this.isComponentMounted = false;
       ejectSaga(key);
     }
 
     injectors = getInjectors(this.context.store);
 
+    isComponentMounted = false;
+
     render() {
+      if (!this.state.isReady) return null;
```
julienben: I'm not sure we can do this. It could lead to flashes of white screens and maybe break animations/page transitions for those who use them, no?

tonyex52:
@julienben I tried to use `constructor` instead of `componentDidMount`, but I found that the store in context would be undefined
image
here is my code
```javascript
constructor(props) {
  super(props);
  this.injectors = getInjectors(this.context.store);
  const { injectSaga } = this.injectors;
  injectSaga(key, { saga, mode }, this.props);
}
```

app/utils/sagaInjectors.js
tonyex52: I know the default mode has been switched to DAEMON, and in #2121, to avoid the `componentWillMount` issue, it changed to DAEMON.

But the Pull Request is to fix `componentWillMount` issue

julienben: It was also changed to DAEMON because most people found it more intuitive. Please leave it that way.

If it must be changed again, it should be done in a separate PR after we've had a discussion about it and agreed that there's a more intuitive choice than DAEMON.

PR #2473, opened 2018-11-18
Rewritten the generate-templates-for-linting script

Mensae:
### A problem

In order to test the generator functionality the `test-generators-without-a-trace` is executed on Travis. This runs all generators for components/containers and language.
Then the linter is scheduled to run so it inlcudes these generated components etc.

But any, snapshot, test for components wrapped in `<LanguageProvider />` fail because of the added language by the language generator.

### A solution

Therefore I've rewritten the `test-generators-without-a-trace` script. It runs all generators as before but checks for linting issues right after and on the component, container or modified file, if the generation was successful.
When the generator actions were successful and no linting problems were found the component, container or any added files are removed and modified files are restored.
In case any error occured the component, container or any added files are NOT removed and modified files are NOT restored so issues can be traced.

### DIFF
--- a/internals/generators/index.js
+++ b/internals/generators/index.js
@@ -11,6 +11,12 @@ const componentGenerator = require('./component/index.js');
 const containerGenerator = require('./container/index.js');
 const languageGenerator = require('./language/index.js');
 
+/**
+ * Every generated backup file gets this extension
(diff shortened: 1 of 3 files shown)


Mensae:
Thanks! :-)

PR #2482, opened 2018-11-20
Update issue templates

gretzky:
We should consider adding 2 separate templates for bug and feature to further guide people when they visit with their ideas/problems. imo the more guidance the better.


gretzky:
Thanks @jwinn !

PR #2498, opened 2018-11-27
Subfolder deployment

nshimiye:
## React Boilerplate

## Description
My Devops team has ran into issues while deploying the app because they were using an NGINX server that serves other apps.
So, I searched for solutions and came up with one of **[using environment variables to dynamically change the basename during build process.](https://medium.com/friday-knowledge/webpack-compiler-placeholders-3598f5f723b9)**

My thinking is that documenting the process is enough.
However, the changes that I made, do not affect current development setup. So, I chose to include them.

### DIFF
--- a/app/utils/history.js
+++ b/app/utils/history.js
@@ -1,4 +1,3 @@
 import createHistory from 'history/createBrowserHistory';
-const basename = process.env.PUBLIC_PATH;
-const history = createHistory({ basename });
+const history = createHistory();
 export default history;
--- a/app/utils/history.js
+++ b/app/utils/history.js
@@ -1,3 +1,4 @@
 import createHistory from 'history/createBrowserHistory';
-const history = createHistory();
+const basename = process.env.PUBLIC_PATH;
+const history = createHistory({ basename });
 export default history;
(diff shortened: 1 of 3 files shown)


gretzky:
Thanks for bringing this up! It looks great.

My recommendation would be to put all of your code changes into that section of the docs rather than make the changes in the code. For us, it's a lot harder to justify making code changes for just 1 use case-- we want it to be skeletal, if that makes sense.

julienben:
Thanks for this! I think it's a great addition.

I agree with @gretzky though. Can we bring the code changes into the `deployment.md` instructions?

For a few reasons:
- This is just too rare of a use case to force a potentially brittle change onto all users. (I say brittle because all the changes make sense when you look at them from this PR but might be hard to understand without the context.)
- It only works for people deploying to NGINX or Apache, no? If yes, it needs to be specified. People wanting to do the same thing on, say, Netlify would have to do it differently.
- The code change would've also needed to be reflected in the template app's `history.js` file.

However, let's keep the switch to Webpack's EnvironmentPlugin. I wasn't aware of it but it's definitely nicer than the current version.

nshimiye:
@julienben and @gretzky make sense! I will make the changes later today

nshimiye:
@julienben and @gretzky I have made the changes! Let me know if that looks fine

nshimiye:
@julienben fixed as requested!

julienben:
Thanks for your diligence!! Will merge this as soon as I'm back in front of a computer.

PR #2540, opened 2019-01-11
update to include react-redux:6.0.0 and connected-react-router:6.1.0

bumi001:
Update package.json to include react-redux:6.0.0 and connected-react-router:6.1.0 and others

Modify createReducer to return a function that calls combineReducers

Modify configureStore to call the modified createReducer

Import history and call the modified createReducer

Set contextType using ReactReduxContext, get store from context and pass it to getInjectors

Add package-lock.json

Integrate react-redux:6.0.0 and connected-react-router:6.1.0 into react-boilerplate

## React Boilerplate


app/utils/tests/sagaInjectors.test.js
```diff
@@ -24,7 +24,7 @@ describe('injectors', () => {
 
   describe('getInjectors', () => {
     beforeEach(() => {
-      store = configureStore({}, memoryHistory);
+      store = configureStore({}, createMemoryHistory());
```
julienben: Is it possible to just run `createMemoryHistory` once before the tests and pass that each time?

julienben: (Same thing for the other tests which use this function. It seems like an unnecessary performance overhead.)

app/utils/injectSaga.js
```diff
@@ -9,35 +9,35 @@ import getInjectors from './sagaInjectors';
...
-
-      this.injectors.injectSaga(key, { saga, mode }, this.props);
+      injectSaga(key, { saga, mode }, this.props);
     }
 
     componentWillUnmount() {
-      this.injectors.ejectSaga(key);
+      const { ejectSaga } = this.injectors;
+
+      ejectSaga(key);
```
julienben: Is this change necessary? The behavior is exactly the same.

julienben: (Same comment for the `constructor` and the equivalent places in `injectReducer.js`.)

app/reducers.js
```diff
@@ -5,20 +5,18 @@
...
-  });
-
-  return rootReducer;
+  return history =>
+    combineReducers({
+      router: connectRouter(history),
+      global: globalReducer,
+      language: languageProviderReducer,
+      ...injectedReducers,
+    });
```
julienben: I didn't understand why `createReducer` needs to rerun `connectRouter` with the `history` each time. Would you mind clarifying that please?

And if it's necessary wouldn't it be easier if `history` was just a second argument of `createReducer`? So that you only call one function. At first look, this seems more convoluted than it needs to be.

app/utils/injectSaga.js
```diff
@@ -9,35 +9,35 @@ import getInjectors from './sagaInjectors';
...
  * @param {function} saga A root saga that will be injected
- * @param {string} [mode] By default (constants.DAEMON) the saga will be started
- * on component mount and never canceled or started again. Another two options:
- *   - constants.RESTART_ON_REMOUNT — the saga will be started on component mount and
- *   cancelled with `task.cancel()` on component unmount for improved performance,
- *   - constants.ONCE_TILL_UNMOUNT — behaves like 'RESTART_ON_REMOUNT' but never runs it again.
+ * @param {string} [mode] By default (constants.RESTART_ON_REMOUNT) the saga will be started on component mount and
+ * cancelled with `task.cancel()` on component un-mount for improved performance. Another two options:
+ *   - constants.DAEMON—starts the saga on component mount and never cancels it or starts again,
+ *   - constants.ONCE_TILL_UNMOUNT—behaves like 'RESTART_ON_REMOUNT' but never runs it again.
```
julienben: The default mode is DAEMON so this change shouldn't be here.

julienben:
Also, did you check the implications of updating `immutable` to v4? Any API changes that should be reflected in the app or the documentation?

julienben:
I just tested your changes locally and I'm getting a slew of JS errors in the console. Did you run this? Are you sure that it works?

app/utils/injectSaga.js
bumi001: ok. Will fix it.

app/reducers.js
bumi001: I just followed the example and recommendations at connected-react-router for this change.

app/utils/tests/sagaInjectors.test.js
bumi001: I just replaced what was already there to make it work. I will take a look at it again.

app/utils/injectSaga.js
bumi001: You are right! Not needed.

bumi001:
I ran it and it runs fine for me. Is there a way for me to see the error messages you are getting?

bumi001:
Didn't look for API changes in immutable v4. Will do.

app/utils/tests/sagaInjectors.test.js
bumi001: Didn't know your requirements about running the clean script will do. I did npm run build and it builds and runs in production mode fine.

bumi001:
> 
> 
> Didn't look for API changes in immutable v4. Will do.
Actually, react-boilerplate only uses fromJS afaik. Is there anything else you remember?

julienben:
image

Getting this on localhost.

julienben:
> Actually, react-boilerplate only uses fromJS afaik. Is there anything else you remember?

I'm not sure off the top of my head... You should do a global search inside the repo and also check that the Immutable portion of the documentation isn't out of date after your update.

Also, check your latest commit. It includes several files which shouldn't be there.

bumi001:
image

I am not seeing those errors. I am running node v10.13.0 and npm 6.5.0

bumi001:
image

This image shows that fromJS is the only immutable.js function directly used by react-boilerplate. The other places where you see immutable is used by connected-react-router and that code is already in the current master.

julienben:
Indeed this works! Apologies for doubting it!

My mistake was not switching from the old Context API's `contextTypes` key to `contextType` as in your code.

Thanks a lot for this! I expected this update to require a lot more changes.

Before we can merge though, your branch does contain a lot more changes than it needs to.

For one, none of the changes to the history/reducers setup are necessary. If you look at them closely, you'll see that they don't actually do anything besides make the code more convoluted. Also, none of the test changes seem necessary for the test suite to pass.

Take a look at [this branch](https://github.com/react-boilerplate/react-boilerplate/compare/dev...react-redux-v6). This is the extent of the changes that are necessary for this to work.

Please clean up your branch (except for the other dependency upgrades, these are fine IMO) and I'll be happy to merge it!

Cheers!

bumi001:
I am not sure if you are saying you don't need this:
https://github.com/supasate/connected-react-router/blob/master/FAQ.md#how-to-migrate-from-v4-to-v5

julienben:
We don't need it because it's already built exactly that way. Look at your code closely and you'll see that you're just passing `history` from a different location but it's still the exact same `history` object.

bumi001:
OK. I see that. You are using an extra variable and keeping it outside createRouter. Yeah, that is better. I will make another pull request shortly.

bumi001:
In utils/history.js you currently have
import createHistory from 'history/createBrowserHistory';
But, I don't see createHistory at https://github.com/ReactTraining/history/blob/v4.7.2/modules/createBrowserHistory.js.
Am I missing something?

julienben:
It's a default export in createBroswerHistory.js so you can name it whatever you want when importing it.

bumi001:
Another thing I noticed is that you have
import { memoryHistory } from 'react-router-dom';
in utils/test/injectReducer.test.js
But, I don't see memoryHistory at 
https://github.com/ReactTraining/react-router/blob/v4.3.1/packages/react-router-dom/modules/index.js
I believe this needs to be updated as well.

bumi001:
> It's a default export in createBroswerHistory.js so you can name it whatever you want when importing it.

Oops. Yeah. You are right!

julienben:
> import { memoryHistory } from 'react-router-dom';

Indeed that's `undefined`. It was in an older version of `react-router-dom` and probably didn't get updated here because the tests didn't fail when it was removed from `react-router-dom`.

bumi001:
That was my guess. That is why I had changed it in the zip file I asked you to download. In the most recent version I uploaded, I reverted it back to what it was.

bumi001:
> We don't need it because it's already built exactly that way. Look at your code closely and you'll see that you're just passing history from a different location but it's still the same history object.

That was in the master. In the dev version, you have code that is similar to what I did. However, you already had that code.

bumi001:
But, you didn't return a function that takes history as a parameter in the dev version also, which I did.

vospascal:
i tried to add the changes to my project it all works except the backbutton in the route looks to be broken

bumi001:
Not sure if you tried this version:
https://github.com/bumi001/react-boilerplate/tree/withReactReduxV600
Could you give it a try?

vospascal:
@bumi001 I will try this today

bumi001:
Not sure what you mean by action pop. Can you post some code?

vospascal:
never mind i found out it was this [blockedupdates ](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/blocked-updates.md)problem ... your code is perfect as far as i can tell

bumi001:
OK. Thank you. I also ran into that problem a couple of times.

julienben:
@bumi001 I'd like to merge this but your branch still contains many more changes than it needs to. Please look at the [changes in your branch](https://github.com/react-boilerplate/react-boilerplate/pull/2540/files) vs the [changes required for this update to work](https://github.com/react-boilerplate/react-boilerplate/compare/dev...react-redux-v6).

bumi001:
Julienben, I am tied up with work at the moment. Will have some time next weekend (3/2-3/3) and I will clean it up then.

julienben:
I merged it into a temporary react-boilerplate branch so I can clean it up there before merging to another branch.

I'd like to work on Hooks-based saga+reducer injectors and need the compatibility with React's new Context API (as used by react-redux V6) for it to work.

bumi001:
OK. Sounds good.

PR #2583, opened 2019-02-25
[RFC] Hooks-based saga and reducer injectors

julienben:
## React Boilerplate

The saga and reducer injectors are normally used via HOCs.

The HOCs will still work with Hooks-based components but IMO, we can offer a nice alternative which doesn't pollute the render tree:

`useInjectSaga` and `useInjectReducer` hooks!

You can see this alternative in this PR's diff [here](https://github.com/react-boilerplate/react-boilerplate/pull/2583/files?w=1) (with the no whitespace option enabled).

We probably shouldn't remove the HOCs as the Hooks-based injectors won't work with Class components.

I do have one question about the InjectSaga HOC: It passes props to `injectSaga`. What's the purpose of this and is there any way around it?

This is more of an RFC than anything else. Looking for thoughts and feedback before completing the branch.

Cheers!

PS: `react-helmet@5` + `useEffect` in the same component = 🧨💥🔥 so I upgraded us to the 6.0.0 beta.

image

### DIFF
--- a/app/containers/HomePage/index.js
+++ b/app/containers/HomePage/index.js
@@ -35,7 +35,7 @@ import saga from './saga';
 
 const key = 'home';
 
-function HomePage({
+export function HomePage({
   username,
   loading,
   error,
--- a/app/containers/HomePage/index.js
+++ b/app/containers/HomePage/index.js
@@ -12,8 +12,8 @@ import { connect } from 'react-redux';
 import { compose } from 'redux';
 import { createStructuredSelector } from 'reselect';
 
-import injectReducer from 'utils/injectReducer';
-import injectSaga from 'utils/injectSaga';
+import { useReduxReducer } from 'utils/injectReducer';
+import { useSaga } from 'utils/injectSaga';
 import {
   makeSelectRepos,
   makeSelectLoading,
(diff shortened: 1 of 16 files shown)


app/containers/HomePage/index.js
```diff
@@ -128,11 +132,4 @@ const withConnect = connect(
...
 
-const withReducer = injectReducer({ key: 'home', reducer });
-const withSaga = injectSaga({ key: 'home', saga });
-
-export default compose(
-  withReducer,
-  withSaga,
-  withConnect,
-)(HomePage);
+export default compose(withConnect)(HomePage);
```
jwinn: Do we want to use `React.memo()` here to account for the loss of extending from `PureComponent`? I know it's premature performance optimization, but was thinking about consistency.

julienben: I'm fairly sure it provides no benefit at all in this particular case (neither did `PureComponent`) but I'll add it anyway because IMO it's not a commonly used API and we can promote its existence by using it in the sample app.

Remove two HOCs, add another one! 😅

app/containers/HomePage/index.js
```diff
@@ -33,68 +33,72 @@ import { makeSelectUsername } from './selectors';
...
+const key = 'home';
 
-  render() {
-    const { loading, error, repos } = this.props;
-    const reposListProps = {
-      loading,
-      error,
-      repos,
-    };
+function HomePage({
```
jwinn: Let's add `export` here so the component can be tested without the HOC, if needed. Also, this _should_ get the tests to pass. 😄

julienben: Good catch. Updating now.

Note: It doesn't fix ALL the tests (notably the ones which check features that rely on useEffect) but I'm gonna wait til we merge #2592 to take care of that.

jwinn: Agreed. Nice job! 👍

julienben:
@gretzky @jwinn This is ready for final review.

PR #2586, opened 2019-02-25
Master changes back into dev

julienben:
## React Boilerplate

@gretzky, I think what was happening was some of the commits in master were also in dev but not all of them. So git/github were freaking out.

I created a new branch from the same point as master and rebased those commits on top of dev. Everything should be good now and we'll be able to merge dev into master when we're ready for a release.


julienben:
(As you can see, it only replayed the last 3 commits from master. The others are already in dev.)

PR #2594, opened 2019-03-06
Add support for eslint-plugin-compat

julienben:
## React Boilerplate

What do you think of adding [eslint-plugin-compat](https://github.com/amilajack/eslint-plugin-compat)?

Basically, it's an ESLint plugin which checks the browser compatibility of your code.

It relies on a browserslist config (which I've added to package.json) and you can also tell it which features are polyfilled so it won't trigger errors for those.

Note that it uses data directly from caniuse.com and MDN so it should be very accurate.


julienben:
Travis is failing for a super weird (and seemingly unrelated) reason :(

image

julienben:
Ok Travis is failing everywhere so it's not related to this PR. Will investigate ASAP.

gretzky:
Weird, I commented this in the other PR but maybe a dev merge might solve? With new testing lib in and all

julienben:
Yup it's fixed! Ready for review ;)

gretzky:
Cool !

PR #2598, opened 2019-03-07
Maintenance

julienben:
## React Boilerplate

Simple maintenance work for our dependencies.

Includes (among many other things) the latest version of react + a non-alpha version of jest-cli which should work with node v11.11.


julienben:
That was fast! Thanks for the review!

jwinn:
Funny enough, I was doing PR review on another project, so this just slid in naturally--great timing! 😄

PR #2622, opened 2019-04-14
Generator and documentation updates for Hooks

julienben:
## React Boilerplate

**1/** The main change in this branch is that the generators no longer offer you the choice of `Stateless Function` vs `React.PureComponent` vs `React.Component`.

Everything is now a functional component but I added the option to wrap the generated component inside `React.memo`.

I tested but please review this carefully nonetheless.

(Side question: Aren't `PureComponent` and `React.memo` redundant with `reselect`? @jwinn)

**2/** Containers generated with reducers and sagas use the `useInjectReducer` and `useInjectSaga` hooks instead of the HOCs we've had until now.

Note that the HOCs are still there so we still support classes, just like React.

**UPDATE:** I've added tests for the hooks.

**3/** Updates to the generate-templates-for-linting script and the documentation to reflect the aforementioned changes.


app/utils/tests/injectReducer.test.js
```diff
@@ -66,3 +67,32 @@ describe('injectReducer decorator', () => {
     expect(children.props).toEqual(props);
   });
 });
+
+describe('useInjectReducer hook', () => {
```
gretzky: i LOVE you for this test

internals/generators/component/index.js
```diff
@@ -37,6 +26,12 @@ module.exports = {
         return 'The name is required';
       },
     },
+    {
```
gretzky: i LOVE you for this, part II

app/utils/tests/injectReducer.test.js
julienben: Made possible by @Mensae's work on migrating to react-testing-library

gretzky: wtb @Mensae !!!

PR #2624, opened 2019-04-15
New Webpack code-splitting config

julienben:
## React Boilerplate

I tested the code-splitting instructions laid out in [this article](https://hackernoon.com/the-100-correct-way-to-split-your-chunks-with-webpack-f8a9df5b7758). Here are the results of building the sample app before and after (with gzipping turned off for clarity):

Before:
image

After:
image

As you can see, the vendor bundle is broken down into several small scripts and the entrypoint size limit warning is gone. (Perhaps because not all the vendor scripts are needed initially?) Especially for users of HTTP2, this seems like an improvement IMO.

@gretzky You wrote the current config so I'd be particularly interested in hearing your thoughts on this.

### DIFF
--- a/internals/webpack/webpack.prod.babel.js
+++ b/internals/webpack/webpack.prod.babel.js
@@ -46,28 +46,23 @@ module.exports = require('./webpack.base.babel')({
     nodeEnv: 'production',
     sideEffects: true,
     concatenateModules: true,
+    runtimeChunk: 'single',
     splitChunks: {
       chunks: 'all',
-      minSize: 30000,
-      minChunks: 1,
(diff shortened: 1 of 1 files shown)


joshmeads:
@julienben @gretzky 

I've found this is good for a few modules, but ultimately once you hit the max chunks it simply bundles everything into the main.<hash>.chunk.js file. So instead of a large vendor file, we have 10 small split chunks, and a massive main javascript file which is getting changed (and uncached) whenever code changes.

You can see in the above example, main.js went from 24.9kb to 297kb.

Is there a way to bundle all additionals into a vendor.js file, keeping main.js just application code?

PR #2625, opened 2019-04-15
Getting ready for release

julienben:
## React Boilerplate

In this PR:
- The full v4 changelog. Thoughts?
- Some last minute dependency upgrades (including one, stylelint v10, which requires us to raise the min lts/dubnium version to 8.15.1)
- A fix for an accidental difference between internal and template app tests

We're so close to releasing v4 I can smell it


jwinn:
Is the bp minimum lts/carbon (8.15.1) or lts/dubnium (currently 10.15.3)? If it's carbon, should we update the nvmrc to reflect this?

julienben:
Oops! I forgot that 8 is carbon and 10 is dubnium.

8.15.1 is listed as the minimum because that's required by our dependencies (stylelint v10 particularly). And nvmrc is set to lts/dubnium simply because that's the latest lts release.

Does it make sense to leave it that way?

julienben:
Immer v3 was literally just released... Unless someone else can push it quickly to this branch I'll add it when I'm back in front of my work computer tomorrow.

PR #2641, opened 2019-05-05
Polyfill updates and maintenance

julienben:
## React Boilerplate

In this PR:
- Following `@babel/polyfill`'s deprecation, this enables `react-app-polyfill` exclusively and removes duplicate polyfills. It contains `core-js/stable` (v3) and `regenerator-runtime` plus additional polyfills for compatibility all the way back to IE9. It can be updated very easily for IE11 support (instead of 9) or simply no IE support.
- Dependency updates!
- Removes unnecessary polyfill from Jest setup.

PS: Ignore the Codacy check. It's just a disagreement on Markdown formatting and I disabled the integration anyway so it shouldn't appear here.


julienben:
Thanks @jwinn!

PR #2648, opened 2019-05-07
React-redux hooks

julienben:
## React Boilerplate

This is a POC implementation of react-redux hooks in react-boilerplate.

Upsides:
- Less code (138 -> 105 lines!)
- Everything declared at the top, no need to scroll down and understand mapState and mapDispatch
- No more connect HOC
- No more React.memo HOC because what's the point if there aren't any props?

Downsides:
- PropTypes, what are they for? Do we need TypeScript now?? 😵
- You can't test an "unconnected" version of the container anymore

Thoughts anyone?

PS: Yes, it works. Tests need fixing.


gretzky:
I joined the TypeScript darkside about 6 months ago, it's...necessary

jwinn:
I feel like we should discuss officially adding TypeScript to the bp....

On a side note, it concerns me, a little, to use an alpha version of `react-redux`, even if I really like the way it reads and reduces the amount of HOC leveraged. Otherwise, LGTM

julienben:
Re:using an alpha version of react-redux, no worries, I have no intention of merging this before the rr hooks are officially released. Their API might not even be final.

Re:TypeScript, yeah I agree. And it sounds like @gretzky does too and already has an opinion. My initial thought was making it optional but perhaps that won't be practical... Help is welcome, I have 0 TypeScript experience personally.

mxstbr:
I ain't got no skin in the game anymore, but strongly in favour of adding TypeScript. Cannot build any big app without it anymore.

jwinn:
I'm gonna try and carve out some time over the next week or so to create a PR for TypeScript in the bp. 😄

mxstbr:
Sorry accidentally assigned myself 😆

julienben:
He who self assigns has to finish the PR. Such are the laws of the Hub of Gits.

Can-Sahin:
Hi. I think typescript makes its way too pratical. I already have a [fork of typescript](https://github.com/Can-Sahin/react-boilerplate-typescript) (which is also in this repo docs). I ve been working on it for a time now. Take a look at there. Its for the master branch. There were some problems i already tackled and might help you pass them quickly. Or if you want I can help with the typescipt. I have quite alot experience with it. 

I even removed `immer`. Because readonly interfaces are dead simple for that purpose. My friend and I also working on react-redux hooks / typescript version of this repo. 
@julienben 

EDIT:
If you want I can make an example TS version of this read-redux hooks and you can look at for examining and decide upon.

julienben:
@Can-Sahin Thanks! I didn't know your fork was already caught up with 4.0! That's awesome.

At this point, it seems like the move to typescript is happening. I don't see anyone arguing against it at least.

Any help you can provide would be much appreciated!

julienben:
(Removing `immer` is IMO a different story and would need to be discussed after a switch to TS. Personally, I don't see the need, `immer` makes it so much easier to write reducers.)

Can-Sahin:
Ok. I can help for sure. Already did the majority of the work. What do you want me to do at this point? I could make a PR into dev but no point in that because the code is very different and not comparable. I guess, there can be two repo. One js and one ts. People are free to choose either as a start point. Typescript is still mystery for many JS people and some typescript things are still bugging me and putting a burden sometimes. TS is after all still in rise and community-wise not fully settled, personally thinking. 

What do you want me to do? I can simply turn this `redux-hook` branch into TS and discuss it later. (Since i didnt write any tests before for my react project, i didnt adjust testing so far. But can do later)

Edit: typos

julienben:
Hmmm good question about whether we want both a JS and a TS version or just a TS version. I have no idea. Seems like a lot of work maintaining both in parallel. And beyond that, do we even want to offer JS/TS as an option or is it smarter just having one be the official recommended way?

Either way, the work probably shouldn't be done in this branch. I was thinking of doing a v4.1 release with what we have in dev now + maintenance + react-redux hooks. TS is a huge change and we'd probably want to make it into a major release.

Can-Sahin:
I understand. I didn't mean to make changes into this repo or branch actually. Just meant I can make a ts version of this current version in another repo. I already have updated `4.0` to typescript but it is as you say some work for me every time and I didn't even bother with test configuration yet. 

Well, if there is a new version `4.1` I will catch my fork up with it. You can let me know if typescript is happening or when its happening and I can help for sure. I want to help. Appreciate the boilerplate. Happy to be part of.

julienben:
Awesome!! So I guess we'll make a decision as to whether we migrate to TS completely or maintain two versions in parallel and then we'll take it from there!

(If we have two versions, your fork is essentially the foundation for the TS version but we'd need to make sure they're 100% the same repo, except for the language of course.)

julienben:
FYI, I have a strong preference towards keeping only one version because otherwise, the two dev branches will be very hard to keep in sync and we'll need to ask any contributor to do twice the work. I think if we go TS we need to commit to TS.

Can-Sahin:
Yep two repo = duplicate effort. I was only synching master branches and since releases are not so often it was a ok job to do. But merging is hell sometimes due to file structures, extras, and file extension differences (tsx vs js). I can't believe merging every dev commit. 

So single repo makes sense. But then typescript must be embraced fully :) When its time I can work on making TS 100% same with JS. Not much left to go for me. I already did the essential researches, methodology and techniques.

csantiago132:
@julienben im well versed with the repo (which I love btw) and decent with typescript and would love to help out refactoring it to TS if the repo ends up going that way

julienben:
Thanks @csantiago132!! Let's get this effort started soon and with all the expertise here we'll be sure to release an awesome TypeScript v5!

julienben:
I'll soon open a TypeScript branch which will follow dev and where we can progressively merge contributions to the migration.

csantiago132:
> Thanks @csantiago132!! Let's get this effort started soon and with all the expertise here we'll be sure to release an awesome TypeScript v5!

@julienben yes! im finally settling in in life; life has slowed down, I move-in to new apt,  & been at current job for quite some time now so im ready, I learned a lot of react-redux ecosystem and best practices with this repo (way back to 3.3!! ) so Im dying to give back to it

jamesdh:
Just an FYI since I was following this PR, React Redux 7.1.0 w/ Hooks is released! 

https://github.com/reduxjs/react-redux/releases/tag/v7.1.0

Can-Sahin:
I also started to implement tests in typescript to make two of the repos fully synched. Whenever the transition is, it will make it easier to have synched repos

trevorharwell:
@julienben I'm new to the discussion but have been using RB for well over a year now.  Is there any concern in losing the memoized performance benefit of the selector?  Would something like `useSelector(useMemo(makeSelectRepos, []))` be useful?

julienben:
@trevorharwell I'm not sure that we can use useMemo this way? We can probably create memoized selectors with reselect instead. See here: https://react-redux.js.org/api/hooks#using-memoizing-selectors

EDIT: Actually, we already are using reselect in the current version so it's all memoized by default.

julienben:
FWIW, there's a big error in the way it's done right now.

```js
export default function HomePage() {
  const repos = useSelector(makeSelectRepos());
  const username = useSelector(makeSelectUsername());
  const loading = useSelector(makeSelectLoading());
  const error = useSelector(makeSelectError());
```

This code recreates selectors on every render so it's not memoized at all. Gotta declare them outside the component. Will be fixed soon.

trevorharwell:
@julienben The recreation of the memoized selectors on each render was my concern.  I ended up doing exactly what you did in my own code.  While I think `useSelector(useMemo(makeMemoizedSelector))` will work, I think its a little too fancy for its own good 😂

julienben:
I've fixed the HomePage tests. What do y'all think?

Because we're testing the full connected component now, what we're doing feels more like integration tests than unit tests. I don't think that's necessarily a bad thing.

For one, it means that we're receiving the `initialState` directly from `reducer.js` and it helped me notice a discrepancy between what we were testing and what the code actually expected.

I also really like that we're testing user behavior more than individual functions. i.e. we simulate the user typing and submitting a form rather than just testing the detailed behavior of the event handlers.

(Read the comments [here](https://dev.to/gsto/new-redux-hooks-a-before-and-after-comparison-are-they-better-loj) for more thoughts on this topic.)

The one thing that bothers me is that I can't find a clean way to change the `initialState` for a container. The easy way would be to mock `initialState` directly but I couldn't figure out how to mock an object (vs a function). The workaround was to turn it into a function (`getInitialState`) but that felt wrong.

Anyone have suggestions?

Outside of this, I'll check through the docs and generators for any updates but this should be it otherwise.

julienben:
Not ready yet. Still gotta fix the generators 😅

julienben:
Ready to merge IMO! I'll keep it open a little longer for anyone who wants to comment or rereview.

Absvep:
Hi julien, when do u plan to merge? Best regards

julienben:
@Absvep Any minute! Just did some final cleanup and now waiting for checks to pass.

PR #2664, opened 2019-05-24
Remove unused options.babelQuery setting

spawnia:
I think this setting is just a relict from the past, it is no longer passed from either the dev or prod config.

The behaviour does not change, it defaults to using `babel.config.js` anyways.


julienben (approved):

julienben:
Good catch! That's definitely a relic from the past.

PR #2690, opened 2019-07-09
Remove redundant jest-dom/extend imports

julienben:
## React Boilerplate

These 3 lines were redundant as we already add jest-dom/extend-expect inside the Jest config.


jwinn (approved):
LGTM

julienben:
Thanks again @jwinn!

vineyardbovines (approved):

PR #2693, opened 2019-07-10
Cleaner way to use immer's produce function

julienben:
## React Boilerplate

Simple switch to the cleaner/more elegant way to use immer's produce function.


jwinn (approved):
LGTM. Nice job 👍

PR #2702, opened 2019-07-19
Use semantic structure and replace div with header element

realdeepak:
For semantic structure, `<header>` tag is missing in the code and `<footer>` tag is already in the code. And we can see that the code is broken it into header component but not using `<header>` tag.

So, replacing `<div>` with `<header>` for better semantic structure.


julienben:
Why did you close this before we had a chance to comment/discuss? :(

PR #2712, opened 2019-08-11
Converted the request to an async function.

egdavid:


julienben:
Hi @egdavid, I'n happy to merge this but can you please reopen this PR with dev as your base branch please? We don't merge directly into master unless we're releasing a new version.

PR #2713, opened 2019-08-15
Converted the request to an async function

egdavid:
There we go!


julienben (commented):
Thanks for your contribution! Looks good!

julienben (approved):
Thanks for your contribution! Looks good!

PR #2686, opened 2019-06-28
React Boilerplate v4.1

julienben:
# v4.1: "The One with Hooks, Redux (Redux Hooks)" Edition

It's been a little over 2 months since we released v4.0. We haven't had any huge changes since but IMO it would be nice to do a maintenance release.

Anybody have any thoughts or additions? Anything we should be including urgently?

I haven't written a full changelog yet but here are the highlights:
- Switch to react-redux's hooks API
- Very nice testing improvements (Thanks to @tigranpetrossian)
	- Easier test navigation with jest-watch-typeahead
  	- More declarative component tests using jest-dom/extend-expect
- Tests for loadable.js
- Many dependency updates
- New polyfill setup (after deprecation of @babel/polyfill)
- BONUS: Thanks to a bugfix in babel-plugin-styled-components, we now support the css prop

PR #2721, opened 2019-08-19
Upgrade react-intl to v3

mcmxcdev:
Based on the comment by @julienben in https://github.com/react-boilerplate/react-boilerplate/pull/2686/#issuecomment-521299449, I decided to upgrade react-intl to its newest version.

Unfortunately I was not able to test if the [polyfills](https://github.com/formatjs/react-intl/blob/master/docs/Upgrade-Guide.md#migrate-to-using-native-intl-apis) are working since I dont have access to neither Safari, Edge or IE on Ubuntu, maybe someone can help out ;)


mcmxcdev:
@julienben saw your feedback in https://github.com/react-boilerplate/react-boilerplate/pull/2686#issuecomment-531851190 :
How exactly is this PR broken/incomplete? If its just about the conflicting files thats done fast.
 I was waiting for feedback which I didn't receive yet. Would be happy to finish this ;)

julienben:
Hey @mhatvan,

Sorry about that! I just assumed something was broken given that the CI didn't pass. From a quick look at the code, everything seems good and the CI issues seem easy to fix.

I'm assuming the two new dependencies are required by the new version of `react-intl`? I'd like to check the evolution of the build size after this change.

A couple more things:
- Did you check if the documentation needs to be updated? Does it make any reference to now-outdated code? Does it explain how to add a new language and now those instructions should be updated?
- Same for the generators which allow adding a new language. Did you try out `npm run generate language`?

Btw, I have access to a PC with IE and other browsers and will test it out ASAP.

mcmxcdev:
@julienben In my newest commit, I resolved merge conflicts and updated the new dependency imports since the documentation of the actual [upgrade guide](https://github.com/formatjs/react-intl/blob/master/docs/Upgrade-Guide.md) itself changed.
> I'm assuming the two new dependencies are required by the new version of `react-intl`? I'd like to check the evolution of the build size after this change.
>
The new dependencies are necessary due to breaking API changes -> `addLocaleDate` has been 
removed from the `react-intl` dependency. You can read about it here: https://github.com/formatjs/react-intl/blob/master/docs/Upgrade-Guide.md#breaking-api-changes

> Did you check if the documentation needs to be updated? Does it make any reference to now-outdated code? Does it explain how to add a new language and now those instructions should be updated?
>
Good point, I updated the `i18n.md` accordingly.

> Same for the generators which allow adding a new language. Did you try out npm run generate language?
>
Still works as expected, verified it with added fr locale.

> Btw, I have access to a PC with IE and other browsers and will test it out ASAP.
>
That would be great, so that we can get this merged.

vineyardbovines (approved):
LGTM on my end! Thanks @mhatvan

app/i18n.js
```diff
@@ -7,16 +8,19 @@
...
-const deLocaleData = require('react-intl/locale-data/de');
+if (!Intl.PluralRules) {
+  require('@formatjs/intl-pluralrules/polyfill');
+  require('@formatjs/intl-pluralrules/dist/locale-data/de'); // Add locale data for de
+}
+
+if (!Intl.RelativeTimeFormat) {
+  require('@formatjs/intl-relativetimeformat/polyfill');
+  require('@formatjs/intl-relativetimeformat/dist/locale-data/de'); // Add locale data for de
+}
```
julienben, line 23: The sample app has both `de` and `en` so I believe you're missing a couple lines above:

```js
if (!Intl.PluralRules) {
  require('@formatjs/intl-pluralrules/polyfill');
  require('@formatjs/intl-pluralrules/dist/locale-data/de'); // Add locale data for de
  require('@formatjs/intl-pluralrules/dist/locale-data/en'); // Add locale data for en
}

if (!Intl.RelativeTimeFormat) {
  require('@formatjs/intl-relativetimeformat/polyfill');
  require('@formatjs/intl-relativetimeformat/dist/locale-data/de'); // Add locale data for de
  require('@formatjs/intl-relativetimeformat/dist/locale-data/en'); // Add locale data for en
}
```

No?

internals/webpack/webpack.base.babel.js
```diff
@@ -21,7 +21,8 @@ module.exports = options => ({
     rules: [
       {
         test: /\.jsx?$/, // Transform all .js and .jsx files required somewhere with Babel
-        exclude: /node_modules/,
+        // excluding full node_modules folder except for react-intl related deps, see: https://github.com/formatjs/react-intl/blob/master/docs/Upgrade-Guide.md#webpack
+        exclude: /node_modules\/(?!(react-intl|intl-messageformat|intl-messageformat-parser)\/).*/,
```
julienben: IMO, your comment should link to the setup part of the docs instead of the upgrade guide:

https://github.com/formatjs/react-intl/blob/master/docs/Getting-Started.md#esm-build

julienben (changes requested):
Hey @mhatvan! Thanks for your hard work on this! Looks great and almost ready to merge. Please see a couple of final comments.

Also, are you sure about the language generator still working? Shouldn't it also add language-specific locale data for the PluralRules and RelativeTimeFormat polyfills?

```js
// In app/i18n.js, when adding for example 'fr':

require('@formatjs/intl-pluralrules/dist/locale-data/fr');

require('@formatjs/intl-relativetimeformat/dist/locale-data/fr');
```

Right now it's adding it using the old format:

```js
const frLocaleData = require('react-intl/locale-data/fr');

addLocaleData(frLocaleData);
```

See `internals/generators/language/index.js` and related HBS templates.

internals/webpack/webpack.base.babel.js
julienben: Also, did you check if adding those artifacts to the exclude instead of including them specifically mean they do get transpiled correctly? (If you're not sure, I'll find out when checking this upgrade on an IE setup this week.)

julienben:
A couple more things I forgot about:

- Coverage went down from 100% to 99.3%, probably because of polyfills being conditionally required. Can you just add `istanbul ignore` comments to fix that?
- In `app/app.js`, we import `locale-data` as well (+ the generator adds this line for new languages as well). Do you know if this is now redundant because the new polyfill setup takes care of that or not?

(Sorry about this upgrade being so tedious. i18n is always annoying to deal with.)

app/i18n.js
mcmxcdev: You can read [here](https://www.npmjs.com/package/@formatjs/intl-relativetimeformat#usage) that the en locale is the default locale, the example shows only the de locale being imported explicitly.

internals/webpack/webpack.base.babel.js
mcmxcdev: You are right, changed the link ;)

I checked the output of `npm run build` and the libraries appeared as chunks, transpiled to ES5, but please doublecheck this on IE.

mcmxcdev:
> * In `app/app.js`, we import `locale-data` as well (+ the generator adds this line for new languages as well). Do you know if this is now redundant because the new polyfill setup takes care of that or not?
> 
Hard to get a good overview of the intl stuff, but:
After checking the `react-intl` repo, I saw that they don't have `intl` in their dependencies, and the `intl` bundle languages use `IntlPolyfill.__addLocaleData` internally which is deprecated according to the documentation. 
So I would assume that we could remove the code in `app/app.js`.

Apart from that, all browsers seem to be green for at least basic functionality of `intl`, see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#Browser_compatibility

mcmxcdev:
@julienben I pushed another commit which updates the language generation, more specifically, it adds the required language polyfills for new languages e.g.: fr

app/app.js
```diff
@@ -69,25 +69,6 @@ if (module.hot) {
...
-        import('intl/locale-data/jsonp/en.js'),
-        import('intl/locale-data/jsonp/de.js'),
-      ]),
-    ) // eslint-disable-line prettier/prettier
-    .then(() => render(translationMessages))
-    .catch(err => {
-      throw err;
-    });
-} else {
-  render(translationMessages);
```
merqurio, line 91: Hi @mhatvan , thanks for your work.

I tested this PR; removing ther `render()` in line 88 makes the application not load for the first time, although if you make any change in the source code the HMR kicks in and the app renders correctly.  This affects the production build, that doesn't render the application at all. 

_Please, note that I'm working on a fork, but this should have the same results in the template_

Hope this helps with the PR.

mcmxcdev: Hey, thank you for your review!

You are totally right and I am glad you spotted this, I accidentally removed the `render()` when removing the check for `window.Intl`, I re-added it again in the newest commit.

internals/webpack/webpack.base.babel.js
ewief: tested using windows env, as windows have different file paths maybe exclude should be 
`/node_modules[\\,/](?!(react-intl|intl-messageformat|intl-messageformat-parser)[\\,/]).*/`

app/app.js
merqurio: Happy to help,

Are you sure we can remove `window.Intl` safely ? Checking [Can I use](https://caniuse.com/#feat=intl-pluralrules) and [Mozilla](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules) seems like not all the major browsers support `window.Intl.PluralRules`. Not sure if this applies. 

Bests !

mcmxcdev: The new `@formatjs/intl-pluralrules` dependency should take care of that.

Check up in the react-intl upgrade guide: https://github.com/formatjs/react-intl/blob/master/docs/Upgrade-Guide.md#migrate-to-using-native-intl-apis

And you can read the discussion here: https://github.com/react-boilerplate/react-boilerplate/pull/2721#issuecomment-534292622

internals/webpack/webpack.base.babel.js
mcmxcdev: Thank you, I changed it in the newest commit.

nxame:
Can we get this through? Please.

merqurio (approved):

app/i18n.js
julienben: I don't see any mention of `en` being the default locale at this link. IMO, if `@formatjs/intl-relativetimeformat/dist/locale-data/en` didn't need to be included, it wouldn't exist in the `locale-data` folder but it does.

julienben (changes requested):
Sorry it took me so long to review this! I only added one comment regarding `en` as the default locale.

More importantly though, I cloned your fork and tested this branch on a Windows 10 laptop with IE11 and the app was broken. This is what I got in the console:

8UHVO3c

I'm not sure what the issue is but if you have an idea and would like me to try a fix on PC, let me know. I'll make sure to take care of it faster this time!

app/i18n.js
mcmxcdev: I guess I was not imagining it, they removed it from the README.md a while ago: https://github.com/formatjs/formatjs/commit/cda3cf2090abf6a46e293ad049eb7f588f9a66eb#diff-70b3d6472c5178da3042f45cb893ef56

I am updating react-intl and relevant dependencies again, since there were a lot of new version releases in the mean time, I will keep you updated.

julienben: Thanks @mhatvan! Sorry again for the delay on my part!

mcmxcdev:
> More importantly though, I cloned your fork and tested this branch on a Windows 10 laptop with IE11 and the app was broken. This is what I got in the console:
> 
> 8UHVO3c
> 
> I'm not sure what the issue is but if you have an idea and would like me to try a fix on PC, let me know. I'll make sure to take care of it faster this time!

@julienben I found out and was able to reproduce this by adding `delete Intl.PluralRules` in `i18n.js` to simulate a kind of faked IE11 mode.
After updating react-intl and related dependencies to the newest versions again, the error you mentioned went away.

I also added the [full-icu](https://www.npmjs.com/package/full-icu) dependency to fix missing locale data errors in jest, see: https://github.com/formatjs/react-intl/issues/1401

Please run the latest commit again locally against your W10 IE11 environment and tell me if there are any remaining issues.

mcmxcdev:
@julienben I am aware of the conflicting files, but didn't do any further changes since I want you to decide when you have time and are ready to review this MR.

package.json
```diff
@@ -59,12 +59,15 @@
...
   "dependencies": {
+    "@formatjs/intl-pluralrules": "1.3.4",
+    "@formatjs/intl-relativetimeformat": "4.4.3",
     "chalk": "2.4.2",
     "compression": "1.7.4",
     "connected-react-router": "6.5.2",
     "cross-env": "6.0.3",
     "express": "4.17.1",
     "fontfaceobserver": "2.1.0",
+    "full-icu": "1.3.0",
```
ividic, line 70: Should this be a devDependency?

mcmxcdev: The install instructions on https://www.npmjs.com/package/full-icu are displaying the commands for putting it in dependencies (no -D flag), but I am open to move it to the devDependencies, if desired.

I hope we can expect some activity from @julienben soon?

cuongdevjs: Hope to check soon. I have a problem with upgrade intl to v3

cuongdevjs (approved):

Can-Sahin:
Edit: Sorry I am opening this back as a result of not merging the #2935 if that means anything;)

PR #2732, opened 2019-09-05
Maintenance

julienben:
## React Boilerplate

@BenLorantfy I saw you disabled some ESLint rules in the new Airbnb config to keep in line with the react-boilerplate code. Here I went with disabling only the props-no-spreading rule but kept the one about static properties of a class and updated the injectors accordingly. You don't have to follow the same rule. This is just what I went for here.


jwinn (approved):
LGTM. Nice work 👍

justingreenberg (approved):

Issue #2698, opened 2019-07-16
Move reducer/saga injection to separate library


BenLorantfy:
**Is your feature request related to a problem? Please describe.**
I use this boilerplate for multiple projects, which means I have some duplicate code in my utils folders for the reducer/saga injection.

**Describe the solution you'd like**
It would be nice to move the reducer/saga injection to a separate library, that this boilerplate then imports as a dependency. This would reduce code duplication when using this boilerplate for multiple projects. It would also let developers use the reducer/saga injection without using the entire boilerplate.

**Describe alternatives you've considered**
N/A

**Additional context**
I'd like to work on this if I can get the blessing of the maintainers. I'd probably also need help from the maintainers in creating a separate repo within the react-boilerplate organization. Alternatively, we could set up a monorepo, but that would be more work.

BenLorantfy:
@julienben Any thoughts on this?

julienben:
Sounds like a good idea in theory.

A few thoughts:
1. I'm assuming we're talking about the code in these files: [injectReducer](https://github.com/react-boilerplate/react-boilerplate/blob/master/app/utils/injectReducer.js), [injectSaga](https://github.com/react-boilerplate/react-boilerplate/blob/master/app/utils/injectSaga.js), [reducerInjectors](https://github.com/react-boilerplate/react-boilerplate/blob/master/app/utils/reducerInjectors.js), [sagaInjectors](https://github.com/react-boilerplate/react-boilerplate/blob/master/app/utils/sagaInjectors.js) + the corresponding tests. Is that correct?
2. This code is very specific to react-boilerplate. It expects a Redux store as configured inside [configureStore](https://github.com/react-boilerplate/react-boilerplate/blob/master/app/configureStore.js) and as verified by [checkStore](https://github.com/react-boilerplate/react-boilerplate/blob/master/app/utils/checkStore.js). How do we deal with the tight coupling between these? (At the very very least, I think the name of the project should be specific, i.e. react-boilerplate-injectors, and the documentation should be detailed.)
3. Whenever the idea to turn react-boilerplate into a monorepo has arisen, we've tried to stay away from it because many devs come to this repo to learn best practices and a monorepo would feel over-engineered and harder to grok. It could make sense as an additional package in the react-boilerplate organization though.
[...]

PR #2749, opened 2019-09-19
Move injectors code to external library

BenLorantfy:
This PR moves the injector-related code to a new external library called `redux-injectors`. For the full discussion on how we got here, see [this issue](https://github.com/react-boilerplate/react-boilerplate/issues/2698). A few guiding notes:
- See this file for the changes made to the store setup: https://github.com/react-boilerplate/react-boilerplate/pull/2749/files#diff-ea94aca90fbef7f544d2a96cf1da5f8e
   - Removes the store mutation in favour of using a [store enhancer](https://redux.js.org/recipes/configuring-your-store#extending-redux-functionality)
   - Store is now given `createReducer`. It was only given `runSaga` previously. 
- This is an example of how the imports would change: https://github.com/react-boilerplate/react-boilerplate/pull/2749/files#diff-b40dc3de2b66cf46597584200a5599f7R14
- Removed a bunch of code:
  - Removed the injectReducer, injectSaga, checkStore, and sagaInjectors files
  - Removed the related tests
  - Removed the same files in the `templates` folder
- Added `docs/js/redux-injectors.md`. I linked to the `redux-injectors` docs for readers to view the full docs, but let me know if I should go into more detail in this file.
- Changed a few documentation files to use `redux-injectors` (e.g. `docs/js/redux-saga.md`, `docs/js/redux.md`)
- Removed some steps from `docs/js/remove.md` 
- Updated the generators to use `redux-injectors`
- Updated all the templates with the same changes that were done to the source code

A few other links:
- redux-injectors on npm: https://www.npmjs.com/package/redux-injectors
- redux-injectors on GitHub: https://github.com/react-boilerplate/redux-injectors


julienben:
😍😍😍
image

Quick question: Why do we need to add aliases for react deps in the webpack config?

BenLorantfy:
That was just for local development and is needed when using `yarn link` to link `redux-injectors` to `react-boilerplate`. Planning on removing it once `redux-injectors` is published and before merging this.

BenLorantfy:
OK, this is now ready for review whenever anyone has a chance 🙂

BenLorantfy:
@julienben When you have some time, I also need help transferring redux-injectors to the reactboilerplate org as I explained in this comment: https://github.com/react-boilerplate/react-boilerplate/issues/2698#issuecomment-536151715

julienben:
I think the transfer is done:
image
Will review the code asap. Please notice that CI isn't passing though. I think it's related to coverage having dropped.

BenLorantfy:
Oh, I think it dropped because I removed a lot of code that was well tested, I'll see if I can add some more tests elsewhere.

julienben:
I haven't taken a deep look at this PR's changes yet but maybe also check if some of the tests you removed were testing code **outside** the injectors?

BenLorantfy:
I checked and I don't think I removed any tests that were testing other code. I did remove some `configureStore` tests but the coverage for that file is still 100%.

I fixed the coverage by adding some tests for LanguageProvider and Loadable. Also copied the folder `app/containers/LanguageProvider/tests` to `internals/templates/containers/LanguageProvider/tests` because it was missing previously.

julienben:
Thanks!!

Just one thing: Can you undo the copying of `app/containers/LanguageProvider/tests`? This is already handled by the cleaning script (lines 17 to 20).

BenLorantfy:
Ah I see, done

app/configureStore.js
```diff
@@ -28,30 +29,32 @@ export default function configureStore(initialState = {}, history) {
   }
 
   const sagaMiddleware = createSagaMiddleware(reduxSagaMonitorOptions);
+  const runSaga = sagaMiddleware.run;
```
julienben: Slightly prettier maybe:
```js
const { run: runSaga } = sagaMiddleware;
```

app/containers/LanguageProvider/tests/selectors.test.js
```diff
@@ -1,11 +1,17 @@
...
+    const languageState = {};
     const mockedState = {
-      language: globalState,
+      language: languageState,
     };
-    expect(selectLanguage(mockedState)).toEqual(globalState);
+    expect(selectLanguage(mockedState)).toEqual(languageState);
+  });
+
+  it('should select the initial state when state is missing', () => {
```
julienben, line 13: 👌

app/tests/store.test.js
```diff
@@ -6,28 +6,9 @@ import { browserHistory } from 'react-router-dom';
...
-      expect(typeof store.injectedSagas).toBe('object');
-    });
-  });
-
-  describe('runSaga', () => {
-    it('should contain a hook for `sagaMiddleware.run`', () => {
-      expect(typeof store.runSaga).toBe('function');
-    });
+  it('should return a redux store', () => {
+    const store = configureStore({}, browserHistory);
```
julienben: Here browserHistory is undefined because it's an old obsolete API from react-router. You'll see it's been removed in #2791. You'll get a merge conflict here. You can just remove that argument and the test should pass just fine without it.

app/tests/store.test.js
```diff
@@ -6,28 +6,9 @@ import { browserHistory } from 'react-router-dom';
...
+    expect(typeof store).toBe('object');
```
julienben: It's a bit light to say it should return a redux store but then just check that it's of type object. You should use some of the same checks that used to be in checkStore. And IMO there's nothing wrong with also checking for the presence of the keys which are added by redux-injectors (runSaga, injectedSagas, injectedReducers, and replaceReducer too or is that gone now?).

app/utils/tests/loadable.test.js
```diff
@@ -18,6 +22,13 @@ describe('loadable', () => {
     expect(firstChild).toMatchSnapshot();
   });
 
+  it('should render null by default with empty options', () => {
```
julienben, line 25: 👌

docs/js/remove.md
```diff
@@ -10,31 +10,8 @@ If you really want to get rid of it, you will have to remove its presence from s
...
-**app/utils/tests**
-
-1.  Remove two files: `injectSaga.test.js` and `sagaInjectors.test.js`
-
-**app/utils/tests/checkStore.test.js**
-
-1.  Remove `expect(() => checkStore({ ...store, injectedSagas: null })).toThrow();`
-2.  Remove `expect(() => checkStore({ ...store, runSaga: null })).toThrow();`
+4.  Remove statement `const runSaga = sagaMiddleware.run;`
+5.  Remove `runSaga` from `createInjectorsEnhancer` params
```
julienben, line 14: Does this actually work? You can use redux-injectors without redux-saga?

docs/js/redux-injectors.md
```diff
@@ -0,0 +1,27 @@
+# `redux-injectors`
+
+`redux-injectors` is a library we made ourselves by moving code we used to have
+in the utils folder into a seperate library. It allows you to dynamically load
```
julienben: I would replace the first sentence with something like this:

> [`redux-injectors`](https://github.com/react-boilerplate/redux-injectors) is an official `react-boilerplate` companion library. We built it so that it can be used and maintained independently from `react-boilerplate`.
> 
> It allows you to dynamically load...

It feels slightly more professional IMO.

internals/generators/container/index.js.hbs
```diff
@@ -21,11 +21,18 @@ import { useSelector, useDispatch } from 'react-redux';
...
+  {{else}}
+import { useInjectReducer } from 'redux-injectors';
+  {{/if}}
+{{else}}
+  {{#if wantSaga}}
+import { useInjectSaga } from 'redux-injectors';
+  {{/if}}
 {{/if}}
 {{#if wantActionsAndReducer}}
-import { useInjectReducer } from 'utils/injectReducer';
```
julienben, line 28: Do you mind opening a separate PR if you'd like to update the logic for the container generator?

(I agree that there's an issue if `wantActionsAndReducer` is false but `wantSaga` is true but it should be handled separately from this.)

julienben: By this I mean please only update the imports here.

docs/js/redux-injectors.md
```diff
@@ -0,0 +1,27 @@
+# `redux-injectors`
```
julienben, line 1: Are you linking to this file from anywhere else in the docs?

julienben (changes requested):
Hey! Looks awesome overall. Just left a few small comments.

Also:
- Make sure you merge the latest changes from dev and fix the conflicts. (Sorry, I merged a few PRs over the weekend 😅)
- Did you check if code-splitting/loadable still works the same? I'm just worried that we lost that feature partially since the containers import from `redux-injectors` and the store setup does too. To test it, compare the webpack output of a build from the dev branch and your branch. They won't be exactly the same but you should see one file for each container. (Use webpack analyze if necessary.)
- Any comments on files in the app folder also apply to the internals of course.

One more thing you might want to do is run a regression test as described in `deployment.md`. Just make sure you uncomment the `git commit` line in `clean.js`.

docs/js/redux-injectors.md
julienben: You might want to or very few people might ever read it.

docs/js/remove.md
BenLorantfy: kind of. It throws some invariant warnings. I could probably disable them in `redux-injectors`

internals/generators/container/index.js.hbs
BenLorantfy: I could just change the import but then you could up with this:
```
import { useInjectReducer } from 'redux-injectors';
import { useInjectSaga } from 'redux-injectors';
```

Which I'm pretty sure will throw a linting error under https://eslint.org/docs/rules/no-duplicate-imports , and it seems bad to have the generator generate a file with a linting error.

docs/js/redux-injectors.md
BenLorantfy: Linked to it from `js/README.md`

app/tests/store.test.js
BenLorantfy: So I added in a more robust test for checking if it returned a redux store. I'm hesitant to check for `injectedSagas`, `injectedReducers`, etc, because I'd consider that an internal implementation detail of `redux-injectors`.  I'd like to be able to change that in the future if we need to without breaking `react-boilerplate` tests.

BenLorantfy:
Looks like code-splitting is still working, I can see `FeaturePage` and `HomePage` are in two different files in the webpack output:

Screen Shot 2019-10-27 at 1 13 00 PM

### Regression test:

**Example App:**
- [x] Browse example app on development server
  - [x] Browse Features page, change language to `de`
  - [x] Browse NotFound page
- [x] Browse example app on dev tunnel
- [x] Browse example app on Production server
- [x] Browse example app offline

**Internal Commands:**

- [x] `npm run clean` (Be careful, this will commit all your changes.)
- [x] `npm run generate component` TestComp /w defaults
- [x] `npm run generate container` TestPage /w defaults

docs/js/remove.md
julienben: Could it be done by checking which keys are passed to createInjectorsEnhancer? It can be like a configuration object for the injectors and will be especially useful later if/when you add more integrations to them.

julienben: (No emergency here. We can merge this branch without this being present in injectors.)

docs/js/redux-saga.md
```diff
@@ -49,19 +49,18 @@ export default function* rootSaga() {
 Then, in your `index.js`, use our custom hook to inject the root saga:
```
julienben: This text here says "our custom hook". Could be updated as well no?

docs/js/README.md
```diff
@@ -23,6 +23,7 @@ add new parts of your application!
 - [Immer](immer.md)
 - [reselect](reselect.md)
 - [redux-saga](redux-saga.md)
+- [redux-injectors](redux-injectors.md)
```
julienben, line 26: 👌👌

docs/js/redux-saga.md
julienben: (with an additional link to your redux-injectors.md file for example)

internals/generators/container/index.js.hbs
julienben: Oh my bad! I went over this too fast. The logic remains the same but it looks more complicated because you're importing from one package instead of 2 files.

internals/scripts/helpers/get-npm-config.js
```diff
@@ -0,0 +1,3 @@
+const fs = require('fs');
+
+module.exports = JSON.parse(fs.readFileSync('package.json', 'utf8'));
```
julienben: Merge conflict issue. This file shouldn't be here.

internals/templates/tests/i18n.test.js
```diff
@@ -0,0 +1,28 @@
+import { formatTranslationMessages } from '../i18n';
```
julienben: This file shouldn't be here no?

internals/scripts/npmcheckversion.js
```diff
@@ -0,0 +1,8 @@
+const { exec } = require('child_process');
```
julienben: This file shouldn't be here either.

julienben (changes requested):

docs/js/remove.md
BenLorantfy: Possibly, I'd have to look into it more

internals/scripts/helpers/get-npm-config.js
BenLorantfy: ugh not sure how that happened

docs/js/remove.md
julienben: No rush. Just consider it for a future release of injectors.

julienben (approved):
@BenLorantfy it's been a pleasure working with you on this. Thanks for your patience and your attention to detail! I hope you'll continue contributing!

BenLorantfy:
No problem, thanks for your guidance and help with this 👍

PR #2753, opened 2019-09-22
Fix extraneous dependencies linting setup + some maintenance

julienben:
In this PR:
- I enabled the `import/no-extraneous-dependencies` ESLint rule for everything but the internals and server folders. It raised a couple of issues which are now fixed. (I did this because this rule being disabled led to a major issue in one of my work projects.)
- I renamed React.js to React because we're in 2019 ;)
- A little bit of formatting from prettify
- Maintenance on a few dependencies but nothing important or impactful


jwinn (approved):
LGTM. 💯

julienben:
Thanks @jwinn!

PR #2760, opened 2019-09-25
Moving root path to the main server block in nginx config

julienben:
With the addition of the 'Do not cache sw.js' block, the root path, that is in the 'Always serve index.html' block, should move to the main server block or the sw.js will not be found.

(Copied from @Mensae in PR #2047. Must've gotten lost when we did some work on our Git setup a while back.)

### DIFF
--- a/app/.nginx.conf
+++ b/app/.nginx.conf
@@ -73,11 +73,11 @@ server {
 # To let nginx use its own DNS Resolver
 # resolver <IP DNS resolver>;
 
+# Set path
+  root /var/www/;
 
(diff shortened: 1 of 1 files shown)


damminhtien (approved):

julienben:
Thanks @damminhtien!

julienben:
I can't merge w/o a review from someone with write access though... @gretzky You back from Europe? ;)

vineyardbovines (approved):
I am back @julienben !

Issue #2776, opened 2019-10-12
Initial setup fails


yuvalsuede:
## Description
Running `npm run setup` fails during installation. 

## Steps to reproduce
Run `npm run setup`

**Expected behavior**
After installing the missing module, it keeps failing on other parts. 

**Output**
```
> node ./internals/scripts/setup.js

internal/modules/cjs/loader.js:638
    throw err;
    ^

Error: Cannot find module 'compare-versions'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
```

## Versions

- React-Boilerplate: version": "4.0.0",
- Node/NPM:` v10.16.3 / 6.9.0 `
- OS: MacOS 10.14.6

vineyardbovines:
Interesting find, I'll try and take a look at this soon @yuvalsuede -- please ping me in a couple of days if you don't hear from me, I'm just coming back from a long trip.

julienben:
Hey @yuvalsuede,

This is a strange issue because it only seems to have happened for a few users. I can't reproduce it even on a clean setup. Nonetheless, I believe it will be fixed in PR #2783. I added `compare-versions` to the list of dependencies in the `presetup` script (line 29 of `package.json`).

julienben:
Thanks for reporting! Will close this for now as I'm assuming you worked around it on your end.

PR #2783, opened 2019-10-19
Setup updates

julienben:
## React Boilerplate

Changes:
- Add compare-versions to presetup script (Fix for #2776)
- Removed useless template app file (copied from the main app by the clean script)
- Centralized getting required npm/node versions
- Update npm version check script
- Extract git-related setup script functions to a separate file
- Make repo deletion message clearer
- Minor formatting updates


internals/scripts/check-npm-version.js
```diff
@@ -0,0 +1,25 @@
+const chalk = require('chalk');
```
jwinn: If this is used in npm `preinstall`, we really cannot use `chalk` here, as there may be no npm modules installed yet 😉

jwinn (changes requested):
Fails to install if no npm modules installed (specifically `chalk`)

`rm -rf node_modules && npm install`

internals/scripts/check-npm-version.js
julienben: Good point. Same thing for `compare-versions` on line 3. :(

I think I thought of this and just told myself it shouldn't be an issue since we tell users to start with the setup script anyway. (And the setup script has presetup which installs the dependencies for the setup process.)

But yeah this might not be good enough in this situation. If at any point in development a dev needs to delete `node_modules`, they won't be able to `npm install` afterwards.

Is the only solution to avoid usage of `chalk` and `compare-versions`? Because the idea behind using `compare-versions` was that the old logic was somewhat broken. (`parseFloat(stdout)` is not semVer-compatible.)

jwinn: I'm actually not convinced of the value, as a whole, to even checking if the NPM version is there. Why aren't we checking node, and/or other aspects to ensure the user's env is up to spec?

To me, `package.json` denotes minimum reqs in `engines`, and whilst NPM has offered, in the past, `engine-strict` to error, instead of warn, it just seems like another piece the maintainers, of this bp, have to keep up-to-date, for very little upside.

If we decide to keep the check, then we have to accept the minimal check we use, with either `parseFloat` or `parseInt(stdout, 10)`, to check major version reqs, as a means to reduce the number of issues posted due to NPM requirements not being met. We cannot rely on existing NPM packages to be in place, because these packages may not work themselves on earlier/unsupported--by this bp--versions of NPM.

Just my thoughts...

julienben: Good points. I think I'll push a PR to remove it.

julienben: (I meant push a commit, not a PR. It's ready.)

jwinn: Nice work on this. 👍

jwinn (approved):

PR #2793, opened 2019-10-25
Hot reloading updates?

julienben:
## React Boilerplate

In a few situations, hot reloading isn't working as expected.

For example, if you update one of the messages JSON files, the app reloads completely.

There's also been a lot of updates to the hot reloading tools lately, in particular to deal with Hooks.

I thought I'd give a shot at updating the hot reloading setup but... TBH I have no idea what I'm doing here.

Can anyone help?


julienben:
One of my open questions is do we still need `webpack-hot-middleware`? I don't know where it fits with `react-hot-loader` also in the mix.

vineyardbovines:
I'm so far out of the web game currently that I don't think I could even answer this question 😞

Can-Sahin:
@julienben whats the problem here? Is there something not working. I can maybe help if you shortly what is not working right now with this pull request? 

(I spent some good time on hot loaders for typescript in the past so i can give it a a shot)

julienben:
I'll be honest I'm not sure exactly what's broken. It's just not behaving as you would expect it to (automatic reloading of sagas/reducers/components/language files).

Hot reloading is a bit of a mystery to me. I was hoping someone who has a global understanding of 1. how it works but also 2. how to set it up in a Webpack/React/Redux environment like ours could come in and take a look.

Issue #2806, opened 2019-11-18
Switch Redux logic setup to use Redux Toolkit (take 2)


markerikson:
**Is your feature request related to a problem? Please describe.**
Hi, I'm a Redux maintainer. I'd like to offer a couple suggestions.

The current Redux example implementation shows use of reducers with "hand-written" immutable update logic, and has a somewhat complex store configuration.  In addition, the project's current "feature folder" structure results in multiple separate files for the Redux logic in each feature.

**Describe the solution you'd like**

The project should switch to using the official [Redux Toolkit package](https://redux-toolkit.js.org) to implement the Redux logic.

The project should also consider switching to use the "ducks" pattern within each feature folder.

**Describe alternatives you've considered**
N/A

**Additional context**

Redux Toolkit is our new official, opinionated, batteries-included toolset for efficient Redux development.  It includes utilities to simplify several common Redux use cases, including store setup, defining reducers, immutable update logic, and even creating entire "slices" of state at once without writing any action creators or action types by hand.

We are recommending RTK as the default standard way to write Redux logic.  Partly related to that, we are also recommending that developers should prefer to use "feature folders" or "ducks", and RTK makes it particularly easy to write "ducks"-structured Redux code.  (Reference: the work-in-progress [Redux "Style Guide" docs page](https://deploy-preview-3626--redux-docs.netlify.com/style-guide/style-guide), and [the RTK "Usage Guide" page](https://redux-toolkit.js.org/usage/usage-guide#exporting-and-using-slices).)

Since many developers will be starting their projects using `react-boilerplate`, it would be beneficial if Redux Toolkit was included as part of the default project setup, as it will greatly simplify the application code they write.

markerikson:
As an additional observation: the current use of Reselect borders on being wrong.  In particular:

```js
// selectors.js
const makeSelectCurrentUser = () =>
  createSelector(
    selectGlobal,
    globalState => globalState.currentUser,
  );

// elsewhere
const stateSelector = createStructuredSelector({
  currentUser: makeSelectCurrentUser(),
});
```

The only time you need to create multiple instances of a selector is when [it will be called with multiple varying inputs that cause memoization to fail](https://blog.isquaredsoftware.com/2017/12/idiomatic-redux-using-reselect-selectors/#advanced-optimizations-with-react-redux), such as list items that select their own data item from state by ID.

[...]

Issue #2775, opened 2019-10-11
redux-immer: one time only produce.


salvoravida:
**Is your feature request related to a problem? Please describe.**
no

**Describe the solution you'd like**
improve performance and use standard reducers

**Describe alternatives you've considered**
nothing

**Additional context**
https://github.com/salvoravida/redux-immer

Hi, i want to suggest redux-immer:
`redux-immer` is used to create an equivalent function of Redux combineReducers that works with immer state. Like redux-immutable but for immer

Features:
*one time only produce
*standard reducers

If you think it is a good idea, i could work on a PR.
Salvo

vineyardbovines:
Please fill out the template with a more in-depth description of what you want to see and how you would implement it.

julienben:
I like the idea that this removes code but:
- Do we want to completely hide `immer` and the `produce` function? I feel like that's a step too far. You might forget that you're using the `draft` and that it doesn't work exactly the same as if you were actually mutating your state.
- The generators are there to write this code for you anyway.
- This library isn't tested.

salvoravida:
@julienben i agree that redux-immer will hide produce and draft, but you have cleaner code in every reducer and more important it will invoke one time only "produce"

that said it's your choice. thank you for your answer.

PR #2807, opened 2019-11-18
Feature/add redux toolkit

markerikson:
## React Boilerplate

## Details

This PR demonstrates how to convert part of the built-in example application to use Redux Toolkit, as described in #2806 .

It is not meant to be complete and merged as-is, but rather to serve as a starting point for conversion if the maintainers choose.

### Changes

- Added Redux Toolkit
- Removed the explicit dependency on Redux, since RTK re-exports everything.  (I also attempted to remove the Immer and Reselect dependencies, but saw that there were other reducers that currently depend on Immer, and a couple components that depend on Reselect's `createStructuredSelector`, so I re-added those explicit dependencies.)
- Converted the store setup logic in `configureStore.js` to use RTK's `configureStore` method
- Added a new `containers/App/appSlice.js` file that contains equivalent logic to the existing `App/reducers.js` and `App/actions.js` files
- Modified other uses of `App/actions` to use the action creators exported by `appSlice.js`

I have some other changes I'd like to suggest as well, which I'll write up in the issue rather than here.

### DIFF
--- a/app/configureStore.js
+++ b/app/configureStore.js
@@ -40,7 +40,7 @@ export default function configureAppStore(initialState = {}, history) {
 
   const store = configureStore({
     reducer: createReducer(),
-    initialState,
+    preloadedState: initialState,
     middleware: [...getDefaultMiddleware(), ...middlewares],
     enhancers,
   });
--- a/app/configureStore.js
+++ b/app/configureStore.js
@@ -2,23 +2,18 @@
  * Create the store with dynamic reducers
  */
 
-import { createStore, applyMiddleware, compose } from 'redux';
+import { configureStore, getDefaultMiddleware } from '@reduxjs/toolkit';
 import { routerMiddleware } from 'connected-react-router';
 import { createInjectorsEnhancer, forceReducerReload } from 'redux-injectors';
 import createSagaMiddleware from 'redux-saga';
 import createReducer from './reducers';
 
-export default function configureStore(initialState = {}, history) {
-  let composeEnhancers = compose;
+export default function configureAppStore(initialState = {}, history) {
   const reduxSagaMonitorOptions = {};
 
   // If Redux Dev Tools and Saga Dev Tools Extensions are installed, enable them
   /* istanbul ignore next */
   if (process.env.NODE_ENV !== 'production' && typeof window === 'object') {
-    /* eslint-disable no-underscore-dangle */
-    if (window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__)
-      composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({});
-
     // NOTE: Uncomment the code below to restore support for Redux Saga
     // Dev Tools once it supports redux-saga version 1.x.x
     // if (window.__SAGA_MONITOR_EXTENSION__)
@@ -37,18 +32,18 @@ export default function configureStore(initialState = {}, history) {
   const middlewares = [sagaMiddleware, routerMiddleware(history)];
 
   const enhancers = [
-    applyMiddleware(...middlewares),
     createInjectorsEnhancer({
       createReducer,
       runSaga,
     }),
   ];
 
-  const store = createStore(
-    createReducer(),
+  const store = configureStore({
+    reducer: createReducer(),
     initialState,
-    composeEnhancers(...enhancers),
-  );
+    middleware: [...getDefaultMiddleware(), ...middlewares],
+    enhancers,
+  });
 
   // Make reducers hot reloadable, see http://mxs.is/googmo
   /* istanbul ignore next */
--- a/app/containers/App/appSlice.js
+++ b/app/containers/App/appSlice.js
@@ -1,0 +1,38 @@
+import { createSlice } from '@reduxjs/toolkit';
+import appReducer from './reducer';
+
+// The initial state of the App
+export const initialState = {
+  loading: false,
+  error: false,
+  currentUser: false,
+  userData: {
+    repositories: false,
+  },
+};
+
+const appSlice = createSlice({
+  name: 'app',
+  initialState,
+  reducers: {
+    loadRepos(state, action) {
+      state.loading = true;
+      state.error = false;
+      state.userData.repositories = false;
+    },
+    reposLoaded(state, action) {
+      const { repos, username } = action.payload;
+      state.userData.repositories = repos;
+      state.loading = false;
+      state.currentUser = username;
+    },
+    repoLoadingError(state, action) {
+      state.error = action.payload;
+      state.loading = false;
+    },
+  },
+});
+
+export const { loadRepos, reposLoaded, repoLoadingError } = appSlice.actions;
+
+export default appReducer;
--- a/app/containers/App/appSlice.js
+++ b/app/containers/App/appSlice.js
@@ -11,11 +11,12 @@ export const initialState = {
   },
 };
 
+/* eslint-disable default-case, no-param-reassign */
 const appSlice = createSlice({
   name: 'app',
   initialState,
   reducers: {
-    loadRepos(state, action) {
+    loadRepos(state) {
       state.loading = true;
       state.error = false;
       state.userData.repositories = false;
--- a/app/containers/HomePage/saga.js
+++ b/app/containers/HomePage/saga.js
@@ -3,8 +3,11 @@
  */
 
 import { call, put, select, takeLatest } from 'redux-saga/effects';
-import { LOAD_REPOS } from 'containers/App/constants';
-import { reposLoaded, repoLoadingError } from 'containers/App/actions';
+import {
+  loadRepos,
+  reposLoaded,
+  repoLoadingError,
+} from 'containers/App/appSlice';
 
 import request from 'utils/request';
 import { makeSelectUsername } from 'containers/HomePage/selectors';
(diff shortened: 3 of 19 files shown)


markerikson:
Looking at the Travis logs, it appears that the tests themselves are passing, but the global code coverage stats have fallen below a 98% threshold.

vineyardbovines:
Hey it's the redux overlord! Thanks for this, tbh I have no idea what redux-toolkit even is so this will be cool to take a look at.

app/containers/App/appSlice.js
```diff
@@ -0,0 +1,39 @@
...
+export const initialState = {
+  loading: false,
+  error: false,
+  currentUser: false,
+  userData: {
+    repositories: false,
+  },
+};
+
+/* eslint-disable default-case, no-param-reassign */
```
vineyardbovines, line 14: Could you talk more about the way the param reassignment works here? I'm not sure how I feel about turning off a lint rule to accommodate the way a package works. I also think that as I'm reading this, it comes off as if we're mutating state here by reassigning the state params. This might just be me not fully understanding the way this works under the hood.

markerikson: Sure.

RTK specifically uses the [Immer library](https://immerjs.github.io/immer/docs/introduction) library inside.  Immer lets you write "mutating" code, but it wraps the original value in an ES6 Proxy and tracks the modifications, and then safely returns a correctly immutably updated value (as if you'd done all the spreads and stuff by hand).

`react-boilerplate` is already using Immer in its reducers, and in fact I copied the `eslint-disable` line from another file, like the equivalent `App/reducer.js`:

https://github.com/react-boilerplate/react-boilerplate/blob/27538f1595620b235eae2ae4251305da82dc386a/app/containers/App/reducer.js#L23-L30

So, in that sense there's nothing "new" here for `react-boilerplate`, this is already stuff you've got in the codebase.  The only changes are A) the additional `@reduxjs/toolkit` package, B) the `createSlice()` function auto-generating the action creators and action types, and C) consolidating the resulting logic into one file instead of several.

app/configureStore.js
```diff
@@ -2,23 +2,18 @@
...
-  let composeEnhancers = compose;
+export default function configureAppStore(initialState = {}, history) {
   const reduxSagaMonitorOptions = {};
 
   // If Redux Dev Tools and Saga Dev Tools Extensions are installed, enable them
   /* istanbul ignore next */
   if (process.env.NODE_ENV !== 'production' && typeof window === 'object') {
-    /* eslint-disable no-underscore-dangle */
-    if (window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__)
-      composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({});
```
julienben, line 20: This is no longer necessary for the Redux dev tools to work? Or is it simply handled inside Redux Toolkit?

If it's the latter, the parent `if` statement is now empty. Since the saga dev tools don't seem to be maintained + they have limited usability, would you mind just deleting that whole `if`?

(Same in the internals version of this file.)

app/configureStore.js
```diff
@@ -37,18 +32,18 @@ export default function configureStore(initialState = {}, history) {
...
   ];
 
-  const store = createStore(
-    createReducer(),
-    initialState,
-    composeEnhancers(...enhancers),
-  );
+  const store = configureStore({
+    reducer: createReducer(),
+    preloadedState: initialState,
```
julienben, line 43: Is there a rationale for the API using `preloadedState` instead of `initialState`? Is this something we should change in `configureAppStore` and other places too?

(Saw your tweet regarding this btw, we have an ongoing discussion for an eventual move to TS.)

app/containers/HomePage/saga.js
```diff
@@ -20,7 +23,7 @@ export function* getRepos() {
   try {
     // Call our request helper (see 'utils/request')
     const repos = yield call(request, requestURL);
-    yield put(reposLoaded(repos, username));
+    yield put(reposLoaded({ repos, username }));
```
julienben, line 26: 👌👌👌

app/containers/App/appSlice.js
julienben: Yes, we had to disable these rules for our migration to `immer`. You should only need `/* eslint-disable no-param-reassign */` here though as `default-case` is only for switches.

julienben (commented):
This looks great!

As you mention, it's a WIP and requires more work before being ready but unless somebody would like to argue against this switch I think we can safely proceed with the additional changes required for it.

@markerikson Would you like to continue this work or would you like local contributors to take it on?

Quick list of things which still need to be done:
- Obsolete actions/reducers/constants files need to be deleted along with their tests
- The container generators need to be updated
- appSlices need to be tested (+ 100% coverage recovered if it still makes sense)
- Documentation needs to be updated with IMO references to the [usage guide](https://redux-toolkit.js.org/usage/usage-guide)
- Consider if we can actually remove `reselect` and `immer` as dependencies. This would make sense since Redux Toolkit provides its own version of each.
- For myself or anyone else who wants to take it on: Implement the `reselect` recommendations in [this comment](https://github.com/react-boilerplate/react-boilerplate/issues/2806#issuecomment-554855162) after we merge this.

app/configureStore.js
markerikson: RTK already handles this internally.  Our `configureStore()` uses the `redux-devtools-extension` package, which wraps the global window function lookup.

app/configureStore.js
markerikson: Naming-wise, I think the earlier Redux docs did refer to the `createStore` signature as `createStore(reducer, initialState)`.  At some point there was a discussion about intended use cases, and the argument was that "this is really meant for loading state that was persisted or something", so we came up with `preloadedState` as the name.  That name is used in the docs:

https://redux.js.org/recipes/structuring-reducers/initializing-state#initializing-state

https://redux.js.org/api/createstore

So, I reused that name with RTK's `configureStore`.

app/containers/HomePage/saga.js
markerikson: This is due to the behavior of `createAction()`.  By default, generated action creators accept a single argument, which becomes `action.payload`.  That does put the responsibility on the dispatching code to format the payload contents correctly.

If a developer wants to do some formatting of the payload contents, such as generating a unique ID or turning multiple action creators args into a payload, they can define a "prepare callback":

https://redux-toolkit.js.org/api/createaction#using-prepare-callbacks-to-customize-action-contents

https://redux-toolkit.js.org/api/createslice#customizing-generated-action-creators

markerikson:
I'd like to turn the PR over to you folks at this point.  You know the `react-boilerplate` codebase and requirements way better than I do, and I have my own priorities atm (like [rewriting most of the Redux docs](https://github.com/reduxjs/redux/issues/3592)).  

Happy to answer any questions about using RTK as you work through this, though!

It's likely that you can remove Immer as a direct dependency once the other reducers have been converted.  Reselect might not be removable, given that there are uses of `createStructuredSelector`, which RTK does not re-export.  (Technically you can get away with `import {createStructuredSelector} from "reselect"` at a code level with only having an explicit dependency on RTK, but that's dependent on the bundler + package management tool letting you do that, and there's currently an ESLint rule that yells at you for doing that anyway.)

app/containers/App/appSlice.js
vineyardbovines: Word, thanks for the explanations. I like to make my day-to-day life difficult by not really ever using immutability libs, so this is helpful.

justingreenberg: lol yes @gretzky i'm with you—super clever implementation but personally not a fan of immer...setting properties using dot notation just _feels_ wrong. long live explicit immutability 😄

justingreenberg:
hi @markerikson thank you for this PR! 

> Reselect might not be removable, given that there are uses of `createStructuredSelector`, which RTK does not re-export

i don't think it's a huge deal to keep reselect dependency, but would you consider adding `createStructuredSelector` to RTK exports? maybe in a utility export with other commonly used tools from libs, such as redux's compose?

app/containers/App/appSlice.js
markerikson: The good news is that using RTK and Immer doesn't _require_ that you actually use "mutative" logic - you can still return manually written immutable updates.

That said, frankly hand-written immutable update code is overly verbose, hard to read, and very prone to mistakes.  I understand the concerns about this "feeling wrong", but using Immer is actually much safer _and_ easier to read.

markerikson:
We do re-export everything from Redux already.  

I'd like to avoid explicitly re-exporting `createStructuredSelector` for now.  If someone really needs it, they can just declare the dependency on Reselect themselves.  Since RTK already pulls that in, it's just adding another entry to `package.json`.

app/containers/App/appSlice.js
julienben: I don't think there's anything inherently wrong with dot notation for writing immutable updates but hiding `immer` can IMO make things confusing for new users. They might not realize that they're not **actually** mutating the state.

I was previously against this when [someone suggested switching to `redux-immer`](https://github.com/react-boilerplate/react-boilerplate/issues/2775#issuecomment-542213763). I want to make sure that we make it as clear as possible (via docs and/or comments) that there's some work in the bg making it possible to write immutable logic with dot notation.

julienben:
I'm merging this into a React Boilerplate branch called `redux-toolkit` so we can continue working there.

julienben:
See #2808 for ongoing work into this.

app/containers/App/appSlice.js
markerikson: Yeah, the teaching aspect is definitely the hardest part here.

PR #2808, opened 2019-11-19
Add Redux Toolkit and ducks pattern

julienben:
## React Boilerplate

Follows the work done by @markerikson in #2807. This branch will be used to finish the switch to Redux Toolkit and the ducks pattern.

### Changes

- Added Redux Toolkit (RTK)
- Removed the explicit dependency on Redux, since RTK re-exports everything.
- Converted the store setup logic in configureStore.js to use RTK's configureStore method
- Updated the sample and template apps:
  - Created an RTK "slice" for all the HomePage and LocaleToggle features
  - Moved App container's reducer/saga/actions into the HomePage
  - Deleted all redundant actions/constants/reducer files
  - Determined ownership of repos in saga so that RepoListItem can become a component
  - Updated all broken tests and write new tests for slices
- Added ESLint exception for "slice" files so that we can remove ESLint exceptions

### To Do

- [x] Obsolete actions/reducers/constants files need to be deleted along with their tests
- [x] appSlices need to be tested
- [x] Update the template app
- [ ] Update Container generator
- [ ] Restore 100% coverage
- [ ] Update documentation with references to the [Redux Toolkit usage guide](https://redux-toolkit.js.org/usage/usage-guide)
- [ ] Consider if we can actually remove `reselect` and/or `immer` as dependencies.
- [ ] Implement the `reselect` recommendations in [this comment](https://github.com/react-boilerplate/react-boilerplate/issues/2806#issuecomment-554855162).


julienben:
@gretzky @justingreenberg @jwinn @BenLorantfy This PR is becoming dauntingly big. Would you like to review it as is, squash it, and then I can finish the work in another branch?

As you can see from my message up top, there's still quite a lot to do. Mainly documentation and generator updates.

justingreenberg (approved):
nice work @julienben as always, i haven't had time to pull down and test locally yet, but LGTM for merge into RTK branch when checks are green 👍

jwinn (approved):
Agree. Great work @julienben. I haven't had time to test locally, but can do so in a few hours, if you'd like. Otherwise, LGTM 👍

app/containers/HomePage/slice.js
```diff
@@ -0,0 +1,57 @@
...
+
+// The initial state of the HomePage container
+export const initialState = {
+  username: '',
+  repositories: [],
+  loading: false,
+  error: false,
+};
+
+const slice = createSlice({
```
BenLorantfy, line 21: Maybe this is a bigger conversation that should be resolved outside this PR, but I've found doing redux per-page like this makes it hard to re-use state on another page without an awkward coupling between pages. 

i.e. If I need to access the repos on a different page, say "ReposPage", then I have to either duplicate this slice, or import this slice. Then you end up with ReposPage importing from HomePage, which feels awkward.

Alternatively, I believe it's common to have a "features" folder. This could have a "repos" slice; then any number of pages can import from this slice. Another option is what I described in [this comment](https://github.com/react-boilerplate/react-boilerplate/issues/2806#issuecomment-555195354), but that's kind of opinionated and not common.

BenLorantfy (commented):
Looks good to me overall, left one comment that might be a bigger architecture change that we could resolve outside this PR if you'd like.

app/containers/HomePage/slice.js
markerikson: Yeah, notionally the state tree and the component tree are meant to be somewhat independent of each other.  Any number of components could be depending on a given chunk of data in the store, and the state tree isn't necessarily going to be shaped the same way as the component tree.

I've seen some folks talk about having "data layer" folders which just have Redux logic in them, and then there _may_ be more Redux logic with specific "pages" or "containers" if appropriate.  That sounds similar to what you're describing, I think.

That said, I personally don't have any qualms about different slices importing each other in terms of "privacy" or anything.  You're basically acknowledging that there is an explicit dependency between these different parts of the application logic, and that not everything is going to be perfectly isolated and encapsulated.  Note that if you take that far enough, [it _is_ possible to end up with circular reference import problems](https://redux-toolkit.js.org/usage/usage-guide/#exporting-and-using-slices).

Organizing information is _hard_ :)

julienben: Thanks for the comments! Makes a lot of sense. Here's my summary:

**What we had:** Repos management happening inside the App container. Injected as `globalState` and `globalReducer`. This felt clunky to me because it could grow infinitely and I don't think we explain this pattern anywhere in the docs anyway.

**What I tried to do:** Move repos management to the HomePage. Admittedly not a great idea. For all the reasons mentioned above.

**What I ended up doing in [my last commit](https://github.com/react-boilerplate/react-boilerplate/pull/2808/commits/ce63b7e0134ae01504474ac19cd6580813a812f5):** Move repos management to a ReposManager as described by @BenLorantfy. Didn't go all the way to create a `managers` folder because that should indeed be a bigger conversation with accompanying doc updates. I just wanted to at least move repos management outside of the App container. ReposManager is still rendered inside the App (i.e. "globally") so functionality hasn't changed.

Makes sense to everyone? If yes, please re-approve? 🙏 I've updated the base branch to `redux-toolkit-switch` so that we can merge this project in full once it's done.

CC @jwinn @justingreenberg

julienben: Oops no need for re-approvals. Will wait to merge in case anyone has comments though.

justingreenberg: @julienben this is awesome, much cleaner 👏 👍 LGTM!!

justingreenberg (approved):

app/containers/HomePage/slice.js
jwinn: @julienben great work! Should we resolve this thread and create a new discussion for potential changes of structure?

BenLorantfy (approved):
Looks great 👍

app/containers/HomePage/slice.js
BenLorantfy: Looks great 👍

Issue #2255, opened 2018-06-26
Hot Module Replacement with Typescript


Can-Sahin:
## Issue Type

How to enable HMR in Typescript ?

## Description

I turning this into Typescript slowly since I couldn't find any recent forks but I have failed to get HMR working with awesome-typescript-loader.  

I tried using react-hot-loader with babel

So, added tsx rule to webpack
```javascript
module: {
  rules: [
    {
      test: /\.tsx?$/,
      use:     {
         loader: 'babel-loader',
         options: {
             babelrc: false,
             plugins: ['react-hot-loader/babel'],
         },
       },
       'awesome-typescript-loader',
    },
    ...
```
When I update my React.Component I can see the console messages like
```
....
[HMR]  - ./app/containers/App/index.tsx
[HMR] App is up to date.
```

But nothing is changed. I tried react-hot-loader issues and their solutions none seem to work on this boilerplate. I recently started to this, so I cant fully understand what goes on under the hood with webpack and dll etc...

**My questions:** 

- Is react-hot-loader is required in this boilerplate or is webpack.HotModuleReplacementPlugin enough simply? 
- I can see after the component changes typescript does NOT transpile the code but it should, shouldn't it? Because I make changes to .tsx files.  What am I missing? 

# Versions

- React-Boilerplate 3.6.0
- Node/NPM: 9.6.1
- Browser: Chrome

gretzky:
Hey @Can-Sahin, thanks for filling out the issue template. Unfortunately, we try to keep this issue channel for issues with this bp only rather than help forums-- we also don't support TypeScript out of the box so it's harder for us to debug. I suggest checking out Stack Overflow or other forums to see how others have done it.

Can-Sahin:
@gretzky  Ok. I kinda figured that, but I went through so much similar issues etc but no solution. Wanted to check here also. If any1 seeing this has an answer I will appreciate it alot. Excellent boilerplate though :)

julienben:
I found [this article](https://blog.logrocket.com/setup-react-hotloader-in-10-minutes-3175dfdbf38a) which seems to explain pretty well how to switch to react-hot-loader. If you end up making the switch yourself, you would be more than welcome to suggest is here as a PR. Without the TypeScript portions though ;)

PR #2810, opened 2019-11-20
Hot reloading issues fixed

Can-Sahin:
- Moved react-hot-loader to the top component
- i18n seperated into two for components to access directly without creating dependency to json files.
- ConnectedApp function bug fixed.

Following the hot-loading issues at #2793 @julienben 

I worked on react-hot-loader a year ago as I was asking questions on how to do it #2255 here :) My experience on it I can tell several things

1) React Hot Loader isn't necessary actually. HMR works just as fine without it. It helps you to preserve states only. [See this](https://github.com/gaearon/react-hot-loader/issues/920#issuecomment-547731443). With hooks in the game it might be be needed more so included in this pull request. I didn't keep track on what is changed in hot loaders lately. So considering it might be helpful it won't hurt to use it. 

2) Changes in the json translation files also aren't hot loaded in master branch either. When I debugged it, it turns out React Components depending on `i18n.js` file ruining it. Because json files are updated and HMR sees that LanguageProvider uses it but `LanguageProvider` doesn't declare how to handle it. So i separated dependencies into two files. Now `app.js` only depends on `i18n.js` and `module.hot.accept` expects it.

3) Issue with lazy loading (code-splitting). With the current code splitted components react-hot-loader has no benefit over webpack HMR. Because of [this](https://github.com/gaearon/react-hot-loader#out-of-bound-warning). In my tests when the lazy component is hot loaded it does render the component again firing `useEffect((), [])` hook, and other non-zero dependency hooks. Basically what HMR does alone. If you make components not lazy loaded it works super fine. Zero dependency hooks aren't fired.

Note: I just added a file called `locales.js` to separate the dependency on `i18n.js` file. Better naming could be possible or putting it in some file else.


julienben (commented):
@Can-Sahin Thanks a lot for your help on this! I tried this locally and it works MUCH better than my attempt at it.

A few thoughts/questions:
1. I'm not sure I 100% followed your comments regarding react-hot-loader. Do you recommend removing it and sticking to Webpack HMR only? The only downside is we would lose Hooks support?
2. I changed your base branch to `dev` because if we merged first to my branch and then mine into `dev`, it would mark me as the only contributor of this work.
3. I noticed that changes to `en.json`, etc still trigger a full page reload. There's nothing we can do about that right? It's not a huge deal if we can't fix that cause this is already much better than what we currently have.
4. Do you mind fixing the language generator? Right now it adds the new language to `i18n.js` instead of the new file `locales.js`.

Can-Sahin:
**Number 1**: A year ago I was troubled with it too and decided not to use it. Since HMR almost does the same job. But RHL has benefits like hooks states etc. The catch here is we cannot see those benefits with lazy loaded components. My opinion we keep RHL in case people are not lazy loading their components (not dynamically importing) and they wanna use full featured hot reloading stuff with hooks.  
ps: RHL's `react-dom` extension displays the hot loading errors in the view instead of the console. Which is much better because people will immediately see the errors if hot reloading isn't working as expected.

**Number 2**: Thanks :))

**Number 3**: For me I think its working as like other hot loaded components

Hmm I dont see it happening. I see console logs like
```
[HMR] Checking for updates on the server...
process-update.js?e135:125 [HMR] Updated modules:
process-update.js?e135:127 [HMR]  - ./app/translations/en.json
process-update.js?e135:127 [HMR]  - ./app/i18n.js
process-update.js?e135:132 [HMR] App is up to date.
```
It keeps the state. For instance github user input field is preserved. You will see a github users reloading action because of the useEffect hook. As I mentioned due to the lazy loaded components hooks fire again and causes load_repo action. It gives a feeling of full page reload. Can you double check ? (or do you mean by 'full page reload' like re-rendering the component? For me, it means like refreshing a browser page)

**Number 4**: Ok sure i will. I missed that totally ;)

julienben (changes requested):
Ah yes thanks for thinking of the `extract-intl` script! I wouldn't have thought about it. 😅 I was thinking of this file though:

`/internals/generators/language/index.js`

Some of the actions listed are made in `i18n.js` but some of them should be in `locales.js` now.

Can-Sahin:
oh yeah i missed that tiny changes. I just did :) 

CI build fails because it says there is 'fr' option added to test snapshots. I cleaned all my changes before the commit. How could it land there actually. I don't know anything about testing with snapshots. All I did was accidentally committing to the branch without pulling. So pulled and merged and pushed afterwards.

julienben:
If you look at `.travis.yml`, you'll see that we run `generate-templates-for-linting.js` before testing in Travis. It's a way to check that the generators are putting out valid code. One of the steps there is to add 'fr' as a language. After we finish though, the files are restored as they were before. 

image

As you can see it's not backing up and restoring `locales.js`, hence the error you're seeing in Travis. I'm not sure where we decide which files to back up and restore. Looking into it on my side but either way it'll be an easy fix.

julienben:
Got it. This is the fix:

image

Do that and Travis should work.

julienben (approved):
Excellent work @Can-Sahin!

PR #2811, opened 2019-11-21
[REDUX TOOLKIT] Update generators

julienben:
## React Boilerplate

Continuing the work done in #2807 and #2808. The base branch is `redux-toolkit-switch` where we're aggregating all the work for the switch. This PR contains the generator updates.

PR #2812, opened 2019-11-21
Switch to Redux Toolkit and ducks pattern

julienben:
## React Boilerplate

This is a draft PR which aggregates all the changes as we work towards switching to Redux Toolkit and the ducks pattern.

Related issue:
- #2806

### Changes so far

- Added Redux Toolkit (RTK)
- Removed the explicit dependency on Redux, since RTK re-exports everything.
- Converted the store setup logic in configureStore.js to use RTK's configureStore method
- Updated the sample and template apps:
  - Created an RTK "slice" for all the HomePage and LocaleToggle features
  - Moved App container's reducer/saga/actions into a new container called ReposManager
  - Deleted all redundant actions/constants/reducer files
  - Determined ownership of repos in saga so that RepoListItem can become a component
  - Updated all broken tests and write new tests for slices
- Added ESLint exception for "slice" files so that we can remove ESLint exceptions

### To Do

- [x] Obsolete actions/reducers/constants files need to be deleted along with their tests (#2807/#2808)
- [x] appSlices need to be tested (#2807/#2808)
- [x] Update the template app (#2807/#2808)
- [x] Update Container generator (#2811)
- [x] Restore 100% coverage (#2814)
- [ ] Update documentation with references to the [Redux Toolkit usage guide](https://redux-toolkit.js.org/usage/usage-guide)
- [ ] Consider if we can actually remove `reselect` and/or `immer` as dependencies.
- [ ] Implement the `reselect` recommendations in [this comment](https://github.com/react-boilerplate/react-boilerplate/issues/2806#issuecomment-554855162).


joejordan:
Just a personal opinion, but part of the appeal for me in adopting React Boilerplate over others was the comparative lack of "magic" that the framework handled, combined with what appeared to me as the implementation of industry best practices in organization and code.

In that spirit, I'm not sure I agree with the approach of incorporating another library that obfuscates some of Redux's implementation details. I personally like seeing how a store is created and configured with bare Redux, as it offers both a learning opportunity and the ability to customize to one's heart's content.

I am less familiar with the "ducks" pattern discussed here, but the existing pattern/layout of action.js, reducer.js and optionally saga.js with a well-identified constants.js to define how the action appears helps to clearly delineate the mechanisms inherent in Redux.

Can-Sahin:
Here I don't understand why there is a `ReposManager`. I am pretty familiar with RTK now and with ducks pattern too. Only `Homepage` is using `ReposManager` so it can be easily its own slice. No other component is using it. I am super against to using `managers`. I am backend developer and it's a really bad pattern to start calling things managers and putting them into separate folders. People walked away from it years ago

app/containers/ReposManager/slice.js
```diff
@@ -0,0 +1,50 @@
...
+  initialState,
+  reducers: {
+    loadRepos(state) {
+      state.loading = true;
+      state.error = false;
+      state.repositories = [];
+    },
+    reposLoaded(state, action) {
+      const { repos } = action.payload;
+      state.repositories = repos;
```
afcjunior, line 34: Mutating state directly is an antipattern. You should return { ...state, repos }

AlexSSE: RTK uses `immer` inside. So you can use it this way

Tommoore96:
@all-contributors please add Tommoore96 for code

Can-Sahin:
Edit: Sorry I opening this back as a result of not merging the #2935 if that means anything;)

package.json
```diff
@@ -60,6 +60,7 @@
   },
   "dependencies": {
     "@hot-loader/react-dom": "16.11.0",
+    "@reduxjs/toolkit": "1.0.4",
```
cahna, line 63: `@reduxjs/toolkit` includes redux-thunk by default. Not sure if webpack shakes-out this dependency if unused. IMO, toolkit is too opinionated (and hiding `immer` is too much "magic" for my taste). The ducks pattern can be implemented just as easily, and with a similar reduction in boilerplate, using a single-purpose library like `redux-actions`. Using `redux-actions` also allows easier transition to TypeScript, if desired (see [`typesafe-actions`](https://github.com/piotrwitek/typesafe-actions))

mjtechnavious:
Is there a chance when this will be release?

Issue #2809, opened 2019-11-20
change react-intl language triggers page component to rerender


petitkriket:
Doing a personnal project and cannot figure why changing of language rerenders the current page component which reset children components.

Is it related to the router and switch ? Language are not hot reloaded like in react-i18next ?

## Description
Changing language rerender current page component and it's childs

## Steps to reproduce
Change language using language selector

**Expected behavior**
Changing a language  rerenders only texts and concerned items

**Screenshots**
![Peek 20-11-2019 18-55](https://user-images.githubusercontent.com/1753019/69264364-78b8dd00-0bc7-11ea-8fc7-7796c03c22c6.gif)

## Versions

- React-Boilerplate: v4.0.0
- Node/NPM:
- Browser:

julienben:
You mention hot reloading so I assume you're talking about development right? If yes, check out PR #2810. We're trying to fix issues with hot reloading there. Can you try to apply these updates to your app and see if it helps?

petitkriket:
Thank you for your quick reply.

Sorry for the lack of clarity, I am very new to React. 

It's not only about dev environement. I would like the strings to update without re-rendering the whole current displayed component on language update but only the translated parts (strings) as it did with react-i18next + create-react-app in a previous attempt.

Is it possible with react-intl or should I use react-i18next instead ?

Can-Sahin:
Can you send the console logs in the browser ? 

When you change the json file console logs should be printed starting with this text
```
[HMR] Checking for updates on the server...
```

julienben:
@Can-Sahin I don't think that's relevant. It's not a code change so there's no reason for HMR to be visible in the console.

The issue seems to be with changing the locale using the LocaleToggle. Whether in dev or production, the entire page rerenders. I see the same thing in the sample app, both in `dev` and `master` so it's not something we broke recently.

Can-Sahin:
Sorry when I read the `Language are not hot reloaded like in react-i18next ?` section. I thought its a hot loading problem. It's actually not. 

I have a quick suggestion can you try deleting `key={locale}` in `LanguageProvider` Component in `containers` folder. See if it works ? 

It should look like this without the key props

```javascript
export function LanguageProvider(props) {
  return (
    <IntlProvider
      locale={props.locale}
      messages={props.messages[props.locale]}
    >
      {React.Children.only(props.children)}
    </IntlProvider>
  );
}

```

[...]

PR #2818, opened 2019-11-24
Fix for full page rerender on language switch

julienben:
## React Boilerplate

As suggested by @Can-Sahin in #2809, removing the key from the `IntlProvider` stopped it from rerendering the full page when a user changes the language. The key doesn't serve any purpose as far as I can see. It must be a remnant from a previous version.


app/containers/LanguageProvider/index.js
```diff
@@ -24,11 +24,7 @@ const stateSelector = createSelector(
 export default function LanguageProvider(props) {
   const { locale } = useSelector(stateSelector);
   return (
-    <IntlProvider
-      locale={locale}
-      key={locale}
-      messages={props.messages[locale]}
-    >
+    <IntlProvider locale={locale} messages={props.messages[locale]}>
```
BenLorantfy, line 24: I believe `key` is here because of https://github.com/formatjs/react-intl/issues/234#issuecomment-163366518 , I've run into the same issue before.  If `react-intl` v3 uses the new context API (see https://github.com/facebook/react/issues/2517#issuecomment-360450920), this can probably be removed.

julienben: Switching locales works just fine without the key even though react-intl v2.9 does use the old Context API.

This issue is from 2015 though so it's probably been resolved since then, unintentionally or not, via the many changes that occurred in react, react-redux and/or react-intl.

And in any case we're switching to v3 which uses the new Context API.

If that makes sense, please approve so I can merge?

BenLorantfy (approved):

app/containers/LanguageProvider/index.js
BenLorantfy: kk 👍

PR #2819, opened 2019-11-24
Maintenance

julienben:
## React Boilerplate

- Lots of updates!
  - Didn't update `react` to v16.12 yet as `@hot-loader/react-dom` is still on v16.11 but will make sure to do it before we release v5.0.
- Synced some changes between the demo and the sample apps
- Fixed hot reloading in the sample app as we forgot to do that in #2810

I ran our full regression tests on this. Everything should be working.


jwinn (approved):
LGTM. Great work! 👍

app/containers/App/selectors.js
```diff
@@ -10,22 +10,13 @@ const selectGlobal = state => state.global || initialState;
 const selectRouter = state => state.router;
 
 const makeSelectCurrentUser = () =>
```
BenLorantfy, line 12: As per https://github.com/react-boilerplate/react-boilerplate/issues/2806#issuecomment-554855162, do we want to change some of these selector factories to regular selectors?

BenLorantfy (approved):
Looks good to me, left a comment we might want to fix before v5 release?

app/containers/App/selectors.js
julienben: Yes it's in the to do list for #2812. The changes you see here are just because prettier wanted them.

PR #2837, opened 2019-12-09
 Log errors or exceptions into .keep file

getspooky:
In order to effectively manage a web server, it is necessary to get feedback about the activity and performance of the server as well as any problems that may be occuring


julienben:
Please review our contribution guidelines and follow them before reopening your PR.

A few issues here off the top off my head:
- Tests don't pass
- Your PR mixes 3 different things and there's no explanation as to what you're trying to change
- Deleted the PR template instead of filling it in
- Not sure the changes themselves make sense. Adding French is cool but 1. it breaks our tests and 2. it's only the demo app so it doesn't really add anything useful.

PR #2846, opened 2020-01-16
Migrate to styled-components v5

balonsom:
This PR upgrades `styled-components` and `jest-styled-components` to the latest major version. The new version has no braking changes.

### DIFF
--- a/app/containers/App/tests/__snapshots__/index.test.js.snap
+++ b/app/containers/App/tests/__snapshots__/index.test.js.snap
@@ -29,6 +29,6 @@ exports[`<App /> should render and match the snapshot 1`] = `
     />
   </Switch>
   <Footer />
-  <GlobalStyleComponent />
+  <UNDEFINED />
 </ForwardRef(App__AppWrapper)>
 `;
--- a/app/components/Footer/tests/__snapshots__/index.test.js.snap
+++ b/app/components/Footer/tests/__snapshots__/index.test.js.snap
@@ -2,7 +2,7 @@
 
 exports[`<Footer /> should render and match the snapshot 1`] = `
 <footer
-  className="Wrapper-wcfdfo-0 dPvBog"
+  className="Wrapper-wcfdfo-0 gkRjxJ"
 >
   <section>
     <span>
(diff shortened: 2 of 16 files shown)


vineyardbovines:
Please fix the broken tests

Can-Sahin:
I see the broken tests are from snapshots. I don't understand this snapshotting with styled-components yet actually. If the class name generated is different then the snapshot fails. I dont know how the unique names are generated either. Now, with this pull request there are no changes but generated class names are different so the snapshot fails. Maybe it because of `v5` but not sure. 

In such cases I think `--updateSnapshot` option should be considered with jest. So snapshots are updated and test wont brake.

vineyardbovines:
It's best practice to use `--updateSnapshots` manually, rather than adding it into the script itself (if that's what you're suggesting?). That allows you to keep track of snapshots breaking, which indicates that for some reason your UI has changed.

My guess with styled-components is that the update caused the classnames to regenerate. I don't know much about that either.

Can-Sahin:
Yep I was suggesting to use `--updateSnapshots` for specific cases like that. Not generally obviously.

Can-Sahin:
Yep I was suggesting to use --updateSnapshots for specific cases like that. Not generally.

app/containers/App/tests/__snapshots__/index.test.js.snap
```diff
@@ -29,6 +29,6 @@ exports[`<App /> should render and match the snapshot 1`] = `
     />
   </Switch>
   <Footer />
-  <GlobalStyleComponent />
+  <UNDEFINED />
```
Can-Sahin, line 32: This is weird ha? Something must be wrong.

Can-Sahin: Looking at the [jest-styled-components issues](https://github.com/styled-components/jest-styled-components/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc) there are some problems with the `v5`. Can you investigate more in this if you wanna merge your pull request. 

Side note:  Tests fail when you just add `import 'jest-styled-components'` into the `containers/App/tests/index.test.js` file which is discussed in the issues. Also there are issues on why classnames are changing too there. So I guess `--updateSnapshots` was a wrong suggestion after all. 

Can you look deeper into this?

muthukumarse:
Hello, is there the possibility of v5 released soon with redux-toolkit?

Can-Sahin:
Edit: Sorry I opening this back as a result of not merging the https://github.com/react-boilerplate/react-boilerplate/pull/2935 if that means anything;)

app/containers/App/tests/__snapshots__/index.test.js.snap
danactive: I noticed a change like this too, I think it was resolved with `jest-cli` v26

PR #2866, opened 2020-02-23
Update react intl v3 - saga based translation files

colbysadams:
When using react-boilerplate in production, I found that all of the translation messages were being bundled together and loaded at the entrypoint.  In the example app, this only wastes ~1KB , but when you're dealing with larger apps with many languages, this wasted overhead can be significant.  

This fix would move the fetching of translation files into a redux-saga, allowing only the necessary translations to be sent to the client.

All commits for updated react-intl in the pull request #2721 submitted by @mhatvan are merged with the dev branch and then I made the following changes on top of that:

`i18n.js`

- Despite the comment at the top of the file, i18n.js is not being used
  by the extract-intl script anymore, and can therefore use import/export
  syntax.

- The translation files are now imported dynamically so that only the
  necessary translations are included. In the example app, this doesn't
  make much difference, but when there are many locales, loading json
  files individually will make a significant difference in the initial load.  

- Create `fetchMessages` convenience function for importing/formatting
  translations.

`app.js`

- Update dynamic imports of Intl polyfills to @formatjs/intl-pluralrules
  and @formatjs/intl-relativetimeformat.

- Remove translationMessages as they are now being fetched by the
  getTranslationMessages Saga.

`LanguageProvider`

- The CHANGE_LOCALE action now triggers the getTranslationMessages Saga.
  The Saga will use the fetchMessages function in i18n.js to fetch the
  appropriate messages file and then update the locale and messages.

- If the messages are unable to be fetched, the saga will fall back to
  the default locale and messages.

`generators`

- Fix all language generators so that they work properly for new
  i18n.js/app.js implementaiton.


Can-Sahin:
Edit: Sorry I am opening this back as a result of not merging the #2935 if that means anything;)

PR #2873, opened 2020-03-10
Switch to network-first response strategy in offline-plugin

Can-Sahin:
The default `cache-first` strategy of the `offline-plugin` is a general problem as seen in #2750  and this [solution](https://github.com/react-boilerplate/react-boilerplate/issues/2750#issuecomment-536215256) prevents this

**Changes:**
- Switched to `network-first` strategy
- Added notes to `docs` some gotchas and reference links

### DIFF
--- a/docs/general/README.md
+++ b/docs/general/README.md
@@ -88,6 +88,9 @@ connection from the instant your users load the app. This is done with a
 ServiceWorker and a fallback to AppCache, so this feature even works on older
 browsers!
 
+> For better performance you can switch to [cache-first](https://github.com/NekR/offline-plugin/blob/master/docs/options.md#responsestrategy-cache-first--network-first) strategy.
+> You can also force update the SW/AppCache by following the [instructions](https://github.com/NekR/offline-plugin/blob/master/docs/updates.md), which will avoid delayed updating problems for the new deployments
+
 > All your files are included automatically. No manual intervention needed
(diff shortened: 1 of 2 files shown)


julienben (commented):
I'm trying this locally with my team and will be happy to see if it works well for us too.

Just in `app.js`, didn't you want to also add this?

```js
OfflinePluginRuntime.install({
  onUpdateReady: () => {
    // Tells to new SW to take control immediately
    OfflinePluginRuntime.applyUpdate();
  },
});
```

Unless I'm mistaken, it isn't there right now.

Can-Sahin:
Nope it isn't needed. It's for updating the app while the user is browsing in it.  (I updated my comment there. I realized it wasn't needed later)

bbar:
Honestly, I like the default responseStrategy of 'cache-first'. Load times are just so much better.

/app/app.js

```
import * as OfflinePluginRuntime from 'offline-plugin/runtime';

...

if (process.env.NODE_ENV === 'production') {
  OfflinePluginRuntime.install({
    onUpdateReady: () => {
      OfflinePluginRuntime.applyUpdate();
    },
    onUpdated: () => {
      // feel free to ask the user first
      window.location.reload();
    },
  });
}

```

/internals/webpack/webpack.prod.babel.js

```
new OfflinePlugin({
      // Add these 
      ServiceWorker: {
        events: true
      },
      AppCache: {
        events: true
      },
      ...rest
})
```

isthiiii (approved):
This worked for me

Issue #2880, opened 2020-03-16
MaterialUI production build imports the whole library


fakeplayerr:
## Description
I'm trying to add material ui library instead of the styled components.
 
I followed the guide in another issue [#2196](https://github.com/react-boilerplate/react-boilerplate/issues/2196) and it worked perfectly until i needed to build for production

When i built the app, the whole material-ui was added in the chunks.

I tried the examples for minifying on the material-ui docs but it didn't change the result

## Steps to reproduce
-- install @material-ui/core and @material-ui/icons
-- add material-ui to the app and use a button component 
-- run production build

**Expected behavior**
I want to minify the material ui bundle. Even if i'm using just a component, it takes the whole library

Among the chunks generated in the build i have this.

> npm.material-ui.488ad5b1cdb53aedb9db.chunk.js    446 KiB       1  [emitted]  [big]  npm.material-ui
> npm.material-ui.488ad5b1cdb53aedb9db.chunk.js.LICENSE.txt  663 bytes          [emitted]        > npm.material-ui.488ad5b1cdb53aedb9db.chunk.js.gz    108 KiB          [emitted]       

I don't know how to minify it. I tried a lot of things but I couldn't come up with anything.
From what I see, even the LICENSE.txt of material-ui gets bundled.

Any other ideas comes to mind?

Thank you and sorry if i put the question in the wrong place

## Versions

- React-Boilerplate: latest
- Node/NPM: latest LTS
- Browser: Chrome latest

Can-Sahin:
If you can provide a clone of this repo that material-ui as added very simply and still it bundles all together I can have a look at it. Important thing is can you clone the `dev` branch not `master`. There have been lot of changes since.

Can-Sahin:
Can you try avoiding the named imports to see if that works
Using this
`import Button from '@material-ui/core/Button';` 
instead of
`import { Button } from '@material-ui/core';`

If this works then webpack is the problem and I'll look into it

edit: similar with #2835

julienben:
This doesn't happen automatically. It needs to be configured via a babel plugin: https://material-ui.com/guides/minimizing-bundle-size/#option-2

Can-Sahin:
I also read that but it says its necessary if you wanna enable for development mode too

[Here](https://material-ui.com/guides/minimizing-bundle-size/#when-and-how-to-use-tree-shaking) it says it comes automatically with webpack > 2 but I just tried it doesn't work in production mode.  

> 
[...]

Issue #2835, opened 2019-12-08
Tree shaking


Avizura:
## Description
Tree shaking seems doesn't work with [styled-icons](https://github.com/jacobwgillespie/styled-icons).

## Steps to reproduce
1. `yarn add styled-icons`
2. import one icon e.g. `import { Plus } from 'styled-icons/fa-solid';` 
3. `yarn build`

**Expected behavior**
Bundle contains only imported icon.

**Screenshots**
Bundle contains all icons from `styled-icons/fa-solid`
![image](https://user-images.githubusercontent.com/11854101/70395985-84a0fd80-1a15-11ea-9bcb-3c2a8dce42a6.png)

## Versions

- React-Boilerplate: 4.0.0
- Node/NPM: 12.13.0 / 6.12.0

Can-Sahin:
https://github.com/jacobwgillespie/styled-icons/issues/814 

Can it be simply their issue ?

Avizura:
I don't think so because create react app does it properly. I have this issue with some other libs (e.g. [polished](https://github.com/styled-components/polished))

Can-Sahin:
Same problem -> #2880 

Check my [comment](https://github.com/react-boilerplate/react-boilerplate/issues/2880#issuecomment-601360031) for a quick fix.

PR #2881, opened 2020-03-19
fix: tree-shaking

Can-Sahin:
Fixing tree-shaking problems => #2880 #2835 

`jsnext:main`(which is [deprecated](https://github.com/jsforum/jsforum/issues/5)) is ruining the es6 modules resolution. Default value should be used so I removed it. 

See the [doc](https://webpack.js.org/configuration/resolve/#resolve-mainfields)


vineyardbovines (approved):

julienben:
Excellent job figuring that out!!

julienben (approved):

PR #2927, opened 2020-06-19
docs: Removed deprecated prettier setting reference

zzeebbii:
`prettier.eslintIntegration` setting is deprecated and it is not needed here. We also don't need to install the prettier extension in VSCode anyways to prettify our code as it is done by the prettier plugin in eslint config itself.


zzeebbii:
My apologies 😿 I accidentally created a pull request from my branch to the master. Now I know more about pull requests I guess 😄

PR #2928, opened 2020-06-19
docs: Removed deprecated prettier setting reference

zzeebbii:
`prettier.eslintIntegration` setting is deprecated and it is not needed here. We also don't need to install the prettier extension in VSCode anyways to prettify our code as it is done by the prettier plugin in eslint config itself.


Can-Sahin:
Edit: Sorry I opening this back as a result of not merging the #2935 if that means anything;)

